PR #67, opened 2016-09-09
Ocd 875

dlucas1:


andlar:
Two primary issues with this PR. @kekey1 might have more.

1) The same problem as in your data-model PR; the RELEASE_NOTES are obsolete, as you need to pull upstream/development into your local development branch, and then merge that into OCD-875 (while fixing any resulting merge conflicts).

2) In KeyValueModelStatuses, which extends from KeyValueModel, you redefine id, name, & description. The point of making a Class that extends from another Class is that the base Class methods and members are already in the new Class. I think you should be able to drop all code that references id/name/description from KeyValueModelStatuses, and just add to it with the 'statuses' portions. Overriding the equals/hashCode methods is fine, but I don't see that you actually changed the implementation of either, so I'm not sure why that was done.

kekey1:
Agree with Andrew - the KeyValueModel fields should not be rewritten in the subclass. That's the main thing. The only other thing I noticed was: 

* In DeveloperCertificationStatusesEntity

	/** Return the type of this class. Useful for when dealing with proxies.
	* @return Defining class.
	*/
	@Transient
	public Class<?> getClassType() {
		return DeveloperEntity.class;
	}

is returning the wrong type. Is this method ever called? It's in some of the oldest Entity classes that were auto-generated and probably got here from copying and pasting, but I think it's likely that it's not needed. Either fix or remove.

* Same comment for ProductCertificationStatusesEntity

PR #77, opened 2016-09-30
Ocd 897

dlucas1:
Made some housekeeping updates


kekey1:
In case I end up running later that expected on Monday my comments are below. Not sure how thorough they are... it's a lot of changes! 

- Need to resolve merge conflicts in the release notes. Maybe add a step to personal notes document about resolving local conflicts before submitting pull request?

- Remove environment.auth.properties from git entirely. It has things like our database password that we don’t want made public.

- You shouldn’t need to pass a reference to the “parseActivities” object to all the set* methods in ParseActivities class. You should just be able to call parseActivitiesInstance.setSomething(10) and the setSomething(int num) method of ParseActivities should be able to say this.something = num;  Think of ParseActivities as you would any other java class except that it happens to have a main() method.

- Similar comment as above regarding the getters in ParseActivities. You shouldn’t need to pass in the ParseActivities instance itself as an argument.

- ParseActivities getSummaryActivities() consider using SimpleDateFormat (or I just learned about the new java.time classes and they have great date/time formatting and timezone handling) instead of substring concatenation to format dates.  (currently lines 137 and 189) See this link for advice: http://stackoverflow.com/questions/16210409/is-the-tostring-method-in-java-util-date-locale-independent

- The switch statement in setCommandLineArgs can reduce duplication in the error message by saying

case 0:
case 1:
 throw new Exception(“…”);
 break;

- Line 313 either use or remove parseActivities method argument to loadEmailPropertes()

- Does this end up creating CSV files on the system that need to be removed over time? Do we want to keep an archive of these files? If not, could they be removed by the code when the job is finished?

- Do all test still pass? Noticed some developers/vendors were added to the test data - just wanted to make sure that didn’t affect other tests.

- In general classes that you are only creating with “new” like new Table() do not need to be declared as Spring components with the @Component annotation. It probably isn’t hurting anything but it’s not necessary to do. I think this is true of the Table* classes. Also the ReflectiveHelper class since it only has static methods. Anything you want to access as @Autowired *should* be a @Component.
 
--- 

Possible future improvements:

- Consider using the Apache Commons library that’s currently imported into chpl-service. It’s currently used for parsing upload files in CSV format (and maybe writing CSV somewhere but I can’t recall where). More robust and standard way to create CSV formatted file.

- Future-proof this code by making ActivitiesOutput accept a map like Map<String, List<DTO>> mapping each year to a list of associated DTOs rather than distinct lists for 2014, 2015, etc. We will almost certainly have many changes to make throughout the project for 2016 (and future) products but since this piece is new it’s a good opportunity to make next year easier on ourselves.

PR #90, opened 2016-10-07
Development

kekey1:
The tests should pass now. Decided this isn't a good value to check although I did get the same result as dev after running reset-test-only. In the future we may add more criteria which will just make this test fail again even though that failure wouldn't necessarily mean anything in the rest of our code is broken.


andlar:
It's still checking that the size is >0, which is probably good enough for now

PR #102, opened 2016-10-18
Ocd 955

dlucas1:
Please review and merge if all looks well.

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/AnnouncementManagerImpl.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/AnnouncementManagerImpl.java
@@ -92,12 +92,12 @@ public class AnnouncementManagerImpl extends ApplicationObjectSupport implements
 		this.announcementDAO = announcementDAO;
 	}
 
-	@PreAuthorize("hasRole('ROLE_ADMIN') or hasRole('ROLE_ONC_STAFF')")
+	@PreAuthorize("hasRole('ROLE_ADMIN')")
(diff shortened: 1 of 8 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/AnnouncementManagerImpl.java
```diff
@@ -92,12 +92,12 @@ public void setAnnouncementDAO(AnnouncementDAO announcementDAO) {
 		this.announcementDAO = announcementDAO;
 	}
 
-	@PreAuthorize("hasRole('ROLE_ADMIN')")
+	@PreAuthorize("hasRole('ROLE_ADMIN') or hasRole('ROLE_ONC_STAFF')")
```
andlar: I don't think ONC_STAFF needs access to manage Announcements. Is that what's being done here? This isn't the Announcement Activity stuff, right?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/ApiKeyManagerImpl.java
```diff
@@ -160,13 +160,13 @@ public void logApiKeyActivity(String keyString, String apiCallPath) throws Entit
 	}
 
 	@Override
-	@PreAuthorize("hasRole('ROLE_ADMIN')")
+	@PreAuthorize("hasRole('ROLE_ADMIN') or hasRole('ROLE_ONC_STAFF')")
```
andlar: Why does ONC_STAFF need to be able to get a list of all of the API keys?

PR #106, opened 2016-10-26
OCD-988 Caching fixes

dlucas1:
Please review and merge to dev for additional testing. Also, please note that we need to add this argument for Tomcat: 
-Djndi.name=java:comp/env/jdbc/openchpl


kekey1:
I'm fine with pushing these changes to test. The thing I'm most interested in seeing is if the transaction issue reappears on dev/stg or if that is resolved.

My only concern is whether this should be merged into the staging branch instead of dev since it's a "hot fix" and we don't want to include other changes that may already be in dev. I think staging is also already set up to provide the new jdni.name argument anyway.

dlucas1:
@kekey1 I believe we would like to make these changes within our regular deployment schedule. I think we'd like to let the changes sit on staging for awhile and allow Ashwini to perform a large set of regression tests.

PR #113, opened 2016-11-09
Ocd 996

kekey1:


kekey1:
@dlucas1 I agree that it could be brittle, however prior to my changes these same columns were listed in 3 places throughout the DAO and now they are all consolidated to 1 place. In that sense I think it's an improvement. I'm not sure how to apply the suggestions in that StackOverflow link to what we are doing in this file which is joining the main certified product details view to other views. Example below:

`String queryStr = "SELECT "
				+ "b.certified_product_id_cqms as \"certified_product_id\", " 
				+ COLUMNS_MINUS_CERTIFIED_PRODUCT_ID
				+ "FROM "
				
				+ "(SELECT certified_product_id_cqms FROM ( "
				+ "SELECT DISTINCT ON(cqm_id, certified_product_id_cqms) certified_product_id as \"certified_product_id_cqms\" "
			  	+ "FROM openchpl.cqm_result_details  WHERE deleted = false AND success = true AND cqm_id IN (:cqms)) a "
				+ "GROUP BY certified_product_id_cqms HAVING COUNT(*) = :ncqms ) b "

				+ "INNER JOIN openchpl.certified_product_details d "
				+ "ON b.certified_product_id_cqms = d.certified_product_id "
				
				+ "WHERE deleted <> true `";``

PR #117, opened 2016-11-17
OCD-1054: mark methods in product dao as transactional that are retur…

kekey1:
…ning

ownership history


kekey1:
The previous PR made certifiedProductDAO.findAll() transactional so that's already in there. As for developerDAO.findAllIncludingDeleted() ... the reason I needed to add @Transactional is that for any object that is getting a product and it's owner history (usually buried somewhere in the DTO constructor), hibernate was making an extra sql call to fill in that data. By the time that was getting called, the query had long finished and released it's connection to the db hence the "no session" exception. The CertifiedProductDetailsDTO and ProductDTO are the two objects that I think have reference to the product owner history. DeveloperDTO does not... I can make that method Transactional and I don't think it will hurt anything (will just join an existing transaction) but that's why I don't have it there.

PR #120, opened 2016-11-28
OCD-1018: functionality to support API call for meaningful_use_users csv upload with counts

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/Util.java
```diff
@@ -29,5 +31,27 @@ public static String md5(String input) {
 	    }
         return md5;
     }
+
```
kekey1: Are the two methods below used anywhere? I don't see them used throughout this PR. I think they can be removed.

PR #126, opened 2016-12-01
OCD-1016: Add API call to get decertified developers

dlucas1:


chpl/chpl-service/src/test/java/gov/healthit/chpl/dao/impl/DeveloperDaoTest.java
```diff
@@ -251,4 +253,32 @@ public void createDeveloperAcbMap() {
 		SecurityContextHolder.getContext().setAuthentication(null);
 	}
 	
+	/**
+	 * Given that I am authenticated to CHPL as any user
```
andlar: I don't see anything here that suggests you're actually authenticated to the CHPL. Maybe the Given should be something like: "Given the CHPL is accepting search requests"

chpl/chpl-service/src/test/java/gov/healthit/chpl/manager/impl/DeveloperManagerTest.java
```diff
@@ -237,4 +238,19 @@ public void testDeveloperStatusChangeNotAllowedByNonAdmin()
...
+	/**
+	 * Given that I am authenticated to CHPL as any user
+	 * When I call the REST API's /decertified/developers, the controller calls the developerManager.getDecertifiedDevelopers()
+	 * Then the manager returns a list of DeveloperDecertifiedDTO with expected results
+	 */
+	@Transactional
+	@Rollback(true) 
+	@Test
+	public void testGetDecertifiedDevelopers() {
+		SecurityContextHolder.getContext().setAuthentication(adminUser);
```
andlar: Doesn't this line suggest you're an authenticated user? But the "Given" above says "any user". Why authenticate? There's no security on this /decertifications call, right?

chpl/chpl-service/src/test/java/gov/healthit/chpl/web/controller/SearchViewControllerTest.java
```diff
@@ -172,4 +173,20 @@ public void test_getPopulateSearchData_simpleAsTrue_Caching_CompletesWithoutErro
...
+	/** 
+	 * Given that I am authenticated to CHPL as any user
+	 * When I call the REST API's /decertified/developers
+	 * Then the controller method's getDeveloperDecertifications returns expected results
+	 */
+	@Transactional
+	@Rollback(true)
+	@Test
+	public void test_getDeveloperDecertifications_CompletesWithoutError() throws EntityRetrievalException, JsonProcessingException, EntityCreationException{
+		SecurityContextHolder.getContext().setAuthentication(adminUser);
```
andlar: Again, here you're an authenticated user, but why?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
```diff
@@ -393,4 +400,20 @@ public void download(@RequestParam(value="edition", required=false) String editi
 		return searchMenuManager.getPopulateSearchOptions(simple);
 	}
 	
+	@ApiOperation(value="Get all developer decertifications in the CHPL",
```
andlar: I'm not sure why this is in the SearchViewController. Why isn't it in it's own DecertificationsController. It is a kind of search, but it's closer to a "getAll" (with a few filters) but the front end user isn't calling it with any filters at all. I could go either way on this one. @kekey1 ?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/DeveloperDecertificationResponse.java
```diff
@@ -0,0 +1,21 @@
+package gov.healthit.chpl.domain;
+
+import java.util.List;
+
+import gov.healthit.chpl.dto.DeveloperDecertifiedDTO;
```
andlar: This reads oddly to me. Would "DecertifiedDeveloperDTO" make more sense?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/DeveloperDecertifiedDTO.java
```diff
@@ -0,0 +1,60 @@
+package gov.healthit.chpl.dto;
+
+import java.util.List;
+
+public class DeveloperDecertifiedDTO {
```
andlar: Similarly, "DecertifiedDeveloperDTO"?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/DeveloperDecertificationResponse.java
kekey1: Agree

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/DeveloperDAOImpl.java
```diff
@@ -376,6 +378,63 @@ public DeveloperDTO getByVersion(Long productVersionId) throws EntityRetrievalEx
 		DeveloperEntity ve = (DeveloperEntity)result;
 		return new DeveloperDTO(ve);
 	}
+	
+	public List<DeveloperDecertifiedDTO> getDecertifiedDevelopers(){
+		List<CertifiedProductDetailsEntity> result = entityManager.createQuery(
+				"FROM CertifiedProductDetailsEntity "
+				+"WHERE developerStatusName IN ('Suspended by ONC', 'Under certification ban by ONC') ", CertifiedProductDetailsEntity.class).getResultList();
```
kekey1: I have an enum gov.healthit.chpl.entity.DeveloperStatusType . Could you use that instead of typing the names in here? If they ever decide these status names need to be changed we might get lucky and only have to change them in one place in the code.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
kekey1: I could see it maybe going into the DeveloperController as /developers/decertified or something like that. That said, if we ever ended up with thousands of decertified developers we might decide to make it more of a "search" and do the filtering on the server side but we could always move the method again if that happened.

andlar: I'd be wary of putting it in the DeveloperController, knowing we're going to be doing Decertified Certified Products next. I think those Decertified things should stay together, whether in the Search or in a different Controller

dlucas1: The title of OCD-1016 in Jira is "Add new /search for decertified Developers". Given the title name, I figured that if I am going to add a search, then I would do it in the SearchViewController. Should I move the code to the DeveloperController or a DecertificationsController?

kekey1: I'd say leave it where it is but don't have a strong feeling either way.

PR #128, opened 2016-12-01
OCD-1016: fixed nullPointerException; fixed unit tests; refactored project

dlucas1:
* Handled case where sometimes null numMeaningfulUse would cause nullPointerException when getting decertified developers at the DAO
	* Organized imports throughout unit tests and main project
	* Refactored all unit tests to be transactional and rollback any modifications
	* Added an ONC_ACB to testData.xml and updated a broken unit test to use it
	* Fixed sporadic failing unit tests


dlucas1:
Note that commit 2aced74 is actually for OCD-1015, so I undid that with commit 392f351 to keep OCD-1016 separate from OCD-1015.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/DecertifiedDeveloperDTO.java
```diff
@@ -46,15 +46,27 @@ public Long getNumMeaningfulUse() {
...
+		
 	}
 	
 	public void addAcb(String acb){
 		this.oncacb.add(acb);
 	}
 	
 	public void addNumMeaningfulUse(Long numMeaningfulUse){
-		this.numMeaningfulUse+= numMeaningfulUse;
+		if(this.numMeaningfulUse != null){
```
kekey1: I'm confused why the setter and this add method are doing the same thing (although the logic is reversed). Maybe the setter didn't need to be changed?

dlucas1: Just committed a change but pushed too soon. Need to make some fixes...

PR #132, opened 2016-12-05
OCD-1016: modified response object to include developer and certifyingBody

dlucas1:


andlar:
+1

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/DecertifiedDeveloperDTO.java
```diff
@@ -56,8 +40,24 @@ public void setNumMeaningfulUse(Long numMeaningfulUse) {
 		}
 	}
 	
-	public void addAcb(String acb){
-		this.oncacb.add(acb);
+	public void addAcb(Long acb){
+		this.acbList.add(acb);
```
kekey1: Would this work if someone got one of these objects by first calling the default constructor? Or if they passed in "null" as the list argument of the overloaded constructor?

dlucas1: A NullPointerException would be thrown if someone created a DecertifiedDeveloperDTO() object with the default constructor and attempted to add an acb to the acbList before initializing DecertifiedDeveloperDTO.acbList. Nothing breaks if null is passed in as the list argument of the overloaded constructor.

PR #133, opened 2016-12-05
OCD-1016: Fixed access denied by using DAO instead of manager

dlucas1:
Still need to pin down the cause to Andrew's 500 response for decertifications/developers before this can get merged.


dlucas1:
There's one more instance of using the CertificationBodyManager that needs to be replaced with the DAO.

PR #136, opened 2016-12-07
OCD-1064: added /decertifications/certified_products API call

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
```diff
@@ -497,4 +499,54 @@ public void download(@RequestParam(value="edition", required=false) String editi
...
+		SearchResponse fullResp = new SearchResponse();
+		SearchResponse decertifiedCPs = new SearchResponse();
+		List<CertifiedProductSearchResult> cpResultList = new ArrayList<CertifiedProductSearchResult>();
+		
+		if (pageNumber == null){
+			pageNumber = 0;
+		}
+		
+		SearchRequest searchRequest = new SearchRequest();
+
```
kekey1: Why don't we set the certificationstatuses filter of the searchRequest here? Then the DB could filter the results rather than having to put the logic in java below

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
```diff
@@ -497,4 +499,54 @@ public void download(@RequestParam(value="edition", required=false) String editi
...
+		
+		if (orderBy != null){
+			searchRequest.setOrderBy(orderBy);
+		}
+		
+		if (sortDescending != null){
+			searchRequest.setSortDescending(sortDescending);
+		}
+		
+		fullResp = certifiedProductSearchManager.search(searchRequest);
```
kekey1: Does this really get all the results? I thought the search page size defaulted to 20.

PR #144, opened 2016-12-15
OCD-1075: add CP_ids caching, asynchronous initialization at startup

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheInitializorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+@Component
+public class CacheInitializorImpl {
+	private static JWTAuthenticatedUser adminUser;
+	private static final Logger logger = LogManager.getLogger(CacheInitializorImpl.class);
+	
+	@Autowired private CertificationIdController certificationIdController;
+	@Autowired private SearchViewController searchViewController;
+	
+	@Async
+	public void initializeSearchOptions() throws IOException, EntityRetrievalException {
```
andlar: Can you cache the `getPopulateSearchData(true)` call as well? I think that's the one that's used in the admin section, and I'd love it if that were faster :)

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/CertificationIdController.java
```diff
@@ -125,6 +128,7 @@
 			+ "Returns a list of basic product information, " 
 			+ "Criteria and CQM calculations, and the associated CMS EHR Certification ID if one exists.")
 	@RequestMapping(value="/create", method=RequestMethod.POST, consumes=MediaType.APPLICATION_JSON_VALUE, produces={MediaType.APPLICATION_JSON_VALUE})
+	@CacheEvict("certificationIds")
```
andlar: This is the line that refreshes the cache when/if someone creates a new CMS ID, right? Wanted to make sure this was in here

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheInitializorImpl.java
dlucas1: Yes, can do/will do!

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheInitializor.java
```diff
@@ -0,0 +1,21 @@
+package gov.healthit.chpl.domain;
```
kekey1: The "domain" package contains objects that we use to communicate back and forth with the front end - the objects that get serialized into JSON. That doesn't seem like the right place for this class.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheInitializorImpl.java
```diff
@@ -0,0 +1,49 @@
+package gov.healthit.chpl.domain;
+
```
kekey1: Same comment as above.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheInitializorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+	
+	@Async
+	public void initializeCertificationIds() throws IOException{
+		adminUser = new JWTAuthenticatedUser();
+		adminUser.setFirstName("Administrator");
+		adminUser.setId(-2L);
+		adminUser.setLastName("Administrator");
+		adminUser.setSubjectName("admin");
+		adminUser.getPermissions().add(new GrantedPermission("ROLE_ADMIN"));
+		SecurityContextHolder.getContext().setAuthentication(adminUser);
```
kekey1: I'm pretty sure this is setting the authenticated user as the admin user globally for the entire application and it doesn't look like it's getting re-set at the end of this method to whatever the user was previously. Can you call a manager or dao method instead to get around needing the authentication?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheInitializorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import gov.healthit.chpl.web.controller.SearchViewController;
+
+@ContextConfiguration(classes = { gov.healthit.chpl.CHPLConfig.class })
+@Component
+public class CacheInitializorImpl {
+	private static JWTAuthenticatedUser adminUser;
+	private static final Logger logger = LogManager.getLogger(CacheInitializorImpl.class);
+	
+	@Autowired private CertificationIdController certificationIdController;
+	@Autowired private SearchViewController searchViewController;
```
kekey1: Accessing controllers this way really violates Spring's MVC principles. Check out the discussion here http://stackoverflow.com/questions/20374972/invoke-controller-method-from-java-class. I would really try to use the managers instead - if some business logic needs to be moved from the controllers into the managers so that can be done, that's fine.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheInitializor.java
dlucas1: Do you have a recommendation on where to put it?

kekey1: Not really, I was hoping you would look around and think of where might be appropriate. Two ideas: we are using the manager[.impl] packages as our service layer so that might be appropriate. Or you could make a new package under chpl (as a sibling of domain, manager, dto, etc) for caching stuff.

dlucas1:
Added caching for /search_options(true) in addition to /search_options(false)

RELEASE_NOTES.md
```diff
@@ -11,6 +11,7 @@ _Date TBD_
 * Add API call to get decertified certified products with pageCount set to the total number of decertified products
 * Add certificationDateStart and certificationDateEnd as advanced search parameters
 * Change corrective action plan search parameters to new surveillance search parameters - NOT BACKWARDS COMPATIBLE
+* add caching for /certifications_ids/ and /search_options(false), increase cache timeout for /search_options/, initialize cache at server startup
```
andlar: It's not just `/search_options(false)` right?

dlucas1:
As discussed in openchpl chat, I am going to re-work a lot of the caching and therefore close this for now. I will probably not have a PR ready until after I get back from vacation.

PR #151, opened 2016-12-23
Ocd 885

kekey1:
The file needs executable permissions and needs a cron set up on dev/stg/prod at the time we want it to run. Already ran it once on dev for testing purposes.

### DIFF
--- a/chpl/chpl-service/cleanupXml.sh
+++ b/chpl/chpl-service/cleanupXml.sh
@@ -6,10 +6,10 @@
 # This will run it at 0515 UTC, which (depending on DST) is 0015 EST
 
 #find the folder that has the XML files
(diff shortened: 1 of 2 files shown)


chpl/chpl-service/cleanupXml.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+(set -o igncr) 2>/dev/null && set -o igncr; # this comment is required to trick cygwin into dealing with windows vs. linux EOL characters
```
andlar: I think you should comment this line out for the final push. It's going to be run on a Linux machine, and I think we've commented out those lines on the various scripts that live on the Linux machines

PR #153, opened 2016-12-29
OCD-1075: Speed up website performance with Caching

dlucas1:
Note that this will not be merged/deployed to dev and staging until after we deploy the current release to production.

Just a heads up that I added a parameter to environment.properties and environment.properties.template as follows:
enableCacheInitialization=true

Cache initialization is disabled by default for when enableCacheInitialization is null/missing from the properties file. When deployment occurs, we'll need to remember to update the environment.properties file to enable asynchronous cache initialization at server startup. Also, Andrew and Katy will need to update their local environment.properties file to enable cache initialization.


dlucas1:
When we deploy this to dev/staging/production, we'll need to compare memory and disk utilization before & after the caching. In particular, we should closely monitor production utilization with many users hitting the server in order to make sure that the server does not run out of memory or disk space. I tried to limit the amount of data stored in memory to 1,000 on the heap in order to alleviate this potential problem, but again we should be vigilant regarding server resource utilization.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SearchRequest.java
```diff
@@ -144,4 +147,62 @@ public Boolean getHasHadSurveillance() {
...
 		this.hasHadSurveillance = hasHadSurveillance;
 	}
+	
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result
+				+ ((searchTerm == null) ? 0 : searchTerm.hashCode());
+		result = prime * result
```
kekey1: Do you mean to do "=" here and reassign the result of "result" or did you mean to do "+=" to add everything together? Overriding equals and hashcode can affect tons of underlying behavior so it *has* to be right.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/CertifiedProductManagerImpl.java
```diff
@@ -28,9 +28,9 @@
 import gov.healthit.chpl.dao.CQMResultDAO;
```
kekey1: This class has an update method and a createFromPending method. Should those also trigger all caches being cleared? They can result in new certified products (would change search results) or developer status changes.

andlar:
New branch created: https://github.com/chpladmin/chpl-api/tree/next-version

@dlucas1

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/CacheInitializorImpl.java
```diff
@@ -0,0 +1,107 @@
...
+import gov.healthit.chpl.dao.PendingCertifiedProductDAO;
+import gov.healthit.chpl.dao.impl.CertifiedProductSearchResultDAOImpl;
+import gov.healthit.chpl.domain.SearchRequest;
+import gov.healthit.chpl.domain.SurveillanceSearchOptions;
+import gov.healthit.chpl.dto.CertificationStatusDTO;
+import gov.healthit.chpl.manager.impl.CertificationIdManagerImpl;
+import gov.healthit.chpl.manager.impl.SearchMenuManagerImpl;
+
+@Component
+public class CacheInitializorImpl {
```
dlucas1: Asynchronous execution is weird and requires being called from another class in order to run on another thread, so I can't move CacheInitializorImpl into CacheInitializor. I can definitely rename to something more specific, though.

PR #154, opened 2017-01-02
OCD-1075: Speed up website performance with Caching

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SearchRequest.java
```diff
@@ -144,4 +147,62 @@ public Boolean getHasHadSurveillance() {
 	public void setHasHadSurveillance(Boolean hasHadSurveillance) {
 		this.hasHadSurveillance = hasHadSurveillance;
 	}
+	
+	@Override
+	public int hashCode() {
```
andlar: Why override hashCode? Also, is the algorithm you're using below some sort of standard kind of algorithm for generating a hashcode?

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/AsynchronousCacheInitialization.java
```diff
@@ -0,0 +1,124 @@
...
+		logger.info("Starting cache initialization for /pending");
+		CertificationStatusDTO statusDto = statusDao.getByStatusName("Pending");
+		pcpDao.findByStatus(statusDto.getId());
+		logger.info("Finished cache initialization for /pending");
+	}
+	
+	@Async
+	@Transactional
+	public void initializeSearch() throws EntityRetrievalException{
+		logger.info("Starting cache initialization for /search DAO method");
```
andlar: This method caches the "default" search, right? I'm wondering if it would be reasonable to make the default parameters some sort of config setting. Or if it's worthwhile trying to come up with a few more "versions" of search parameters that should be cached. One thing that worries me is that there are a lot of hard-coded parameters in here, and if we don't update these when the search gets updated (like with the impending new certification status that's coming, or if an ACB changes their name again) this cache wouldn't be very helpful

kekey1: Agree with not liking the hard-coded parameters. We let ACBs change their names right? What if Drummond Group changes their name to Drummond Group, Inc?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SearchRequest.java
```diff
@@ -144,4 +147,62 @@ public Boolean getHasHadSurveillance() {
...
+		result = prime * result + ((certificationEditions == null) ? 0 : certificationEditions.hashCode());
+		result = prime * result
+				+ ((certificationCriteria == null) ? 0 : certificationCriteria.hashCode());
+		result = prime * result
+				+ ((cqms == null) ? 0 : cqms.hashCode());
+		result = prime * result
+				+ ((certificationBodies == null) ? 0 : certificationBodies.hashCode());
+		result = prime * result
+				+ ((surveillance == null) ? 0 : surveillance.hashCode());
+		result = prime * result
```
kekey1: Certification bodies are counted twice?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SearchRequest.java
dlucas1: Serialization is required for an object to be placed into cache. There is no downside to serializing an object. In order to prevent errors when caching now and in the future, I serialized most of the classes.

hashCode needs to be overridden in order to get a unique hashCode for SearchRequest depending on the values of each of its properties. This is what allows someone to make a /search and have their unique search be cached. I believe I found the algorithm online in a tutorial.

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/AsynchronousCacheInitialization.java
dlucas1: The default here is meant to match the default provided by the UI for the search; however, given that these parameters are subject to frequent change, I'm fine with not initializing the /search if maintaining it is going to be a headache.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SearchRequest.java
andlar: Works for me. Do you have a link to the algorithm? I'm curious to see the reasoning behind it, though I don't have an issue with it here

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/AsynchronousCacheInitialization.java
andlar: I'd definitely like to cache the default search, but maybe we can find a way to put the "default search" as an external properties file? To make updating it in the future easier? If there's no way of doing that I'd still lean towards caching, even with hard-coded values, and maybe add some comments to the /search call somewhere to remind us to update the cache if those "defaults" change

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SearchRequest.java
dlucas1: I just spent a couple minutes searching but couldn't find the link where I found my specific algorithm. However, here is an article with some background information on the subject: http://crunchify.com/how-to-override-equals-and-hashcode-method-in-java/

PR #156, opened 2017-01-03
OCD-1092: /authenticate now returns 401 Unauthorized for bad credentials instead of 500 Internal Server Error

dlucas1:


dlucas1:
Updating release notes in a second...

PR #162, opened 2017-01-06
OCD-1075: cached /decertifications/developers

dlucas1:
After further inspection, I see that since /decertifications/certified_products is composed of two cached queries (countMultiFilterSearchResults(searchRequest) and certifiedProductSearchManager.search(searchRequest)), no change was needed; it was already being cached.

For the /decertifications/developers API call, I had to move things around from the controller to the manager in order to get everything cached.


RELEASE_NOTES.md
```diff
@@ -5,17 +5,19 @@ _Date TBD_
...
+	Add caching for /pending
+	Add caching for /search
+	Add caching for /decertifications/developers
+	Add caching for /decertifications/certified_products
+	Update caching for /search_options API call
+	Cache underlying data that improves call speed for many other API calls
+* TO-DO DURING DEPLOYMENT: Add these lines to environment.properties: 
+	enableCacheInitialization=true
+	cacheClearTimeoutSecs=15
+	cacheInitializeTimeoutSecs=300
```
kekey1: I think Andrew slightly changed the formatting here along with adding backticks around the properties - not sure if/why we need to change it back.

PR #165, opened 2017-01-16
OCD-1096: fixed comparison for boolean value

dlucas1:
Still working on the unit test for this (it's not easy to unit test), but I want to get this out to Ashwini to finish up her testing.

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2014.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2014.java
@@ -699,9 +699,6 @@ public class CertifiedProductHandler2014 extends CertifiedProductHandler {
 				ttEntity.setTestToolVersion(testToolVersion);
 				TestToolDTO testTool = testToolDao.getByName(testToolName);
 				if(testTool != null) {
-					if(testTool.isRetired() && product.getIcs() == false) {
-						product.getErrorMessages().add("Test tool '" + testToolName + "' has been retired for a product with ICS=false. Please remove it from the upload file and add a different test tool if necessary.");
-					}
 					ttEntity.setTestToolId(testTool.getId());
 				}
 				cert.getTestTools().add(ttEntity);
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2014.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2014.java
@@ -699,7 +699,7 @@ public class CertifiedProductHandler2014 extends CertifiedProductHandler {
 				ttEntity.setTestToolVersion(testToolVersion);
 				TestToolDTO testTool = testToolDao.getByName(testToolName);
 				if(testTool != null) {
-					if(testTool.isRetired() && product.getIcs().equals("false")) {
+					if(testTool.isRetired() && product.getIcs() == false) {
 						product.getErrorMessages().add("Test tool '" + testToolName + "' has been retired for a product with ICS=false. Please remove it from the upload file and add a different test tool if necessary.");
 					}
 					ttEntity.setTestToolId(testTool.getId());
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/AmbulatoryModular2014Validator.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/AmbulatoryModular2014Validator.java
@@ -10,10 +10,12 @@ import gov.healthit.chpl.dao.TestToolDAO;
 import gov.healthit.chpl.domain.CertificationResult;
 import gov.healthit.chpl.domain.CertificationResultTestTool;
 import gov.healthit.chpl.domain.CertifiedProductSearchDetails;
+import gov.healthit.chpl.dto.CertifiedProductDetailsDTO;
 import gov.healthit.chpl.dto.PendingCertificationResultDTO;
 import gov.healthit.chpl.dto.PendingCertificationResultTestToolDTO;
 import gov.healthit.chpl.dto.PendingCertifiedProductDTO;
 import gov.healthit.chpl.dto.TestToolDTO;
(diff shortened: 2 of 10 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2014.java
```diff
@@ -699,7 +699,7 @@ private void parseTestTools(PendingCertifiedProductEntity product, PendingCertif
 				ttEntity.setTestToolVersion(testToolVersion);
 				TestToolDTO testTool = testToolDao.getByName(testToolName);
 				if(testTool != null) {
-					if(testTool.isRetired() && product.getIcs().equals("false")) {
+					if(testTool.isRetired() && product.getIcs() == false) {
```
kekey1: In the parent class to this one I have a method "asBoolean". Would this have been of help to you? Users will often enter different boolean values 0, 1, true, false, tRuE, FALSE, etc and I tried to handle that variety

dlucas1: The asBoolean() method gets called at CertifiedProductHandler2015.handle() line 78 in parseCertifiedProductDetails(). This is done before CertifiedProductHandler2015.parseCriteria().parseTestTools() at line 176. Thus, a check for asBoolean() in CertifiedProductHandler2014/2015.parseTestTools() would be redundant.

kekey1: Got it! As discussed in slack, I think this logic actually needs to be moved into one of the validator classes. Leaving as changes requested for now.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/AmbulatoryModular2014Validator.java
```diff
@@ -91,8 +91,9 @@ protected void validateDemographics(PendingCertifiedProductDTO product) {
 							product.getErrorMessages().add("There was no test tool found matching '" + testTool.getName() + "' for certification " + cert.getNumber() + ".");
 						} else {
 							TestToolDTO tt = ttDao.getById(testTool.getTestToolId());
-							if(tt != null && tt.isRetired()) {
-								product.getErrorMessages().add("Cannot use test tool '" + tt.getName() + "' since it is retired. Please choose a different test tool for " + cert.getNumber() + ".");
+							if(tt != null && tt.isRetired() && product.getIcs() == false) {
+								product.getErrorMessages().add("Test Tool '" + testTool.getName() + "' for Certification Result '" + cert.getNumber() 
+								+ "' has been retired for product with ICS=false.");
```
andlar: Can we change this line to something like `can not be used, as it is a retired tool, and this Certified Product does not carry ICS`? Anywhere we have `ICS=false` that phrase is too programmery for our end users

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/AmbulatoryModular2014Validator.java
```diff
@@ -91,8 +93,25 @@ protected void validateDemographics(PendingCertifiedProductDTO product) {
 							product.getErrorMessages().add("There was no test tool found matching '" + testTool.getName() + "' for certification " + cert.getNumber() + ".");
 						} else {
 							TestToolDTO tt = ttDao.getById(testTool.getTestToolId());
-							if(tt != null && tt.isRetired()) {
-								product.getErrorMessages().add("Cannot use test tool '" + tt.getName() + "' since it is retired. Please choose a different test tool for " + cert.getNumber() + ".");
+							String[] idParts = product.getUniqueId().split("\\.");
+							if(idParts.length < 9) {
+								logger.error("CHPL ID must have 9 parts separated by '.'");
+							}
+							if(tt != null && tt.isRetired() && idParts[6].toString().equals("0")) {
```
kekey1: Looks like if the id doesn't have 9 parts it logs an error but then continues on trying to access the 6th part. Should this be in an else or reorganized?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/AmbulatoryModular2014Validator.java
```diff
@@ -91,8 +93,25 @@ protected void validateDemographics(PendingCertifiedProductDTO product) {
...
-							if(tt != null && tt.isRetired()) {
-								product.getErrorMessages().add("Cannot use test tool '" + tt.getName() + "' since it is retired. Please choose a different test tool for " + cert.getNumber() + ".");
+							String[] idParts = product.getUniqueId().split("\\.");
+							if(idParts.length < 9) {
+								logger.error("CHPL ID must have 9 parts separated by '.'");
+							}
+							if(tt != null && tt.isRetired() && idParts[6].toString().equals("0")) {
+								Boolean containsInheritedCertificationStatus = false;
+								for (String error: product.getErrorMessages()) {
+									if(error.contains("Inherited Certification Status is true")){
```
kekey1: I don't think it's best practice to compare on this string if we decide to change the error message in the future.  A better way might be to set a protected field in the parent class where the error about ICS conflict is originally generated and just check that. Then you won't have to worry about parsing the ID into its separate parts either.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
```diff
@@ -195,17 +195,17 @@
...
-		if(updateIcs == false && existingProductIcs == true && hasRetiredTestTool == true){
-			List<String> retiredTestToolNumbers = new ArrayList<String>();
-			for(CertificationCriterionEntity cce : retiredTestTools){
-				retiredTestToolNumbers.add(cce.getNumber());
-			}
-			
-			updateRequest.getErrorMessages().add("Cannot set ICS to false for a Certified Product with ICS=true and attested criteria that have a retired Test Tool. "
-					+ "The following are attested criteria that have a retired Test Tool: " + retiredTestToolNumbers.toString());
-		}
+//		Boolean updateIcs = updateRequest.getIcs();
```
kekey1: delete :)

PR #174, opened 2017-01-23
OCD-1159 API: Allow editing of date

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/MeaningfulUseAccurateAsOf.java
```diff
@@ -0,0 +1,97 @@
...
+	private static final long serialVersionUID = -4803763243075068608L;
+	
+	private Long id;
+	private Long accurateAsOfDate;
+	private Boolean deleted;
+	private Long lastModifiedUser;
+	private Long creationDate;
+	private Long lastModifiedDate;
+	
+	private Set<String> warningMessages = new HashSet<String>();
```
kekey1: I don't think the error and warning messages are necessary

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/MeaningfulUseAccurateAsOfEntity.java
```diff
@@ -0,0 +1,88 @@
...
+	
+	@Id 
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
+	@Basic(optional = false)
+	@Column(name = "muu_accurate_as_of_date_id", nullable = false)
+	private Long id;
+	
+	@Basic(optional=false)
+	@Column(name = "Accurate_as_of_date", updatable = true, nullable = false)
+	private Date accurateAsOfDate;
```
kekey1: Case must not matter here although I would have thought on Windows it would. Make "accurate" in the column name lowercase for consistency.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/MeaningfulUseController.java
```diff
@@ -0,0 +1,185 @@
...
+	}
+	
+	@ApiOperation(value="Get a single date value to indicate when the last meaningful use user file data is accurate as of.", 
+			notes="This is a single system-wide value.")
+	@RequestMapping(value="/accurate_as_of", method=RequestMethod.GET,
+			produces="application/json; charset=utf-8") 
+	public @ResponseBody MeaningfulUseAccurateAsOf getAccurateAsOfDate() {
+		MeaningfulUseAccurateAsOfDTO dto = muManager.getMeaningfulUseAccurateAsOf();
+		MeaningfulUseAccurateAsOf muuAccurate = new MeaningfulUseAccurateAsOf(dto);
+		if(StringUtils.isEmpty(muuAccurate.getId().toString()) && StringUtils.isEmpty(muuAccurate.getAccurateAsOfDate().toString())){
```
kekey1: - If muuAccurate.getId() is null or muuAccurate.getAccurateAsOfDate() is null this will throw an NPE. 

- Unless there are requirements for other reasons to use errorMessages and warningMessages, I don't think that is appropriate here since there is only ever one error message and never any warning messages. You probably just want to throw an exception of a type that is handled as Controller Advice (or add a handler for a different type to that class) with the error message. Maybe EntityNotFoundException? or EntityExistsException?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/MeaningfulUseController.java
```diff
@@ -0,0 +1,185 @@
...
+			muuAccurate.getErrorMessages().add("Could not obtain the accurate as of date.");
+		}
+		return muuAccurate;
+	}
+	
+	@ApiOperation(value="Update the Meaningful Use Accurate As Of date.", 
+			notes="Accurate as of date value can be edited by a user with ROLE_ADMIN and ROLE_CMS_STAFF.")
+	@RequestMapping(value="/accurate_as_of", method=RequestMethod.POST,
+			produces="application/json; charset=utf-8") 
+	public @ResponseBody MeaningfulUseAccurateAsOf updateMeaningfulUseAccurateAsOf(@RequestParam(required=true) MeaningfulUseAccurateAsOf meaningfulUseAccurateAsOf) {
```
kekey1: In terms of RESTful API design, I wouldn't expect a user to have to send in this whole object when really all they want to set is a single date. Could you just let them pass in a date? I think all the other info (last modified user, etc) is already available to you in the UI. I thought we originally said that the table will always have a single row so I don't think you need the id passed in to update it. I could have missed some discussion so if there is reason for doing it this way, that's fine.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/MeaningfulUseAccurateAsOf.java
dlucas1: I'm adding an error in the controller when the POST update is called without an accurateAsOfDate. I added the warningMessages in as convention since that seems to be the standard in other objects returned by the API.

kekey1: I left a comment down there, too. :) Only Certified Products and Surveillance have errorMessages/warningMessages since those are things that are uploaded and can have a long list of things wrong with them. In cases where it's a single error that's occurring the restful way to do it is just send an appropriate error code with a message describing what is wrong.

dlucas1: Okay, that makes sense.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/MeaningfulUseController.java
dlucas1: I originally had the controller passing in a Long timeinMillis and changed it to this after thinking about it some. Here were my thoughts:
1. Designing the controller with the object passed in allows greater flexibility and forward compatibility. I added functionality to the controller that allows a user to pass in only the accurateAsOfDate, so the user doesn't need to know the id or pass that in. The user can pass in null for everything but accurateAsOfDate.
2. What if we want to update the deleted column or insert more rows in the future? 
3. What if we add more columns in the future? 
4. What if this table ends up housing more than one row?
5. I didn't find any examples of API calls in CHPL where the user passed in primitive type rather than an object, so I was trying to follow existing examples by using an object.

PR #190, opened 2017-02-16
OCD-1243: add surveillance activities to weekly statistics email

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/app/ParseActivities.java
```diff
@@ -420,7 +504,25 @@ else if(dto.getYear().equals("2015")){
 	public void setEmailProperties(){
 		 email.setEmailTo(props.getProperty("summaryEmail").toString().split(";"));
 		 email.setEmailSubject("CHPL - Weekly Summary Statistics Report");
-		 email.setEmailMessage(summaryOutputTable.getTable());
+		 Calendar calendarCounter = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+		 String htmlPreText = "<pre><br>";
```
andlar: What happens if you don't put the `<pre>` tags in here? I think it was probably of use when you were making the ASCII table, but for an `<ul>` I don't know if fixed-width font is required. Not sure if changing it is important or not, but I am curious.

dlucas1: Removing the pre tag improves the appearance of the email somewhat because it removes some unnecessary spacing.

PR #194, opened 2017-02-22
Ocd 1240

kekey1:


dlucas1:
This will require updating the environment.properties file during deployment. How are we agreeing on reminding ourselves that we'll need to do this now that we're not updating the release notes with a "TO-DO Deployment" note?

PR #197, opened 2017-02-27
OCD-1250: validate user's authority for create/update/delete surveillance

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/SurveillanceValidator.java
```diff
@@ -349,4 +355,26 @@ public void validateSurveillanceNonconformities(Surveillance surv) {
 			}
 		}
 	}
+	
+	public void validateSurveillanceAuthority(Surveillance surv) {
```
andlar: This function has some issues. 

1. A user can have multiple ROLEs, but if they do, the call to create a surveillance will only ever pass in one ROLE, as chosen by the user doing the activity. Assuming that ROLE is one of the correct ones to use, the system should use that passed in role. If it's not one of the Authorized roles the system should return an "Unauthorized" error.
2. CMS_STAFF has no ability to do anything about Surveillances. There are three ROLEs that are allowed to do stuff with Surveillance. ROLE_ADMIN, ROLE_ACB_ADMIN, and ROLE_ACB_STAFF. Of those three, ROLE_ADMIN is allowed to do everything, but the other two are the ACB versions, so they are not allowed to edit/delete a surveillance that was created by ROLE_ADMIN
3. ROLE_ONC_STAFF should also not be here at all. That ROLE is only for viewing reports & downloading the CMS file. It should have no ability to edit or modify any listings

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
```diff
@@ -668,4 +680,19 @@ private void checkUploadedSurveillanceOwnership(Surveillance pendingSurv) {
 			} 
 		}
 	}
+	
+	private void updateAuthority(Surveillance surv){
```
andlar: Is this method intended to update the Authority of an already created Surveillance? If so, it should not do that. Once a Surveillance has been created with a particular ROLE, then that ROLE shall not be able to be changed

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/SurveillanceValidator.java
dlucas1: We are using @PreAuthorize on methods in the SurveillanceManager to restrict access to users for create/delete/update, for example:

@PreAuthorize("hasRole('ROLE_ADMIN') or "
			+ "((hasRole('ROLE_ACB_STAFF') or hasRole('ROLE_ACB_ADMIN')) "
			+ "and hasPermission(#acbId, 'gov.healthit.chpl.dto.CertificationBodyDTO', admin))")

andlar: PreAuthorize is fine for broad blocking, but it won't stop an ACB_ADMIN user from editing a Surveillance that was created by a ROLE_ADMIN user

dlucas1: #1. Agreed that the Surveillance object's authority will only have one String value. Regarding the return of Unauthorized - we're doing authorization in the managers with the @PreAuthorize annotation. The purpose of the SurveillanceValidator is primarily to add an error message if the business logic is not met. I don't think we should use the value of the Surveillance object's authority to determine authorization; I think we should use the User object's role and permissions on the ACB. Currently, we _will_ return unauthorized because of the manager methods' @Preauthorize annotations. I thought that an error message would suffice for the case when an ACB_ADMIN tries to update a surveillance that was created by ROLE_ADMIN, but if we're in agreement that this should instead cause an "Unauthorized" error, then I'll go ahead and change that functionality.
#2. I think I misinterpreted the requirement in OCD-1180 for "ONC Admin" and OCD-1250 for "ONC user" as being ROLE_ONC_STAFF. I think I need to change references to ROLE_ONC_STAFF to ROLE_ADMIN. I think I also need to add the ability for surveillance authority to equal ROLE_ACB_STAFF, right? Or will you always pass in ROLE_ACB_ADMIN for ACB users?
#3. See my comment to #2.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
dlucas1: No; the purpose of this method is to handle when the UI passes in a null Surveillance authority.

andlar: Okay, but this method is being called in the /update method. Why? And on create, if the ROLE isn't supplied, then that's an invalid /create call, so the request shouldn't be accepted

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/SurveillanceValidator.java
andlar: 1. I'm fine with the error message related to the ROLE
2. The UI will pass in one of three ROLEs: ROLE_ADMIN, ROLE_ACB_ADMIN, or ROLE_ACB_STAFF

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
dlucas1: Since I believe the UI will not allow the user to update the authority, I think I can remove the call to updateAuthority() from the /update method. 

For the create method, the API controller requires a RequestBody with a Surveillance object, but I don't think it will catch if one of the Surveillance properties is null--that's why we have the validators, right? I thought the requirement in OCD-1250 stated that if the request has a null ROLE, then to use the role from the User object. Thus, we shouldn't deny the request from being accepted. Am I misunderstanding something?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
```diff
@@ -157,8 +160,7 @@ public void streamDocumentContents(@PathVariable("documentId") Long documentId,
 			produces="application/json; charset=utf-8")
 	public synchronized @ResponseBody Surveillance createSurveillance(
 			@RequestBody(required = true) Surveillance survToInsert) 
-		throws InvalidArgumentsException, ValidationException, EntityCreationException, 
-		EntityRetrievalException, CertificationBodyAccessException, JsonProcessingException {
+		throws Exception {
```
kekey1: This isn't wrong, but it's preferable to have a specific list of exceptions that can be thrown by the method over the generic Exception in case the code or user that handles the errors wants to handle different types of exceptions differently. I don't think the throws clause needed to be changed in this case.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
```diff
@@ -390,7 +403,7 @@ public String deleteNonconformityDocument(@PathVariable("surveillanceId") Long s
 			produces="application/json; charset=utf-8")
 	public synchronized @ResponseBody Surveillance confirmPendingSurveillance(
 			@RequestBody(required = true) Surveillance survToInsert) 
-		throws InvalidArgumentsException, ValidationException, EntityCreationException, EntityRetrievalException, JsonProcessingException {
+		throws Exception {
```
kekey1: Same comment as above on the throws clause

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
andlar: I'd forgotten about the possibility of a NULL coming from the UI. I think I'm okay with taking the ROLE from the user in that case

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/SurveillanceValidator.java
kekey1: I think by the time the validator code is run, the *only* check should be that the authority in the surveillance object isn't null or empty. Based on looking at the rest of the code, the authority property of surveillance is either getting filled in by the UI and passed in on the request or set in the controller or manager code based on the role the user has. I could be missing something but I feel like this can be greatly simplified.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
kekey1: Why is it ONC_STAFF here? I thought it was ONC_ADMIN and ACB_ADMIN or ACB_STAFF that we cared about.

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/PendingSurveillanceEntity.java
```diff
@@ -51,18 +51,21 @@
...
+	@Column(name = "creation_date", insertable = false, updatable = false)
 	private Date creationDate;
 	
-	@Column( name = "last_modified_date", insertable = false, updatable = false )
+	@Column(name = "last_modified_date", insertable = false, updatable = false)
 	private Date lastModifiedDate;
 	
+	@Column(name = "user_permission_id", insertable = false, updatable = true)
+	private Long userPermissionId;
+
```
dlucas1: Correct, this code shouldn't have been leftover.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
dlucas1: The ONC_STAFF issue was discussed in the comment above this one. This is my reply:

> I think I misinterpreted the requirement in OCD-1180 for "ONC Admin" and OCD-1250 for "ONC user" as being ROLE_ONC_STAFF. I think I need to change references to ROLE_ONC_STAFF to ROLE_ADMIN.
>

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/SurveillanceValidator.java
dlucas1: Really? When I review the requirements in OCD-1180 and OCD-1250, I think more checks are needed.

kekey1: I'm envisioning the Controller works pretty much as you have it and either the Authority is put on the Surveillance object by the UI or you fill it in based on the user's role. Then validate() is called from the Controller which would check that the Authority is indeed filled in (you're right - it should also check that the authority value is valid based on the roles the current user has and the possible choices but it can still be simplified from what is there). The Manager would be where roles are checked... so in the survManager.create() method, as long as the validator passed in the controller you'd know the surveillance has an allowed authority and can be created. The Manager update method is slightly more complicated only in that you need to check that the user performing the update isn't ACB_ADMIN if the Authority in the existing surveillance db row is ONC_ADMIN (if ONC created the surveillance then ACB can't be the one to update it) - basically just check that the user performing the update is allowed to.

dlucas1: Right now, the SurveillanceController has an updateNullAuthority() method that updates the authority depending on the User object's role. I'm confused because you say that the Controller works pretty much as I have it (it's checking the User roles and updating them there), but then you're advocating for checking roles in the manager. 

Again, I'm doing the below functionality in the SurveillanceValidator.validateSurveillanceAuthority (which gets called from the validate(Surveillance surv) method):

> The Manager update method is slightly more complicated only in that you need to check that the user performing the update isn't ACB_ADMIN if the Authority in the existing surveillance db row is ONC_ADMIN (if ONC created the surveillance then ACB can't be the one to update it) - basically just check that the user performing the update is allowed to.

> 

Isn't it better/faster to check the the user's roles and throw a ValidationException before getting to the manager?

RELEASE_NOTES.md
```diff
@@ -2,8 +2,12 @@
 
 ## Version TBD
 _27 March 2017_
```
andlar: Don't want future dates in here now; we're not sure when things will be pushed, and by having this here it'll make it harder to keep track

PR #202, opened 2017-03-20
OCD-1285: improve performance of pending survs

dlucas1:


dlucas1:
The errors and error counts are the same after changes as before the changes. I uploaded many different criterion that are invalid for the requirement type the product has attested to and saw the appropriate error. Here are examples of the kinds of errors that were tested:
- Pending surveillance is supposed to replace existing surveillance with id 1 but no surveillance with that ID could be found.
- The requirement '170.315 (d)(10)' is not valid for requirement type 'Certified Capability'. Valid values are any of the criterion this product has attested to.

PR #209, opened 2017-03-28
OCD-1287: Cache results of pending CPs

dlucas1:


dlucas1:
Note: can only deploy to development until Wednesday 3/29/2017 when Staging thaws.

dlucas1:
By the way, I tried using "JOIN FETCH" in the underlying query, but the performance degraded significantly. I think this is in large part because we don't need all the data for every Pending CP, but adding the "JOIN FETCH" means every call gets everything. 

It turns out that all of the SQL queries/lazily-joined-data occurs during the creation of the DTO from the entity, and that happens in PendingCertifiedProductDAO.findByAcbId(Long acbId), so by caching that method I'm caching all of those additional database calls.

PR #213, opened 2017-03-30
OCD-1332: Disallow CAPEndDate for Open NC

dlucas1:


dlucas1:
Yes, release notes were already updated and should cover this change:
* Show error to ONC-ACB during upload and edit for a Surveillance Activity nonconformity that violates business rules for "Corrective Action Plan End Date"

PR #216, opened 2017-04-05
OCD-1352: update GET /search endpoint to allow all params as POST

kekey1:
Please look closely at the parameter documentation which is all the @ApiImplicitParam annotations. I am not sure that I got the defaults right on all of them or that they are all accurate descriptions of the behavior.  Mostly things like how we exclude Retired by default... are there other things like that?? Feel free to change any of them or just make a list for me and I will do it.

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
@@ -38,7 +38,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import gov.healthit.chpl.dao.CertifiedProductSearchResultDAO;
 import gov.healthit.chpl.dao.EntityRetrievalException;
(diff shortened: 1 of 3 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
```diff
@@ -264,40 +271,287 @@ public void download(@RequestParam(value="edition", required=false) String editi
...
+						throw new InvalidArgumentsException("No surveillance search option matches " + surveillanceParam);
+					}
+				}
+				searchRequest.setSurveillance(surveillanceSearchOptions);
+			}
+		}
+		
+		searchRequest.setHasHadSurveillance(hasHadSurveillance);
+		if(!StringUtils.isEmpty(developer)) {
+			searchRequest.setDeveloper(developer.trim());
```
andlar: Does `StringUtils.isEmpty` count a string of spaces (`   `) as empty?

PR #217, opened 2017-04-07
OCD-1394: add /cache_status endpoint and show in Swagger docs

kekey1:
was looking for something i could complete in a shortish amount of time this afternoon to avoid starting something and forgetting about it over the weekend. this seems to work!

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,7 @@ _Date TBD Next
 
 ### Features Added
 * Update parameters for HTTP GET /search call. All the same options that were previously available only to POST exist for the GET. Updated related API documentation.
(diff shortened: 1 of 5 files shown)


andlar:
I pulled in these changes and tested them to see if the automated deployment will work. Unless we want to require an API key to check this status endpoint, which I don't think we do, we'll need to add it to the list of exceptions, which I did in the added commit. I think I could be okay undoing that, but then we'd need to run command line stuff with API Keys as request parameters, and I'd rather avoid that. @kekey1 , thoughts? @dlucas1 ?

kekey1:
I agree!!

dlucas1:
I also agree with allowing a call to /cache_status endpoint without an API key.

PR #227, opened 2017-04-18
OCD-1393: remove numbers from email body

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/app/statistics/SummaryStatistics.java
```diff
@@ -189,53 +189,51 @@ public void setEmailProperties(Statistics stats, List<File> files){
 		 Calendar calendarCounter = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
 		 StringBuilder emailMessage = new StringBuilder();
 				 emailMessage.append("<ul><li>Date: " + calendarCounter.getTime() + "</li>");
```
andlar: I don't think this line should be part of the list. The Date isn't part of the statistics, but is more of a meta-data value for the stats

PR #245, opened 2017-05-03
Development

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/CertificationResultDAOImpl.java
```diff
@@ -1091,10 +1091,11 @@ public CertificationResultTestFunctionalityDTO lookupTestFunctionalityMapping(Lo
...
 		
-		Query query = entityManager.createQuery( "SELECT tf "
-				+ "FROM CertificationResultTestFunctionalityEntity tf "
-				+ "LEFT OUTER JOIN FETCH tf.testFunctionality "
-				+ "where (NOT tf.deleted = true) AND (tf.id = :entityid) ", 
+		Query query = entityManager.createQuery( "SELECT crtf "
+				+ "FROM CertificationResultTestFunctionalityEntity crtf "
+				+ "LEFT OUTER JOIN FETCH crtf.testFunctionality tf "
+				+ "JOIN FETCH tf.certificationEdition edition "
+				+ "where (NOT crtf.deleted = true) AND (crtf.id = :entityid) ",
```
andlar: I know this is a tiny thing, but the mis-match of `UPPERCASE SQL` and `lowercase sql` things is odd to me. I don't care about which one we use, but I feel like we should at least try to keep them consistent inside a single query ;)

PR #248, opened 2017-05-12
OCD-1452: surveillance oversight daily/weekly reports with subscriptions

dlucas1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/app/SurveillanceOversightReportWeeklyApp.java
```diff
@@ -1,115 +1,99 @@
...
 		
-        //write out a csv file containing all surveillance
-        String surveillanceReportFilename = downloadFolder.getAbsolutePath() + File.separator + FILENAME;
-        File surveillanceReportFile = new File(surveillanceReportFilename);
+		surveillanceReportFilename = props.getProperty("oversightEmailWeeklyFileName");
+    	surveillanceReportFile = new File(downloadFolder.getAbsolutePath() + File.separator + surveillanceReportFilename);
         if(!surveillanceReportFile.exists()) {
         	surveillanceReportFile.createNewFile();
         } else {
         	surveillanceReportFile.delete();
```
andlar: Why delete this file?

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/SurveillanceOversightReportWeeklyApp.java
```diff
@@ -133,47 +117,66 @@ public static void main( String[] args ) throws Exception {
...
+        // Send one email for each ACB to all of its subscribed recipients
+        for(Map.Entry<CertificationBodyDTO, Set<String>> entry : acbEmailMap.entrySet()){
+        	List<File> files = new ArrayList<File>();
+        	String fmtAcbName = entry.getKey().getName().replaceAll("\\W", "").toLowerCase();
+        	surveillanceReportFilename = fmtAcbName + "-" + props.getProperty("oversightEmailWeeklyFileName");
+        	surveillanceReportFile = new File(downloadFolder.getAbsolutePath() + File.separator + surveillanceReportFilename);
+            if(!surveillanceReportFile.exists()) {
+            	surveillanceReportFile.createNewFile();
+            } else {
+            	surveillanceReportFile.delete();
```
andlar: Also. What scenario would there be where the file would already exist, and you'd want to delete it, but then attach the file below?

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/SurveillanceOversightReportWeeklyApp.java
```diff
@@ -133,47 +117,66 @@ public static void main( String[] args ) throws Exception {
...
+            }
+            // Generate this ACB's download file  	
+            this.getPresenter().presentAsFile(surveillanceReportFile, acbDownloadMap.get(entry.getKey()));
+    		files.add(surveillanceReportFile);	
+        	
+        	subject = entry.getKey().getName() + " " + props.getProperty("oversightEmailWeeklySubject");
+        	htmlMessage = props.getProperty("oversightEmailAcbWeeklyHtmlMessage");
```
andlar: Can you add the ACB name to this message? I think the body of the email is currently something like:
```
<h3>title</h3>
either <p>nothing</p>
or <ul>things broken</ul>
```
Can you put an `<h4>ACB Name</h4>` after the `<h3>` and before either the `<p>` or `<ul>`?

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/SurveillanceOversightReportWeeklyApp.java
kekey1: I think I wrote this bit of code so I'll try to answer... and it may be that the code isn't clear enough/right (although it's working). I think the intent is to either create a new file or erase the contents of an existing file since we are creating files with the same names every time.  I was trying to just always make sure we started writing in a blank file rather than possibly have any of the content in the file last time it was written get mixed in with the current write. Dan, do you want to see if there is a better way to do this? 

The file gets immediately re-created when presentAsFile() is called in the next few lines. It's always written before it's attached to any emails.

dlucas1: It looks like the application functions just fine with lines 86-89 removed. I tested these two scenarios successfully with lines 86-89 commented out (for both send oncDaily/WeeklyEmail() and sendAcbDaily/WeeklySubscriptionEmails():

Test Case 1:
File in \environmentDownloadFiles already exists (has old/different data). Verify that old file is not attached to email.

Test Case 2:
No file exists. Verify that application does not break; new file should be created only once and attached to email.

andlar:
You removed the part where you delete a file that exists already, right? What happens if there already is a file now? Will it overwrite it?

PR #258, opened 2017-05-23
OCD-1517: Temp app to update cert status to 'Withdrawn by Developer' for CCHIT's 2014 edition retired listings

dlucas1:
When this is deployed, we'll need to update Bamboo for the environment.properties parameters. Also, this will need to be run during off-hours (manually executed or with crontab?) for the following reasons:
1. It takes ~1 minute per CP to run the update, and this app updates 216 CPs
2. Each update evicts the cache, which means there's going to be a lot of cache refreshing going on
3. This app will use some system resources; it's best not to hog resources during business hours.


dlucas1:
Since there's now an HttpUtil.java and Token.java, should we perhaps consider keeping the pom.xml references since they're utilized in these two classes that we may want to keep around? Is there any harm in keeping the pom references?

Things to remove after the app is successfully run:
- UpdateCertificationStatusApp.java
- UpdateCertificationStatusApp.sh
- environment.properties.template properties for temp-app properties (keep disableSuspiciousActivityEmails)
- environment.properties properties for temp-app properties (keep disableSuspiciousActivityEmails)
- **?? HttpUtil.java ??**
- **?? Token.java ??**
- **?? pom.xml references for gson, httpcomponents, org.json, and org.apache.commons ??**

dlucas1:
Alternatively, maybe we should keep a package under gov.healthit.chpl.app called "temp" so that we know these are apps/files that are used temporarily? That would reduce the headache of removing the app artifacts, plus it would make it easier to reference these past apps for similar future ones without making people think that the apps are in active service.

PR #261, opened 2017-06-01
OCD-1517: add connectTimeout and socketTimeout so job does not get hung up on one call

dlucas1:
Also handles responses as UTF-8 which can prevent a bad POST request due to a special character being returned in the GET.


chpl/chpl-service/src/main/java/gov/healthit/chpl/app/HttpUtil.java
```diff
@@ -29,6 +29,8 @@
 
 public class HttpUtil {
 	private static final Logger logger = LogManager.getLogger(HttpUtil.class);
+	private static final int CONNECTION_TIMEOUT = 600000;
+	private static final int SOCKET_TIMEOUT = 600000;
```
andlar: Can you add a comment that converts this number to a "people friendly" size? Is this 10 minutes? 60? Just looking at it I have no idea even what order of magnitude this is

PR #277, opened 2017-07-07
Development

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/SurveillanceManagerImpl.java
```diff
@@ -510,6 +509,14 @@ private Surveillance convertToDomain(PendingSurveillanceEntity pr) {
 				surv.getRequirements().add(req);
 			}
 		}
+		
+		if(pr.getValidation() != null && pr.getValidation().size() > 0) {
+			for(PendingSurveillanceValidationEntity validation : pr.getValidation()) {
+				if(validation.getMessageType() == ValidationMessageType.Error) {
+					surv.getErrorMessages().add(validation.getMessage());
```
andlar: There were two types of message types added in the data-model PR, right? Warning and Error? Should this section have something like:

    if (validation.getmessagetype == validationtype.warning)
        surv.getwarningmessages.add(validation.getmessage)
?

kekey1: Yes, you would have thought so, except our existing Surveillance object ONLY had a field for error messages. I can add a field for warning messages if you think it would be better to have and an else-if here, but it would currently never be used. I left the db and enum ValidationMessageTypes as having both error and warning to keep them general if we want to use them for pending certified products or something else later.

andlar: Sounds good to me.

andlar: Err, leaving it exactly as it is, that is

PR #282, opened 2017-07-27
Development

kekey1:


chpl/chpl-service/src/main/resources/environment.properties.template
```diff
@@ -7,9 +7,15 @@ basePath=/rest
 questionableActivityEmail=theonc.certification@hhs.gov;onc_chpl@hhs.gov
 questionableActivityThresholdDays=0
 maxActivityRangeInDays=60
+
+#cache properties
 enableCacheInitialization=true
 cacheClearTimeoutSecs=15
 cacheInitializeTimeoutSecs=300
+#how long to wait before first loading the prefetched listing cache
+listingCacheRefreshInitialDelayMillis=300000
```
andlar: Does this line mean that our "wait for the cache to be loaded" step in the tomcat/apache start scripts will have to wait an additional 300 seconds before that cache even starts loading?

kekey1: The search results that are cached immediately upon app startup and get served to the UI have not changed. There are actually two caches of the search results, the one just mentioned and one that is called "pre-fetched" in the code which will get loaded every X seconds. The pre-fetched one was always the one that was getting refreshed whenever something changed and then the existing current search result cache was swapped with the latest pre-fetched results. As I recall, the idea was that we'd always have some cached results that could be used. This code change loads the pre-fetched cache on a timer and the first time the pre-fetched cache gets loaded is after this initialDelayMillis property. We can make it a shorter delay of course but I thought it probably didn't need to start the loading/reloading cycle until some time after the app has started.

andlar: Got it. That sounds like a plan.

Please add these properties into the build files 😄

PR #290, opened 2017-08-24
Ocd 1749

andlar:
Sorry about the whitespace updates; I really tried not to do mass file changes


chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2015.java
```diff
@@ -577,27 +577,29 @@ private void parseTestParticipants(CSVRecord record, PendingCertifiedProductEnti
 			participant.setGender(record.get(colIndex++).trim());
 			String ageStr = record.get(colIndex++).trim();
 			if(StringUtils.isEmpty(ageStr)) {
-				throw new InvalidArgumentsException("An age range is required for participant " + participant.getUniqueId());
-			}
-			participant.setUserEnteredAge(ageStr);
-			AgeRangeDTO ageDto = ageDao.getByName(ageStr);
-			if(ageDto != null) {
-				participant.setAgeRangeId(ageDto.getId());
+                logger.error("Age range is empty.");
```
kekey1: I forget, but is the participant age_range_id a required not null field in the db? If it has to be not null but is null because the user-entered value doesn't match an accepted one, I think that will throw some kind of exception with a poor error message. Same for education level. Did you test that to see if removing this bit of code and entering a bad age range still works appropriately?

PR #296, opened 2017-08-29
OCD-1540 - retrieve entire family tree of listings

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductDetailsManager.java
```diff
@@ -1,10 +1,14 @@
...
+import java.util.ArrayList;
+
+import gov.healthit.chpl.dao.EntityRetrievalException;
+import gov.healthit.chpl.domain.CertifiedProductSearchDetails;
+
+public interface CertifiedProductDetailsManager {	
+	public CertifiedProductSearchDetails getCertifiedProductDetails(Long certifiedProductId) 
+			throws EntityRetrievalException;
+
+	ArrayList getIcsFamilyTree(Long certifiedProductId) throws EntityRetrievalException;
```
kekey1: Is this method supposed to have the default access or should it say public/private/protected? Elsewhere in the code we have generic List<Type> or Set<Type> return types rather than the concrete ArrayList and HashSet. You should at least say ArrayList<Type> but to be consistent with everywhere else it should be List<Type>.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/CertifiedProductDetailsManagerImpl.java
```diff
@@ -1,599 +1,645 @@
...
+	}
+	
+	@Override
+	@Transactional
+	public ArrayList<IcsFamilyTreeNode> getIcsFamilyTree(Long certifiedProductId) throws EntityRetrievalException {
+		
+		ArrayList<IcsFamilyTreeNode> familyTree = new ArrayList<IcsFamilyTreeNode>();
+		Map<Long,Boolean> queue = new HashMap<Long,Boolean>();
+		ArrayList<Long> toAdd = new ArrayList<Long>();
+
```
kekey1: Similar comment as above - the <Map> is consistent with the CHPL coding style but ArrayList<> lines are not. I would expect something like 
List<IcsFamilyTreeNode> familyTree = new ArrayList<IcsFamilyTreeNode>();

brianlindsy:
I agree that the logic is a little weird. The problem I was running into was that I needed to remove elements from the queue while I was iterating over it and ran into concurrent modification exception problems. So instead of removing elements I just marked them completed. There is likely to be a more simplistic way if I take another look at it though.

andlar:
Yeah, I can see the problem. The multiple loops things likely won't be too much of a process drain, as we're likely to have small lists, and you bail out of the for loop early for ones that were already handled. I'm good with it for now; if it starts looking like an issue we can revisit later

andlar:
I had thought I'd approved it with my first comment; oops. :)

PR #300, opened 2017-09-13
Ocd 1540 - ICS family tree with cached search call

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/search/CertifiedProductSearchDAOImpl.java
```diff
@@ -130,4 +151,47 @@ public CertifiedProduct getByChplProductNumber(String chplProductNumber)
...
+		prod.setName(result.getProduct());
+		node.setProduct(prod);
+		ProductVersion pv = new ProductVersion();
+		pv.setVersion(result.getVersion());
+		node.setVersion(pv);
+		ArrayList<CertifiedProduct> childrenList = new ArrayList<CertifiedProduct>();
+		if(result.getChild() != null){
+			String[] children = result.getChild().split("☺");
+			for(String child : children){
+				String[] childInfo = child.split("☹");
```
andlar: Is it possible these characters are set somewhere in properties files or something? If they are, you should use them here.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationStatus.java
```diff
@@ -16,9 +16,9 @@
 	
 	private String name;
 	
-	public CertificationStatus(Map<String, Object> certStatus){
-		this.id = Long.valueOf(certStatus.get("id").toString());
-		this.name = certStatus.get("name").toString();
+	public CertificationStatus(String name){
+		this.id = id;
+		this.name = name;
```
andlar: Why is this file changing? I'm concerned that changing domain objects will have far-reaching effects. As well, there's no `id` that `this.id` can be set to in this method

chpl/chpl-service/src/test/java/gov/healthit/chpl/manager/impl/CertifiedProductDetailsManagerTest.java
```diff
@@ -34,6 +34,7 @@
 import gov.healthit.chpl.domain.InheritedCertificationStatus;
 import gov.healthit.chpl.domain.MacraMeasure;
 import gov.healthit.chpl.manager.CertifiedProductDetailsManager;
+import gov.healthit.chpl.manager.CertifiedProductManager;
```
andlar: This is the CPDManager test, not the CPManager test. These tests should be in a different file

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationStatus.java
brianlindsy: I created this domain object for the ICSFamilyTreeNode object, its not used anywhere else. I will get rid of the id though as that's not being used anymore.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/search/CertifiedProductSearchDAOImpl.java
brianlindsy: I cannot find anywhere in the code where they are currently properties, but I can add them as properties if they should be.

andlar: If they're not already there's no need. If, in the future, we find ourselves needing to change them then we can extract it

PR #309, opened 2017-09-26
Ocd 1731

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/statistics/ChartDataDAOImpl.java
```diff
@@ -0,0 +1,183 @@
...
+		ChartDataEntity entity = null;
+
+		String queryStr = "from ChartDataEntity "
+				+ "where (chart_data_id = :entityid)";
+		Query query = entityManager.createQuery(queryStr, ChartDataEntity.class );
+		query.setParameter("entityid", entityId);
+		List<ChartDataEntity> result = query.getResultList();
+
+		if (result.size() > 1){
+			throw new EntityRetrievalException("Data error. Duplicate certificaiton body id in database.");
```
andlar: Typo on `certificaiton`; should be `certification`

kekey1: I have been pondering adding code to the update call because the tight coupling of putting code in there to fix all the statistics was not sitting well. Especially as we talked about adding other types of statistics that might be affected by various other things being updated. Can we implement this in a different way using a custom Spring event listener? Here's an example tutorial although we don't do this anywhere else so I don't know for sure that it will work with our versions of everything. https://www.tutorialspoint.com/spring/custom_events_in_spring.htm Basically, the CP Controller will become a class that can broadcast events and when they broadcast a custom ListingUpdatedEvent there will be a ChartDataUpdateHandler that listens for that event (and other events in the future) and handles it appropriately by going back in time and updating all that chart data. If that programming pattern works out we might be able to apply it to some other things, too.

andlar: That sounds promising. Definitely don't like the high-coupling issue. @brianlindsy does that make sense to you? Would you like some of @kekey1 's help with this?

I know we're running way long on the task, and feel like we should double-down on this one and get it done

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/GenerateDailyChartData.java
```diff
@@ -0,0 +1,200 @@
...
+import org.apache.logging.log4j.Logger;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.FilterType;
+import org.springframework.context.support.AbstractApplicationContext;
+import org.springframework.stereotype.Component;
+
+@Component("generateDailyChartData")
+public class GenerateDailyChartData {
```
kekey1: There is a base class App that I think you could benefit from inheriting from. It will read the properties for you and has an abstract method where you could load your spring beans. Will probably reduce some of the setup code that all the apps need to do.

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/GenerateDailyChartData.java
```diff
@@ -0,0 +1,200 @@
...
+		List<Long> certifiedIdsWithChangedHistory = app.getListingStatisticsDao().getProductsWithChangedCertificationStatusPostApril2016(date1);
+		System.out.println(certifiedIdsWithChangedHistory.size());
+		for(Long id : certifiedIdsWithChangedHistory){
+			List<CertificationStatusEventDTO> events = app.getCertificationStatusEventDao().findByCertifiedProductId(id);
+			for(int i=0;i<events.size()-1;i++){
+				if(events.get(i).getEventDate().after(date1)){
+					Date beginDate = events.get(i).getCreationDate();
+					Date endDate = events.get(i+1).getCreationDate();
+					// if going from status 1 to something else decrement all dates in between begin and end
+					if(events.get(i).getStatus().getId().equals(1L)){
```
kekey1: It might just be that I'm not as familiar with this code but I'm not clear on exactly what this is doing (I can see it's loading daily statistics but I'm more lost with the getting and setting of the list variable below). Could you add a comment block to explain a little bit? Or maybe use some helpfully-named constants here instead of 0, 1, 3, 4?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/ChartDataController.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@Api(value="data")
+@RestController
+@RequestMapping(value="/data")
```
kekey1: In the SearchViewController, there are many URLs starting with /data/* and they are all lookup type things (like /data/age_ranges to get the list of available age ranges in the system). I think that this controller might be better with a root @RequestMapping of /statistics; you could move /data/statistics_types into the SearchViewController and use one of the KeyValueModel objects similar to the other calls in there which has fields for name and ID.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/StatisticsController.java
```diff
@@ -0,0 +1,13 @@
...
+
+import io.swagger.annotations.Api;
+
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@Api(value="data")
+@RestController
+@RequestMapping("/data")
+public class StatisticsController {
```
kekey1: Is this class empty? Is it needed?

PR #310, opened 2017-09-29
API consistency changes

kekey1:


kekey1:
Yeah, I couldn't figure that out. The auth project needs it for a unit test to pass and the service project needs it because the auth project needs it. I can think about it some more and try to come up with a solution.

andlar:
I wouldn't worry about it now. If we end up with a lot of duplication that will be when we should take another look

PR #323, opened 2017-10-31
Questionable Activity changes

kekey1:
Mainly fixed a typo in a package name and complete changed around the CSV output. Once Dustin saw the sample output based on our original idea in the ticket he asked for it to be different.


kekey1:
I did add the new sample CSV to the ticket!

PR #327, opened 2017-11-10
New search API

kekey1:
I probably won't be able to address comments until late today or Monday but wanted to get this out there!


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/search/SearchRequest.java
```diff
@@ -0,0 +1,276 @@
...
+    String practiceType = null;
+    String certificationDateStart = null;
+    String certificationDateEnd = null;
+    SurveillanceSearchFilter surveillance = new SurveillanceSearchFilter();
+    
+    String orderBy = ORDER_BY_PRODUCT;
+    Boolean sortDescending = false;
+    Integer pageNumber = 0;
+    Integer pageSize = 20;
+
```
andlar: Do the pageNumber/pageSize parameters need to be cleaned / checked for "isNaN" status?

kekey1: Page size is checked for not being greater than 100. Because they are declared as Integers in this object and in the GET request params, spring will automatically send back a bad request error if the user passes in something it can't be converted to integer.

PR #332, opened 2017-11-21
Ocd 1904 - consolidate acb, atl roles

brianlindsy:


RELEASE_NOTES.md
```diff
@@ -1,5 +1,13 @@
 # Release Notes
 
+## Version TBD
+
+### Backwards compatability breaking features
+* Change ROLE_ACB_ADMIN to ROLE_ACB
+* Change ROLE_ATL_ADMIN to ROLE_ATL
+* Remove ROLE_ACB_STAFF
+* Remove ROLE_ATL_STAFF
+
```
andlar: This section is missing the `---` section break between the "Version TBD" and "Version 15.0.0" section. Also needs the `_Date TBD_` part

kekey1: Do these changes really break backwards compatibility? I'm trying to think of which API calls would no longer work.

PR #347, opened 2017-12-28
OCD-1861

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2014Version1.java
```diff
@@ -64,7 +62,33 @@ public TemplateColumnIndexMap getColumnIndexMap() {
     }
     
     public String[] getCriteriaNames() {
-        return this.criteriaNames;
+    	criteriaList.put("a", a);
```
kekey1: I feel like this sacrificed some readability and I'm not sure what the benefit is instead of just writing out the criteria names the way they were. If it's important for some other change in the code that's fine and I wouldn't hold up the PR for it but it's definitely harder to understand!

andlar: Agreed. In two months we'll have no idea what these are doing without having to re-read through way too much stuff

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/CertifiedProduct2014Validator.java
```diff
@@ -89,49 +238,88 @@ public void validate(PendingCertifiedProductDTO product) {
...
-        for (PendingCertificationResultDTO cert : product.getCertificationCriterion()) {
             for (int i = 0; i < g3ComplementaryCerts.length; i++) {
                 if (cert.getNumber().equals(g3ComplementaryCerts[i]) && cert.getMeetsCriteria()) {
                     hasG3Complement = true;
                 }
             }
         }
-
+        if (!hasG4) {
+            product.getErrorMessages().add("Required certification criteria 170.314 (g)(4) was not found.");
```
andlar: Should be an error in the properties file.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/CertifiedProduct2014Validator.java
```diff
@@ -89,49 +238,88 @@ public void validate(PendingCertifiedProductDTO product) {
...
+                }
+            }
+
+            if (!hasG2Complement) {
+                product.getWarningMessages().add("(g)(2) was found without a required related certification.");
+            }
+        }
+
+        if (hasG1Cert && hasG2Cert) {
+            product.getWarningMessages().add("Both (g)(1) and (g)(2) were found which is not typically permitted.");
```
andlar: Should be in the errors.properties file

chpl/chpl-service/src/main/resources/errors.properties
```diff
@@ -223,3 +269,46 @@ user.notFound=User not found.
 #Warnings
 transparencyAttestation.save=Warning: The transparency attestation for the developer is different in the system than in the upload file. To save the updated transparency attestation, as well as all other developer information, please click 'Save as Developer Information' above before proceeding.
 
+uniqueCHPLId=255
```
andlar: Needs a comment above to indicate what this section is for, and possibly a prefix to indicate what type of thing we're looking at

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/CertifiedProductValidatorImpl.java
```diff
@@ -99,6 +99,18 @@
...
+    	return Integer.parseInt(String.format(
+    			messageSource.getMessage(new DefaultMessageSourceResolvable(field),
+    			LocaleContextHolder.getLocale())));
+    }
+    
+    public String getMaxLengthErrorMessage(String errorField){
+    		return String.format(
+    				messageSource.getMessage(new DefaultMessageSourceResolvable(errorField),
+    				LocaleContextHolder.getLocale()));
+    }
```
andlar: These should be combined. All of the uses are of the form:
```
if value != null && check length of value is bad then
add error message (from this function)
```
The `checkMaxLength` function is never used without the `getMaxLengthMessage` function, so they should be combined, something like:
```
public checkLength(product, field, errorField)
if field != null && field.length > Integer.parseInt('maxlength.' + propertiesFile.errorField)
product.addErrorMessage(String.format...(getMessage('product.' + errrorField + 'length'))
``` 
up here, and in the checks below, something like (for the certification edition):
```
checkLength(product, product.certificationEdition, 'certificationEdition');
```

PR #353, opened 2018-01-11
Ocd 1862

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler.java
```diff
@@ -297,7 +297,32 @@ protected PendingCertificationResultEntity getCertificationResult(String criteri
         result.setMeetsCriteria(asBoolean(columnValue));
         return result;
     }
+    
+    protected Boolean asBooleanEmpty(String value) {
+        value = value.trim();
+
+        if (StringUtils.isEmpty(value)) {
+            return null;
+        }
```
andlar: Should this return `false`? I thought we have the theory that blank counts as false, in general.

brianlindsy: This is a new method that is only used to parse G1 and G2 success fields right now, for those fields they are required to enter something, either 0 or 1. For us to detect whether or not they entered something we had to create this new method that set empty to null so that we can detect if those fields are actually filled in and not just set that field to false.

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/listing/pending/PendingCertifiedProductEntity.java
```diff
@@ -204,7 +204,8 @@
     @Column(name = "pending_certified_product_id", nullable = false)
     private Set<PendingCertifiedProductParentListingEntity> parentListings;
     
-    @Transient
+    @Basic(optional = false)
+    @Column(name = "has_qms", nullable = false)
     private boolean hasQms;
```
kekey1: There should be a data model PR to match this - where is it? It's why I suggested you push your data model changes to Jenkins via a new job b/c your build was failing due to missing this column.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/CertifiedProduct2014Validator.java
```diff
@@ -58,7 +61,7 @@ public String getErrorMessage(String errorField){
     }
     public String getErrorMessage(String errorField, String input){
     	return String.format(messageSource.getMessage(
-                new DefaultMessageSourceResolvable("listing.criteria.badG1MacraMeasure"),
+                new DefaultMessageSourceResolvable(errorField),
                 LocaleContextHolder.getLocale()), input);
     }
```
kekey1: This method looks useful and like something that's not specific to 2014 listings and could be in whatever the superclass of all validators is.

PR #357, opened 2018-01-17
chore: Removing "deployment specific" file

andlar:
This is not a file that's part of CHPL, but is instead a thing that is deployment specific, so, if it should be anywhere, should be in the chpl/resources project. Someone cloning CHPL wouldn't be able to use this at all, and that means it shouldn't be here. If you want to put it in your local environment that's okay, but it should be .gitignored, like my `deploy-webapp.sh` script. I have these commands in a file in my personal docs directory, and just copy/paste them as needed.


andlar:
Great minds... ;)

PR #358, opened 2018-01-17
Ocd 1946

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/FuzzyChoicesDAOImpl.java
```diff
@@ -0,0 +1,120 @@
...
+    private FuzzyChoicesEntity getEntityByType(FuzzyType type)
+        throws EntityRetrievalException {
+
+        FuzzyChoicesEntity entity = null;
+
+        String queryStr = "SELECT fuzzy from FuzzyChoicesEntity fuzzy where "
+            + "(fuzzy_choices_id = :entityId)";
+
+        Query query = entityManager.createQuery(queryStr, FuzzyChoicesEntity.class);
+        query.setParameter("entityId", (type.ordinal()+1));
```
kekey1: I don't think this is quite the right way to do this... it requires your java enum values to be in the same order as your database enum values, which they are so it will work but.... . Do you have to get the fuzzy type by ID? Maybe you could use the name?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/PendingCertifiedProductSystemUpdateDAOImpl.java
```diff
@@ -0,0 +1,45 @@
...
+        return new PendingCertifiedProductSystemUpdateDTO(entity);
+	}
+	
+	private void create(PendingCertifiedProductSystemUpdateEntity announcement) {
+
+        entityManager.persist(announcement);
+        entityManager.flush();
+    }
+
+    private void update(PendingCertifiedProductSystemUpdateEntity announcement) {
```
kekey1: Is this method used anywhere? If not, remove it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/FuzzyChoicesManagerImpl.java
```diff
@@ -0,0 +1,92 @@
...
+			e.printStackTrace();
+		}
+		String result = null;
+    	for(ExtractedResult er : results){
+    		result = er.getString();
+    	}
+    	if(result != null){
+    		PendingCertifiedProductSystemUpdateEntity entity = new PendingCertifiedProductSystemUpdateEntity();
+    		entity.setChangeMade("Changed " + type.toString() + " name from " + query + " to " + result);
+    		entity.setPendingCertifiedProductId(product.getId());
```
kekey1: I feel like it would have been better to make the table store a previous value and changed value so that you could generate this sentence in the business logic of the API rather than save it in the database.

PR #364, opened 2018-01-29
OCD-2025: add pending reference to confirmed listings

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/listing/CertifiedProductEntity.java
```diff
@@ -125,6 +125,10 @@
     @Column(name = "product_additional_software")
     private String productAdditionalSoftware;
 
+    @Basic(optional = false)
```
andlar: Is this optional? I would think it has to be to support legacy listings that don't have references

PR #401, opened 2018-05-07
Fix downloadable file generation issue

andlar:
Mostly whitespace changes. Only actual code change is the last part, and should fix the fact that DEV and STG haven't had new downloadable files for a few days. Hopefully will also fix it in PROD before we break it tonight


andlar:
I thought about RELEASE_NOTES, but don't think there's anything here that's changing anything externally visible, which suggests no need for notes.

PR #408, opened 2018-05-10
OCD-2035: Criterion/Product Chart

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/CriterionProductStatisticsDAOImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    private CriterionProductStatisticsEntity getEntityById(final Long id) throws EntityRetrievalException {
+        CriterionProductStatisticsEntity entity = null;
+
+        Query query = entityManager.createQuery(
+                "from CriterionProductStatisticsEntity a where (NOT deleted = true) AND (id = :entityid) ",
+                CriterionProductStatisticsEntity.class);
+        query.setParameter("entityid", id);
+        List<CriterionProductStatisticsEntity> result = query.getResultList();
+
+        if (result.size() > 1) {
```
kekey1: I think these bits of code that check if result.size() > 1 are in almost all the DAOs and they are from copying and pasting from Watson's days but I always thought they were kind of odd. How can there be more than one record with the same primary key? Not saying to remove because maybe it is a valid check, I just always thought they were weird!

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/CriterionProductStatisticsEntity.java
```diff
@@ -0,0 +1,130 @@
...
+    @Column(name = "id", nullable = false)
+    private Long id;
+
+    @Basic(optional = false)
+    @Column(name = "product_count", nullable = false)
+    private Long productCount;
+
+    @Basic(optional = false)
+    @Column(name = "certification_criterion_id", nullable = false)
+    private Long certificationCriterionId;
```
kekey1: Do you also want to join in the Certification Criterion here so it's easy to pull out the "170.315"-ish name? That must be done elsewhere and I just missed it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/CriterionProductStatisticsDAOImpl.java
andlar: Fair. I can see some use for this kind of check in entities based on views, as we do sometimes have more than one entity in a view, but certainly seems odd for a query on a primary key. What do you think it ought to do? Just assume the "first" one is the right one? Or do we leave it as is?

kekey1: I guess I think the first one should also be the only one. Probably just leave it as it is though since it's a bit of code that's in most other places and not hurting anything. I just wanted to mention it!

andlar: Fair enough.

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/CriterionProductStatisticsEntity.java
andlar: I am retrieving the name somewhere, but I can't recall where. I had the impression entities were supposed to match the database table structure, which doesn't have the name in it. Am I mistaken with that?

andlar:
I had thought about test cases, but then completely forgot about them. I did test it "in the real world" and it worked, so I'm thinking now I'll merge this as it is (as I just did the data-model and UI), get it deployed, and then spend some time on test cases.

PR #413, opened 2018-05-21
New vs. Incubment Developer charts

andlar:


PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,11 @@
+
+## Version 15.13.0
+_Date TBD_
+
+### Minor Features
+* Add new chart: New vs. Incumbent Developer chart
```
kekey1: missing space, also since we did the release this should just be one PENDING right?

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/chartdata/IncumbentDevelopersStatisticsCalculator.java
```diff
@@ -0,0 +1,166 @@
...
+            case "2015":
+                developers2015.add(listing.getDeveloper());
+                break;
+            default:
+                LOGGER.info("Listing has no edition");
+            }
+        }
+
+        IncumbentDevelopersStatisticsEntity from2011To2014 = new IncumbentDevelopersStatisticsEntity();
+        from2011To2014.setOldCertificationEditionId(1L);
```
kekey1: Can you get these IDs from the database instead of hardcoding?

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/chartdata/IncumbentDevelopersStatisticsCalculator.java
```diff
@@ -0,0 +1,166 @@
...
+            protected void doInTransactionWithoutResult(final TransactionStatus arg0) {
+                try {
+                    deleteExistingIncumbentDevelopersStatistics();
+                } catch (EntityRetrievalException e) {
+                    LOGGER.error("Error occured while deleting existing CriterionProductStatistics.", e);
+                    return;
+                }
+                try {
+                    for (IncumbentDevelopersStatisticsEntity entity : entities) {
+                        IncumbentDevelopersStatisticsDTO dto = new IncumbentDevelopersStatisticsDTO(entity);
```
kekey1: I think it's a bit repetitive here to have a list of entities, convert each one to a DTO, then in the dao create method they get converted back to an entity for saving to the db. It was always my understanding (from whenever Watson introduced me to the code, I think) that we were supposed to keep the entities used primarily only in the DAOs and everywhere else we should pass around DTO objects... maybe some of the other code in this class could be changed to use DTOs and then it would just get converted to an entity once inside the DAO?

PR #415, opened 2018-05-22
Ocd 1790 - Convert API REST calls to use PUT/DELETE instead of /create /update /delete

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/CertificationIdController.java
```diff
@@ -148,7 +170,8 @@
     //
     // Verify whether one or more specific EHR Certification ID is valid or not.
     // **********************************************************************************************************
-    @ApiOperation(value = "Verify whether one or more specific EHR Certification IDs are valid or not.",
+    @Deprecated
+    @ApiOperation(value = "DEPRECATED.  Verify whether one or more specific EHR Certification IDs are valid or not.",
```
andlar: I don't think we can deprecate this call. It's used by CMS to verify a huge number of IDs all at once, and I think they're sending more data than is allowed in the RequestHeaders of a GET, which is why we built it as a POST. Up for a conversation

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
```diff
@@ -230,17 +253,40 @@ public void streamDocumentContents(@PathVariable("documentId") Long documentId,
...
+        return createNonconformityDocumentForSurveillance(surveillanceId, nonconformityId, file);
+    }
+
+    @ApiOperation(value = "Add documentation to an existing nonconformity.",
+            notes = "Upload a file of any kind (current size limit 5MB) as supporting "
+                    + " documentation to an existing nonconformity. The logged in user uploading the file "
+                    + " must have either ROLE_ADMIN or ROLE_ACB and administrative "
+                    + " authority on the associated ACB.")
+    @RequestMapping(value = "/{surveillanceId}/nonconformity/{nonconformityId}/document",
+            method = RequestMethod.POST, produces = "application/json; charset=utf-8")
```
andlar: This might be a PUT. As in, the intent of the API call is to update a specific Surveillance "thing". That said, we might have to keep it as a POST anyway, as I think file uploads with the Angular tool we have doesn't support PUT. Maybe we have both?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
```diff
@@ -378,15 +471,35 @@ public void streamDocumentContents(@PathVariable("documentId") Long documentId,
...
             @PathVariable("docId") Long docId) throws JsonProcessingException, EntityCreationException,
             EntityRetrievalException, InvalidArgumentsException {
 
+        return deleteNonconformityDocument(surveillanceId, docId);
+    }
+
+    @ApiOperation(value = "Remove documentation from a nonconformity.",
+            notes = "The logged in user" + " must have either ROLE_ADMIN or ROLE_ACB and administrative "
+                    + " authority on the associated ACB.")
+    @RequestMapping(value = "/{surveillanceId}/document/{docId}", method = RequestMethod.DELETE,
```
andlar: Similarly as with the above "add document to nonconformity". This might be a PUT as it's conceptually an edit of that nonconformity

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/TestingLabController.java
```diff
@@ -159,20 +193,36 @@ public TestingLab create(@RequestBody TestingLab atlInfo) throws InvalidArgument
         address.setZipcode(atlInfo.getAddress().getZipcode());
         address.setCountry(atlInfo.getAddress().getCountry());
         toUpdate.setAddress(address);
-
+
```
andlar: There are some random whitespaces issues through these files that I've not been flagging. If you have a chance to clear them that's great, but not required. Checkstyle should have been flagging some of them

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -157,14 +163,33 @@ public User confirmUser(@RequestBody String hash) throws InvalidArgumentsExcepti
...
+            throws InvalidArgumentsException, JWTCreationException, UserRetrievalException, EntityRetrievalException {
+
+        return authorize(credentials);
+    }
+
+    @ApiOperation(value = "Update an existing user account with new permissions.",
+            notes = "Adds all permissions from the invitation identified by the user key "
+                    + "to the appropriate existing user account." + "The correct order to call invitation requests is "
+                    + "the following: 1) /invite 2) /create or /authorize 3) /confirm ")
+    @RequestMapping(value = "/{userId}/authorize", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE,
```
andlar: Should this be a PUT? We're editing a user's permissions with this call, right?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -310,21 +373,44 @@ public String deleteUser(@PathVariable("userId") Long userId)
...
             UserRetrievalException, UserManagementException, UserPermissionRetrievalException, JsonProcessingException,
             EntityCreationException, EntityRetrievalException {
+        
+        return grant(grantRoleObj);
+    }
+
+    @ApiOperation(value = "Give additional roles to a user.",
+            notes = "Users may be given ROLE_ADMIN, ROLE_ACB, "
+                    + "ROLE_ATL, or ROLE_ONC_STAFF roles within the system.")
+    @RequestMapping(value = "/{userName}/roles/{roleName}", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE,
```
andlar: Isn't this an edit of a User also? So a PUT?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -335,34 +421,59 @@ public String grantUserRole(@RequestBody GrantRoleJSONObject grantRoleObj) throw
...
             UserRetrievalException, UserManagementException, UserPermissionRetrievalException, JsonProcessingException,
             EntityCreationException, EntityRetrievalException {
 
+        return revoke(grantRoleObj);
+    }
+    
+    @ApiOperation(value = "Remove roles previously granted to a user.",
+            notes = "Users may be given ROLE_ADMIN, ROLE_ACB, "
+                    + "ROLE_ATL, or ROLE_ONC_STAFF roles within the system.")
+    @RequestMapping(value = "/{userName}/roles/{roleName}", method = RequestMethod.DELETE,
```
andlar: Would think this is a PUT also, as it's not deleting the user, but it is editing it

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -7,5 +7,7 @@ _Date TBD_
   * Update chart data generation application
   * Add API endpoint to retrieve chart data
   * Refactored other Chart data generation to increase speed
+* Updated URLs and verbs for several REST endpoints
+	* Old endpoints have been deprecated
```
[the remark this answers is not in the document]
tmy1313: I'll make this change.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/CertificationIdController.java
tmy1313: I thought that this was only used for a few IDs.  Should I remove the new version or just remove the deprecated annotation for the original method?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/SurveillanceController.java
tmy1313: I felt this was a post because it was not an idempotent transaction.  Each time you upload the same file, you get another file associated to the nonconformity.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/TestingLabController.java
tmy1313: I'll take a look at this.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
tmy1313: This creates a relationship between a user and role, therefore I think it should be a POST.

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
tmy1313: This deletes the relationship between a user and role, therefore  I think it should be a delete.

tmy1313: Plus, if this is a PUT and the one above is a PUT, it would be ambiguous.

PR #419, opened 2018-05-29
Fixed issue with duplicate CHPL Prod Nbrs when merging/changing devs

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -362,6 +390,71 @@ public DeveloperDTO merge(List<Long> developerIdsToMerge, DeveloperDTO developer
         return createdDeveloper;
     }
 
+    private String getDuplicateChplProductNumberErrorMessage(
+            final List<DuplicateChplProdNumber> duplicateChplProdNumbers) {
+
+        String message = "";
+
+        for (DuplicateChplProdNumber dup : duplicateChplProdNumbers) {
+            message += "\n" + dup.toString();
```
andlar: Needs to use system-specific line separator. I think `\n` is for Linux and `\r\n` is for Windows, but there's some Java property to handles it. That said, I'm not sure this is a good idea anyway, as I don't know if it'll display either of those things correctly in all browsers. There you'd need `<br>`, but that doesn't make sense in an API response at all. Might make more sense to put multiple error messages in the response; one for each pair of duplicates. The UI already handles that kind of thing correctly

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -431,4 +524,52 @@ private boolean isStatusHistoryUpdated(DeveloperDTO original, DeveloperDTO chang
...
     }
+
+    private class DuplicateChplProdNumber {
+        private String origChplProductNumberA;
+        private String origChplProductNumberB;
+        private String newChplProductNumber;
+
+        public DuplicateChplProdNumber(final String origChplProductNumberA, final String origChplProductNumberB,
+                final String newChplProductNumber) {
+            super();
```
andlar: This class doesn't extend anything, right? So what does `super()` do here?

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/ProductController.java
```diff
@@ -237,6 +266,83 @@
         return new ResponseEntity<Product>(new Product(updatedProduct), responseHeaders, HttpStatus.OK);
     }
 
+    private String getDuplicateChplProductNumberErrorMessage(
+            final List<DuplicateChplProdNumber> duplicateChplProdNumbers) {
+        String message = "";
+
+        for (DuplicateChplProdNumber dup : duplicateChplProdNumbers) {
+            message += "\n" + dup.toString();
```
andlar: Same issue with `\n` as above

chpl/chpl-service/src/main/java/gov/healthit/chpl/web/controller/ProductController.java
```diff
@@ -316,4 +427,52 @@
         }
         return new ResponseEntity<SplitProductResponse>(response, responseHeaders, HttpStatus.OK);
     }
+
+    private class DuplicateChplProdNumber {
```
kekey1: Should this be a separate class since it's used in both developer and product controllers? It's hard to tell in this view but I think they are identical?

PR #427, opened 2018-06-06
Scheduler: Ready for review

andlar:
I know I'm missing RELEASE NOTES, and getting the latest from upstream, but would like to give everyone a chance to check out these branches and try it on your own ends to test it out.

Still need to:
 * Remove old Notification
 * Get a better cron schedule input
 * Migrate old subscribers

That said, I think it's functioning


chpl/chpl-scheduler/pom.xml
```diff
@@ -0,0 +1,421 @@
...
+      <artifactId>spring-oxm</artifactId>
+      <version>4.2.2.RELEASE</version>
+    </dependency>
+    <dependency>
+      <groupId>com.thoughtworks.xstream</groupId>
+      <artifactId>xstream</artifactId>
+      <version>1.4.8</version>
+    </dependency>
+    <dependency>
+      <groupId>io.springfox</groupId>
```
kekey1: Just wondering if you're using all of these dependencies? Probably not swagger, at least.

PR #428, opened 2018-06-07
Ocd 1503 - Updated to support test functionalities tied to practice type 

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/TestFunctionalityEntity.java
```diff
@@ -36,6 +36,10 @@
     @JoinColumn(name = "certification_edition_id", insertable = false, updatable = false)
     private CertificationEditionEntity certificationEdition;
 
+    @OneToOne(optional = true, fetch = FetchType.EAGER)
```
andlar: I don't think I've seen "EAGER" before. Most of the others use "LAZY". What's the difference? Why one or the other?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/CertifiedProduct2014Validator.java
```diff
@@ -219,6 +246,20 @@ public void validate(final PendingCertifiedProductDTO product) {
...
+                pt.setId(product.getPracticeTypeId());
+                pt.setName(product.getPracticeType());
+
+                product.getErrorMessages().addAll(
+                        getPendingTestFunctionalityValidationErrors(
+                                EDITION_ID,
+                                pt,
+                                product.getCertificationCriterion()));
+            } catch (EntityRetrievalException e) {
+                //TODO - what to do here???
```
andlar: I think we usually just throw it, and shrug.

chpl/chpl-service/src/main/resources/errors.properties
```diff
@@ -139,6 +139,7 @@ listing.criteria.badParticipantProfessionalExperienceMonths=Professional Experie
 listing.criteria.badParticipantProductExperienceMonths=Product Experience Months is required for participant %s.
 listing.criteria.badParticipantComputerExperienceMonths=Computer Experience Months is required for participant %s.
 listing.criteria.fuzzyMatch=Changed %s for criteria %s from %s to %s.
+listing.criteria.testFunctionalityMisMatch=In Criteria %s, Test Functionality %s is not valid for Practice Type %s.
```
andlar: There was a request to add "...something is for something settings only and is not valid...", right?

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/TestFunctionalityEntity.java
kekey1: I usually do Lazy fetches and wherever the test functionality entity is being gotten in the dao i'll add "join fetch" code to also get the practiceType entity. I don't like hibernate doing extra queries behind the scenes which I think is what Eager fetches have the potential to do?

PR #432, opened 2018-06-13
SED business rule mismatch

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/certifiedProduct/CertifiedProductValidatorImpl.java
```diff
@@ -321,7 +321,7 @@ public boolean validateProductCodeCharacters(final String chplProductNumber) {
     @Override
     public boolean validateVersionCodeCharacters(final String chplProductNumber) {
         String[] uniqueIdParts = chplProductNumber.split("\\.");
-        if (uniqueIdParts != null && uniqueIdParts.length == CertifiedProductDTO.CHPL_PRODUCT_ID_PARTS) {
+        if (uniqueIdParts.length == CertifiedProductDTO.CHPL_PRODUCT_ID_PARTS) {
```
kekey1: Curious - since the not null check was removed, does this code get called for all listings? What happens with an old-style ID that doesn't have the 9 parts?

andlar: That was a findbugs thing. It's impossible for the result of a string.split call to be null. For a CHPL-123123 listing the split will have a length of 1, as there are no `.`s, so the length check is enough to "not" do the if block

kekey1: Ohhhh good find... makes sense!

PR #436, opened 2018-06-19
Updated api to validate test functionality

tmy1313:
OCD-1503

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
@@ -100,6 +100,9 @@ public class CertificationResult implements Serializable {
 
     @XmlTransient
     private List<MacraMeasure> allowedMacraMeasures;
+    
+    @XmlTransient
+    private List<TestFunctionality> allowedTestFunctionalities;
 
     /**
      * Any optional, alternative, ambulatory (2015 only), or inpatient (2015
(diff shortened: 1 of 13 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
```diff
@@ -100,6 +100,9 @@
 
     @XmlTransient
     private List<MacraMeasure> allowedMacraMeasures;
+    
+    @XmlTransient
+    private List<TestFunctionality> allowedTestFunctionalities;
```
andlar: Will this addition to this file change the XML download file? If so, please add the XSD-MODIFICATION file to that UI project. Would also recommended verifying that our XSD diff step in the DEV plan works correctly when you do build this project

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/TestFunctionality.java
```diff
@@ -52,4 +60,32 @@ public String getYear() {
     public void setYear(final String year) {
         this.year = year;
     }
+
+    /**
+     * @return the certificationCriterion
```
andlar: I'm not sure if we require JavaDoc for getters and setters. Happy to have them, but don't know if they're necessary. If we _do_ have them, they should probably have the right format; I'm specifically thinking of the warning I always get about "the first sentence must end with a period"

PR #440, opened 2018-07-05
Fixed issue with test functionalities and products w/o practice type

tmy1313:
OCD-1503


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/CertifiedProductDetailsManagerImpl.java
```diff
@@ -708,7 +708,12 @@ private CertificationResult getCertificationResult(
             final CertifiedProductSearchDetails cp) {
 
         String edition = cp.getCertificationEdition().get("name").toString();
-        Long practiceTypeId = Long.valueOf(cp.getPracticeType().get("id").toString());
+        Long practiceTypeId = null;
+        if (cp.getPracticeType().containsKey("id")) {
+            if (cp.getPracticeType().get("id") instanceof String) {
```
kekey1: Can practiceType (cp.getPracticeType()) ever be null? I wasn't sure exactly which piece of it was null before.

tmy1313: I have not seen where it could be, but I did find in the unit testing that cp.getPracticeType().get("id") can be null

PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,8 @@
+
+## Version 15.16.0
```
andlar: This is a bug fix, right? Shouldn't the version be 15.15.1?

PR #442, opened 2018-07-10
Fixed duplicate chpl nbr err msg being displayed when editing product

tmy1313:
OCD-2343

### DIFF
--- a/PENDING_RELEASE_NOTES.md
+++ b/PENDING_RELEASE_NOTES.md
@@ -1,0 +1,8 @@
+
+## Version 15.15.2
+_Date TBD_
+
+### Bugs Fixed
+* Fixed bug where user erroneously receives a Duplicate CHPL PRoduct Number error when editting a product
+
+---
--- a/PENDING_RELEASE_NOTES.md
+++ b/PENDING_RELEASE_NOTES.md
@@ -3,6 +3,6 @@
 _Date TBD_
 
 ### Bugs Fixed
-* Fixed bug where user erroneously receives a Duplicate CHPL PRoduct Number error when editting a product
+* Fixed bug where user erroneously receives a Duplicate CHPL Product Number error when editing a product
 
 ---
--- a/chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ProductController.java
+++ b/chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ProductController.java
@@ -324,8 +324,10 @@ public class ProductController {
 
             for (CertifiedProductDetailsDTO filterCp : filteredResults) {
                 //Add it to the list of duplicates
-                duplicateChplProductNumbers.add(new DuplicateChplProdNumber(
-                    cpDTO.getChplProductNumber(), filterCp.getChplProductNumber(), filterCp.getChplProductNumber()));
+                if (!cpDTO.getChplProductNumber().equals(filterCp.getChplProductNumber())) {
+                    duplicateChplProductNumbers.add(new DuplicateChplProdNumber(
+                        cpDTO.getChplProductNumber(), filterCp.getChplProductNumber(), filterCp.getChplProductNumber()));
+                }
             }
         }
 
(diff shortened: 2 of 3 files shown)


kekey1:
@andlar I think there actually IS a unit test that was failing that would (may?) have caught this. gov.healthit.chpl.questionableActivity.ProductTest.testEditOwnerHistoryItem has been failing (was in chpl-service, now chpl-api). I'd like to know if it's passing now with this change in Todd's environment. @tmy1313 can you update your jenkins jobs so we can see about the unit test? You can look at mine or the chpladmin ones - basically we don't need separate chpl-service and chpl-auth jenkins jobs but rather one top-level chpl build that will do the necessary mvn installs and run all the tests in all the sub-modules.

PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,8 @@
+
+## Version 15.15.2
+_Date TBD_
+
+### Bugs Fixed
+* Fixed bug where user erroneously receives a Duplicate CHPL PRoduct Number error when editting a product
```
andlar: P`R`oduct -> P`r`oduct

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ProductController.java
```diff
@@ -324,8 +324,10 @@
 
             for (CertifiedProductDetailsDTO filterCp : filteredResults) {
                 //Add it to the list of duplicates
-                duplicateChplProductNumbers.add(new DuplicateChplProdNumber(
-                    cpDTO.getChplProductNumber(), filterCp.getChplProductNumber(), filterCp.getChplProductNumber()));
+                if (!cpDTO.getChplProductNumber().equals(filterCp.getChplProductNumber())) {
```
andlar: This is a string comparison, right? Don't we usually do string comparisons with some sort of StringUtil operator in some way? @kekey1 am I crazy?

chpl/chpl-api/src/test/java/gov/healthit/chpl/web/controller/ProductControllerTest.java
```diff
@@ -111,4 +112,31 @@ public void testUpdateDuplicateChplProductNumberValidationException() throws Jso
         
         productController.updateProduct(request);
     }
+	
+	@Transactional
+	@Test
+	@Rollback(true)
+	public void testUpdateProductContactInformation() throws EntityRetrievalException, JsonProcessingException, EntityCreationException, InvalidArgumentsException, ValidationException {
```
andlar: Does this test have anything to do with the feature in this PR? Or was it just added for some other reason?

tmy1313: There was no test in this class to test when the product's contact information was being updated, so I added it.  This was the scenario called out in the ticket - updating a product's contact info caused an error.  I wrote the test before making the change, it failed.  Made the change and it passed.  I'm not sure what I did wrong here...

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ProductController.java
kekey1: I think you might be thinking of the StringUtils.isEmpty methods that we use (avoids checking for both null and blank strings - and forgetting one or the other of those possibilities - all over the place). I'm pretty sure that for string comparisons we just use equals or equalsIgnoreCase as applicable.

chpl/chpl-api/src/test/java/gov/healthit/chpl/web/controller/ProductControllerTest.java
andlar: I didn't say you did anything wrong. It's just that the code that changed doesn't seem to have any relationship to contact information changing, and that looked odd. If the test failed initially, then the code was changed, and the test passed, then that's TDD (or at least the first two steps). Unless you see anything that should need refactoring then it's good

PR #444, opened 2018-07-11
Questionable activity on confirm/creation date of listing

kekey1:


PENDING_RELEASE_NOTES.md
```diff
@@ -2,6 +2,9 @@
 ## Version 15.15.2
```
andlar: If a new feature, should probably update the "middle" number instead of the last

chpl/chpl-api/src/main/java/gov/healthit/chpl/CHPLConfig.java
```diff
@@ -66,7 +68,7 @@ public CommonsMultipartResolver getResolver() throws IOException {
         CommonsMultipartResolver resolver = new CommonsMultipartResolver();
 
         // Set the maximum allowed size (in bytes) for each individual file.
-        resolver.setMaxUploadSize(5242880);// 5MB
+        resolver.setMaxUploadSize(MAX_UPLOAD_FILE_SIZE); // 5MB
```
andlar: Suggest moving `//5MB` to the declaration above, and adding something for the cookie age (even if it's just the type of number [seconds vs minutes...])

chpl/chpl-api/src/test/java/gov/healthit/chpl/questionableActivity/CertificationResultTest.java
```diff
@@ -59,20 +61,21 @@
...
-	
 	@BeforeClass
 	public static void setUpClass() throws Exception {
-		adminUser = new JWTAuthenticatedUser();
-		adminUser.setFirstName("Administrator");
-		adminUser.setId(-2L);
-		adminUser.setLastName("Administrator");
-		adminUser.setSubjectName("admin");
-		adminUser.getPermissions().add(new GrantedPermission("ROLE_ADMIN"));
+	    adminUser = new JWTAuthenticatedUser();
```
andlar: Looks like there's a tab here. I only mention it because it sticks out

kekey1:
Well, I thought I could have updated the creation date in our testData file but that will only work temporarily... i.e. if the test properties sets the questionable activity window to 30 days and I change the creation date to be today, then the test that "questionable activity is not recorded when within the window" will work for 30 days but on the 31st day will record activity. I felt I needed a listing that was always going to have a current creation date to be certain it would fall outside of the window. I probably could have just written a little sql snippet at the beginning of the test to change the creation date of one of our listings after it had been loaded by testData.xml and that would have been simpler. May leave the upload/confirm test as-is because I think that's useful anyway and revisit the questionable-activity specific test.

andlar:
In my head it was going to be something like:

```java
public void testQAWindow()
CertifiedProduct cp = getCertifiedProduct(id);
cp.setCreationDate(now());

//trigger questionable activity
expect(no QA to be found)

cp.setCreationDate(a year ago);
//trigger questionable activity
expect(QA to be found)
```

Or even

```java
cp = new MockCP() //using that Mocking thing Todd added, though I have no real idea how to use it :)
cp.setCreationDate(now());

//trigger questionable activity
expect(no QA to be found)

cp.setCreationDate(a year ago);
//trigger questionable activity
expect(QA to be found)
```

PR #445, opened 2018-07-16
Ocd 2352 - nonconformity charts

brianlindsy:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/StatisticsController.java
```diff
@@ -36,6 +38,20 @@
 
     @Autowired
     private StatisticsManager statisticsManager;
+    
+    /**
+     * Retrieves and returns the Listing counts.
+     * @return a JSON representation of an NonconformityTypeStatisticsResult object
+     */
+    @ApiOperation(value = "Get count of nonconformities by citeria.",
```
andlar: "citeria" is misspelled

chpl/chpl-service/pom.xml
```diff
@@ -25,6 +25,12 @@
             <artifactId>chpl-auth</artifactId>
             <version>0.0.1-SNAPSHOT</version>
         </dependency>
+        
+        <dependency>
+            <groupId>gov.healthit.chpl</groupId>
+            <artifactId>chpl-resources</artifactId>
```
andlar: I think this is already in the pom. Shouldn't be there twice

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/chartdata/NonconformityTypeChartCalculator.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.transaction.TransactionStatus;
+import org.springframework.transaction.support.TransactionCallbackWithoutResult;
+import org.springframework.transaction.support.TransactionTemplate;
+
+public class NonconformityTypeChartCalculator {
+	
+	private static final Logger LOGGER = LogManager.getLogger(NonconformityTypeChartCalculator.class);
+	
+	private SurveillanceStatisticsDAO statisticsDAO;
+    private NonconformityTypeStatisticsDAO nonconformityTypeStatisticsDAO;
```
andlar: Weird indentation here. Have you run this through checkstyle?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/statistics/NonconformityTypeStatisticsDAOImpl.java
```diff
@@ -0,0 +1,73 @@
...
+	public List<NonconformityTypeStatisticsDTO> getAllNonconformityStatistics(
+			DateRange dateRange) {
+		String hql = "FROM NonconformityTypeStatisticsEntity WHERE";
+		
+		if(dateRange == null) {
+            hql += " deleted = false";
+		} else {
+			hql += "(deleted = false AND creationDate <= :endDate) "
+                + " OR "
+                + "(deleted = true AND creationDate <= :endDate AND lastModifiedDate > :endDate)";
```
andlar: If line 38 is false, because the creationDate is greater than :endDate, then line 30 can never be true.

Also, all branches of this if chunk have "deleted = false". That should be taken out of the conditional sections entirely.

Why use a "dateRange" but only use the end date of that range?

Why use a dateRange at all? The only mention the ticket makes of dates is on one end because we can only have data from when CHPL started, and on the other as our way of getting out of needing to do real-time stats.

Line 35 should be inside the else block from line 27, because you're checking dateRange == null twice for no particular reason

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/statistics/SurveillanceStatisticsDAOImpl.java
```diff
@@ -148,4 +152,36 @@ public Long getTotalClosedNonconformities(DateRange dateRange) {
...
+	public List<NonconformityTypeStatisticsDTO> getAllNonconformitiesByCriterion(
+			DateRange dateRange) {
+		String hql = "SELECT COUNT(type), type FROM SurveillanceNonconformityEntity WHERE";
+		
+		if(dateRange == null) {
+            hql += " deleted = false GROUP BY type";
+		} else {
+			hql += "(deleted = false AND creationDate <= :endDate) "
+                + " OR "
+                + "(deleted = true AND creationDate <= :endDate AND lastModifiedDate > :endDate) GROUP BY type";
```
andlar: Same comments re: dates and conditional logic as above

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/surveillance/NonconformityTypeStatisticsEntity.java
```diff
@@ -0,0 +1,97 @@
...
+    @Column(name = "id")
+    private Long id;
+    
+    @Basic(optional = false)
+    @Column(name = "nonconformity_count", nullable = false)
+    private Long nonconformityCount;
+    
+    @Basic(optional = false)
+    @Column(name = "nonconformity_type", nullable = false)
+    private String nonconformityType;
```
andlar: What are the options for nonconformityType? Should this be an enum, or a foreign key to some other table?

chpl/chpl-service/src/test/java/gov/healthit/chpl/app/chartdata/NonconformityTypeChartDataAppTest.java
```diff
@@ -0,0 +1,71 @@
...
+		adminUser.setFirstName("Administrator");
+		adminUser.setId(-2L);
+		adminUser.setLastName("Administrator");
+		adminUser.setSubjectName("admin");
+		adminUser.getPermissions().add(new GrantedPermission("ROLE_ADMIN"));
+	}
+
+    @Test
+    @Transactional
+    public void buildNonconformityTypeStatistics() {
```
andlar: This isn't testing anything in the chartdata app, right? Just the DAO methods? If so, should be in a different test file

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/surveillance/NonconformityTypeStatisticsEntity.java
brianlindsy: The nonconformityType in this table is the Criteria name, might be better if it was called criteriaName or something.

kekey1: Is it the same nonconformity_type field as is in the surveillance_nonconformity table? If so, it can be either a criteria name or 170.523 (k)(1) or (k)(2) which are not criteria.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/StatisticsManager.java
```diff
@@ -70,4 +71,6 @@
      */
     List<ParticipantExperienceStatistics> getParticipantExperienceStatisticsResult(Long experienceTypeId);
 
+	List<NonconformityTypeStatistics> getAllNonconformitiesByCriterion();
+
```
kekey1: Indentation is off

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/statistics/NonconformityTypeStatisticsDAOImpl.java
kekey1: I think checking twice is necessary because the first time is to build the HQL string that makes the query object (and you can't make a query object without it), then once you have the query object you need to set the parameters on it so you have to check the same conditions again to know which parameters to set since the query is being built dynamically and can vary.

kekey1: Regarding moving the deleted = false, I misunderstood at first but now agree... since it's in both of the conditional sections it could/should be moved out of those and added to the hql before the if/else checks.

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/surveillance/NonconformityTypeStatisticsEntity.java
andlar: In that other table it's a varchar of some sort, right? My guess is that we should change it to an enum or lookup at some point, but for now keeping this table and that table in sync as both being Strings is fine

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/chartdata/NonconformityTypeChartCalculator.java
```diff
@@ -0,0 +1,65 @@
...
+        txnManager = (JpaTransactionManager) appEnvironment.getSpringManagedObject("transactionManager");
+        txnTemplate = new TransactionTemplate(txnManager);
+    }
+    
+    NonconformityTypeChartCalculator(final SurveillanceStatisticsDAO statisticsDAO,
+    		NonconformityTypeStatisticsDAO nonconformityTypeStatisticsDAO,
+    		TransactionTemplate txnTemplate) {
+    	this.statisticsDAO = (SurveillanceStatisticsDAO) statisticsDAO;
+    	this.nonconformityTypeStatisticsDAO = nonconformityTypeStatisticsDAO;
+        this.txnTemplate = txnTemplate;
```
andlar: More odd indentation stuff. Did you do checkstyle against all of the files you're editing? That's the best way I've found to ensure gradual improvement over time.

Also should run findbugs and see if there are any bugs you can fix in the code you're changing / adding. Don't have to fix ones outside this code, but shouldn't add new ones, and should burn down old ones.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/statistics/NonconformityTypeStatisticsDAOImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.persistence.Query;
+
+import org.springframework.stereotype.Repository;
+
+@Repository("nonconformityTypeStatisticsDAO")
+public class NonconformityTypeStatisticsDAOImpl extends BaseDAOImpl implements NonconformityTypeStatisticsDAO {
+
+	@Override
+	public List<NonconformityTypeStatisticsDTO> getAllNonconformityStatistics() {
+		String hql = "FROM NonconformityTypeStatisticsEntity WHERE deleted = false";
```
andlar: More indentation things? Do you have Eclipse set up to use spaces instead of tabs in all cases?

brianlindsy: For some reason these errors are not showing up in my eclipse but im going to look at my eclipse style settings

chpl/chpl-service/src/test/java/gov/healthit/chpl/dao/impl/StatisticsDAOTest.java
```diff
@@ -83,23 +85,23 @@ public static void setUpClass() throws Exception {
     }
 
     /**
-     * Given that getTotalDevelopers(DateRange) is called 
-     * When the start date is set to the beginning of time and endDate is set to now()
-     * Then all non-deleted developers are returned
+     * Given that getTotalDevelopers(DateRange) is called When the start date is
```
andlar: The Given-When-Then syntax is a feature of our acceptance criteria. I'd rather have it look right here and be flagged for checksyle than vice versa

PR #450, opened 2018-07-27
Ocd 2385 - update text describing /download endpoint

brianlindsy:


chpl/PENDING_OOC_RELEASE_NOTES.md
```diff
@@ -0,0 +1,9 @@
+
+## Version 15.17.0
+_Date TBD_
+
+### Minor modifications
+* Update text describing /download endpoint
+
```
kekey1: One thing here is that I think you'll need to merge the code from Todd's API PR first and add to that because otherwise when it eventually gets merged together in Bitbucket there will be conflicts in this release notes file since multiple people will have created it and written their own pieces of it.

PR #452, opened 2018-08-09
Inherited Certification Status report

andlar:


chpl/chpl-api/src/main/resources/log4j2.xml
```diff
@@ -26,7 +26,7 @@
             additivity="false">
             <AppenderRef ref="STDOUT" />
         </Logger>
-        <Logger name="net.sf.ehcache" level="INFO" additivity="fase">
+        <Logger name="net.sf.ehcache" level="INFO" additivity="falase">
```
kekey1: this is still misspelled :)

PR #456, opened 2018-08-14
Ocd-2352 - nonconformity charts

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/statistics/NonconformityTypeStatisticsDAOImpl.java
```diff
@@ -59,31 +59,10 @@ public void create(NonconformityTypeStatisticsDTO dto) {
...
-        List<NonconformityTypeStatisticsEntity> result = query.getResultList();
-
-        if (result.size() > 1) {
-            throw new EntityRetrievalException("Data error. Duplicate id in database.");
-        } else if (result.size() == 1) {
-            entity = result.get(0);
-        }
+    public void deleteAllOldNonConformityStatistics() throws EntityRetrievalException {
+        String hql = "UPDATE NonconformityTypeStatisticsEntity SET deleted = true WHERE deleted = false";
+        Query query = entityManager.createQuery(hql);
```
kekey1: I think when setting them to deleted individually you were also updating the last modified user. Is that necessary here as well?

andlar: The same process that's creating them is the one that's updating them, so maybe it's not necessary?

PR #457, opened 2018-08-15
Broken surveillance rules

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/app/TriggerJob.java
```diff
@@ -71,7 +71,8 @@ public static void main(final String[] args) {
                 Scheduler scheduler = sf.getScheduler();
                 for (String group: scheduler.getJobGroupNames()) {
                     for (JobKey jobKey : scheduler.getJobKeys(groupEquals(group))) {
-                        System.out.println("Found job: \"" + jobKey.getName() + "\" in group: \"" + jobKey.getGroup() + "\"");
+                        System.out.println("Found job: \"" + jobKey.getName()
```
kekey1: Since you changed something with this line, can you make it a logged statement instead of system.out?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/schedule/ChplTrigger.java
```diff
@@ -17,6 +17,7 @@
     private ChplJob job;
     private String cronSchedule;
     private String email;
+    private String acb;
```
kekey1: I believe we allow ACBs to change their names and since this value is a string I'm assuming it's the name rather than the ID. Should we be using the ID here instead to cover ourselves when one ACB changes its name?

chpl/chpl-service/src/main/java/gov/healthit/chpl/app/TriggerJob.java
andlar: This app is supposed to be called from the command line, and this particular thing is sort of the "help" for using the app. Will LOGGER output to the command line in all cases?

kekey1: Oh.. well it can be configured such that this class outputs to the console instead of a file so that would do it. Understand if you don't want to mess with the configs and leave as is.

PR #460, opened 2018-08-17
Fix some unit tests; remove unnecessary jobs.xml files

andlar:
[#development]


kekey1:
So something about only merging some of the PRs into DEV... mine (with the validation refactor) has also fixed the API unit tests so we have some of the same fixes (remember I was asking you about my Jenkins build in which all test were passing?). How to handle this? If we had been able to merge all PRs you wouldn't have duplicated the work.

PR #462, opened 2018-08-21
Ocd 2392- api throttle

brianlindsy:


chpl/chpl-api/src/main/java/gov/healthit/chpl/CHPLConfig.java
```diff
@@ -123,4 +140,6 @@ public InternalResourceViewResolver viewResolver() {
         viewResolver.setSuffix(".jsp");
         return viewResolver;
     }
+
+
```
andlar: Don't need blank lines

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+    
+    private int limit;
+ 
+    private Map<String, SimpleRateLimiter> limiters = new ConcurrentHashMap<>();
+    
+    private List<String> whitelist = new ArrayList<String>();
+    
+    public RateLimitingInterceptor(){}
+    
+    public String getMessage(final String messageCode, final String input, final String input2) {
```
andlar: This is getting ubiquitous enough I think it should be in some util class

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+import org.springframework.context.support.DefaultMessageSourceResolvable;
+import org.springframework.core.env.Environment;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+
+@Component
+public class RateLimitingInterceptor extends HandlerInterceptorAdapter implements EnvironmentAware{
+ 
+    private static final Logger logger = LogManager.getLogger(RateLimitingInterceptor.class);
```
andlar: Think this should be allcaps

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+            response.sendError(HttpStatus.TOO_MANY_REQUESTS.value(), getMessage("apikey.limit", String.valueOf(limit), timeUnit));
+            logger.info("Client with API KEY: " + key +  " went over API KEY limit of " + limit + ".");
+        }
+        response.addHeader("X-RateLimit-Limit", String.valueOf(limit));
+        return allowRequest;
+    }
+     
+    private SimpleRateLimiter getRateLimiter(String clientId) {
+        
+        if(limiters.containsKey(clientId)){
```
andlar: Need to run checkstyle; one rule I know we have is that `if(` should have a space: `if (`

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+            return TimeUnit.MINUTES;
+        }else if(unit.equals("hour")){
+            return TimeUnit.HOURS;
+        }
+        return null;
+    }
+
+    @PreDestroy
+    public void destroy() {
+        // loop and finalize all limiters
```
andlar: Does this do anything?

chpl/chpl-resources/src/main/resources/environment.properties.template
```diff
@@ -18,6 +18,14 @@ surveillanceThresholdToProcessAsJob=50
 #file name that daily SED details will be written to
 SEDDownloadName=chpl-sed-all-details.csv
 
+# number of tokens per API Key per time unit below
+rateTokenLimit=5
+# time unit where tokens above will be used
+# values are: second, minute, hour
+rateLimitTimeUnit=minute
+## api keys to be whitelisted seperated by commas
+whitelist=12909a978483dfb8ecd0596c98ae9094
```
andlar: Should not have "real" API key committed to github. Should be something like:
`whitelist=api_key_here,second_api_key`

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
kekey1: I already made a util class for this and it's my PR related to validation changes. If we're going to merge the PRs in order, Brian could bring in my code from that and use the util already written? Another issue with leaving PRs sitting in Dev!

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+     
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler)
+            throws Exception {
+        String key = null;
+        String clientIdParam = request.getParameter("api_key");
+        String clientIdHeader = request.getHeader("API-Key");
+        
+        if (clientIdHeader == clientIdParam) {
```
kekey1: Can't use == on strings, use .equals()

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+                key = clientIdParam;
+            } else if (clientIdParam == null) {
+                key = clientIdHeader;
+            } else {
+                return false;
+            }
+        }
+        
+        // let non-API requests pass
+        if (whitelist.contains(key)) {
```
kekey1: Does this mean that we have to put the UIs API key in the environment.properties file? I don't like having to have it in both the DB and properties. Is there a way for you to just query it from the DB? Or perhaps add a column to the DB to flag any particular API key as whitelisted?

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
andlar: That sounds like a plan

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -2,6 +2,29 @@
 ## Version 16.2.1
 _Date TBD_
 
+### New Features
+* Add nonconformity charts statistics endpoint
```
andlar: These top lines are already on PROD. The next few are on STG. I think only line 16 should be in this section. Also, if there are actually any new features then the Version should be different

PENDING_RELEASE_NOTES.md
```diff
@@ -1,14 +0,0 @@
-
-## Version 16.2.0
-_Date TBD_
-
-### New Features
```
andlar: These need to come back. The are the PENDING RELEASE NOTES, as they're on STG right now

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public void setEnvironment(Environment env) {
+        LOGGER.info("setEnvironment");
+        this.env = env;
+        this.timeUnit = env.getProperty("rateLimitTimeUnit");
+        this.limit = Integer.valueOf(env.getProperty("rateTokenLimit"));
+    }
+
+    /** {@inheritDoc} */
+    public String getMessage(final String messageCode, final String input, final String input2) {
```
andlar: You have Katy's newest code, right? And in that I think she had a utility function for this functionality. Should use that

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+
+        if (!StringUtils.isEmpty(clientIdParam)) {
+            key = clientIdParam;
+        } else if (!StringUtils.isEmpty(clientIdHeader)) {
+            key = clientIdHeader;
+        } else {
+            return false;
+        }
+
+        List<ApiKeyDTO> keyDtos = apiKeyDao.findAllWhitelisted();
```
andlar: Does this make a call to the database every time the interceptor does its checking? If so, isn't that going to be _a lot_? Is there a way to preload the whitelisted keys once on startup and hold on to them?

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
kekey1: The class is ErrorMessageUtil, you can see examples of it being used in the gov.healthit.chpl.validation.listing.reviewer code

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
kekey1: Maybe potentially use the caching framework that we have in place?

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -2,6 +2,9 @@
 ## Version 16.2.1
```
andlar: If the PENDING_RELEASE_NOTES version is at 16.2.0, and this PR now has a new Feature, then the new version should be 16.3.0, not 16.2.1, which would indicate it was only a bugfix

chpl/chpl-api/src/main/java/gov/healthit/chpl/registration/RateLimitingInterceptor.java
kekey1: I think you have to add something to the eh-cache.xml file as well. We are using constants for the cache names (look in the gov.healthit.chpl.caching package). I think all the other cacheable things are in managers as opposed to DAOs but it is probably ok to put that annotation at the DAO level.

PR #463, opened 2018-08-22
Correctly arrange data in csv file

andlar:
[#OCD-2272]


kekey1:
Errr wait do you mean to merge this into staging?

andlar:
Yep. This story is on STG, so needs to be pushed there. Will also get it to DEV, but don't want to get DEV stuff onto STG

PR #467, opened 2018-08-27
OCD-2399 - Migrate Quartz jobs into Tomcat environment

tmy1313:
This change also requires a change to the Bamboo deployment - changes to the web.xml file are required.


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/SchedulerCertifiedProductSearchDetailsAsync.java
```diff
@@ -19,7 +19,8 @@
  */
 @Service("schedulerCertifiedProductSearchDetailsAsync")
 public class SchedulerCertifiedProductSearchDetailsAsync {
-    private static final Logger LOGGER = LogManager.getLogger(SchedulerCertifiedProductSearchDetailsAsync.class);
+    //private static final Logger LOGGER = LogManager.getLogger(SchedulerCertifiedProductSearchDetailsAsync.class);
+    private Logger logger;
```
kekey1: Remove the commented out line? Just curious, is there a reason for moving all the static loggers into the getLogger method? Most of the rest of our code uses the static logger initialization.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/BrokenSurveillanceRulesCreatorJob.java
```diff
@@ -50,47 +49,46 @@
  */
 @DisallowConcurrentExecution
 public class BrokenSurveillanceRulesCreatorJob extends QuartzJob {
-    private static final Logger LOGGER = LogManager.getLogger(BrokenSurveillanceRulesCreatorJob.class);
+    private static final Logger LOGGER = LogManager.getLogger("brokenSurveillanceRulesCreatorJobLogger");
```
kekey1: This statically declared logger didn't have to get moved into getLogger... why?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/SchedulerCertifiedProductSearchDetailsAsync.java
tmy1313: This was done to allow for each job to log to it's own file.  For classes that are shared by multiple jobs, the "job"  class will pass the correct logger to the shared class.  If no logger is set, the getter will initialize the basic logger for the class.

tmy1313: So this particular class is not an actual "job".  I consider it a "helper/shared" class.  Job classes should have a the normal static logger defined.  Helper/shared classes should have the logger passed in so that they use the same logger as the job that uses the helper/subordinate class.

andlar:
The "TriggerJob" application and script still need that jar, and I don't think we have a way to trigger system-level Quartz jobs in any other way, unless the monitoring software actually supports that. I haven't tried :)

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/BrokenSurveillanceRulesCreatorJob.java
tmy1313: Other code called by the job, should be logged to the job's log file.  That is the need for the getter/setter for the logger in the shared files.  I did not go down to the Manager and DAO level, but that can be looked into.  With the changes I made, I captured ~95% of all logger statements into the specific job's logger file.

kekey1: Ok - I guess just keep it in mind if we see logging in the main Tomcat log that isn't the result of user activity on the CHPL. I feel like it could get all mixed up in there and be confusing when trying to debug something but seems like it would be a lot of work to change. It is probably not even an issue until we start logging more stuff. :)

PR #473, opened 2018-08-29
Return correct "full name" in JWT

andlar:
[#OCD-794]

### DIFF
--- a/chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/dto/UserDTO.java
+++ b/chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/dto/UserDTO.java
@@ -27,11 +27,17 @@ public class UserDTO implements UserDetails {
     private boolean credentialsExpired;
     private boolean accountEnabled;
 
-
+    /**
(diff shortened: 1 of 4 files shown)


chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/dto/UserDTO.java
```diff
@@ -111,9 +117,9 @@ public void setTitle(String title) {
         return null;
     }
 
-    public String getName() {
-        return subjectName;
-    }
+//    public String getName() {
+//        return subjectName;
+//    }
```
tmy1313: No big deal, but do we want to remove this code?

andlar: I thought about it, but did one of those "find references" things and it showed that some of the jars were looking for it. I have tried testing with it commented out and it seems to be okay, but on the off chance something weird breaks I'd like to have this still here as a reminder. In a week or two, or the next time anyone's in that code, then I think it could be removed.

PR #474, opened 2018-08-29
OCD-2216 - Within the Summary Statistics email, I want to see the number of open surveillance activities and open NCs by ONC-ACBs

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/statistics/SurveillanceStatisticsDAOImpl.java
```diff
@@ -55,10 +56,54 @@ public Long getTotalOpenSurveillanceActivities(DateRange dateRange) {
...
+                    + "SurveillanceEntity s "
+                + "WHERE s.startDate <= now() "
+                + "AND (s.endDate IS NULL OR s.endDate >= now()) "
+                + "AND cp.certificationBodyId = cb.id "
+                + "AND cp.id = s.certifiedProductId ";
+
+        if (dateRange == null) {
+            hql += "AND s.deleted = false ";
+        } else {
+            hql += "AND ((s.deleted = false AND s.creationDate <= :endDate) " + " OR "
```
andlar: I don't like this weird construction around the deleted flag. The "deleted = false" should be in the hql a few lines above, and this if should only be dealing with the date range

tmy1313: That is existing code that was lifted from the `getTotalOpenSurveillanceActivities` method, which was used as a basis for this method since it gets the totals for all ACBs.  This construct is used throughout the class.  It is also all through `ListingStatisticsDAOImpl`.

If you really want me to change it I can, but I should change it in all methods.

Let me know...

andlar: I know it was existing, but that doesn't mean it's right. I'd prefer if we actually make our code better every chance we can, and this is one of those chances. If you're out of time on the task, or rushing to get it out the door, then feel free to skip it. I wouldn't bother with a different file either, but it'd be a matter of minutes to change the half dozen places where this code has this stuff.

Besides, it's not really existing code if you're creating the methods in this PR.

PR #475, opened 2018-08-30
OCD-2364: fix bugs found so far

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/InpatientModular2014LegacyListingValidator.java
```diff
@@ -10,7 +10,7 @@
 import gov.healthit.chpl.validation.listing.reviewer.edition2014.InpatientG1G2RequiredData2014Reviewer;
 
 @Component("inpatientModular2014LegacyListingValidator")
-public class InpatientModular2014LegacyListingValidator extends Edition2014ListingValidator {
+public class InpatientModular2014LegacyListingValidator extends Edition2014LegacyListingValidator {
```
andlar: Is this the only Legacy 2014 Validator that needs this change? Or do all of the other legacy ones extend from this one?

kekey1: Yeah I checked the other ones and they were right. I guess I just pasted or typed the inheritance wrong on this one.

PR #491, opened 2018-09-20
Ocd 1310 - password strength checker

andlar:


chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/controller/AuthenticationController.java
```diff
@@ -20,110 +24,160 @@
...
+            notes = "This service generates a new password, saves it to the user's account "
+                    + " and sends an email to the address associated with the user's account "
+                    + " containing the new password.")
+    @RequestMapping(value = "/reset_password", method = RequestMethod.POST,
+    consumes = MediaType.APPLICATION_JSON_VALUE,
+    produces = "application/json; charset=utf-8")
+    public String resetPassword(@RequestBody final UserResetPasswordJSONObject userInfo)
+            throws UserRetrievalException, MessagingException {
+
+        String newPassword = userManager.resetUserPassword(userInfo.getUserName(), userInfo.getEmail());
```
kekey1: Is this method resetting it to a password that would pass our strength test?

andlar: I don't know. I didn't want to change current functionality. It probably is, as most of the 15 character random strings I generated to test things pass, but this could generate something with patterns or real words. I didn't think it was important to change it, because I'd expect a user to change their password when they get this email. I know I would; otherwise your actual password was transmitted in the clear.

That said, we probably should send them a link to create a new password instead of sending it in the clear; I think that's a no-no

kekey1: Ok so new ticket for that?

andlar: I have it in my notes to bring up

PR #497, opened 2018-09-26
Ocd 1310 -  Check password even if friendly name is null

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/app/TriggerJob.java
```diff
@@ -28,64 +29,165 @@ private TriggerJob() { }
...
+            default:
+                displayHelp();
+                break;
+            }
+        } else {
+            displayHelp();
+        }
+    }
+
+//    private static void interruptJob(final String[] args) {
```
kekey1: Did you mean to check this in with all the commented out code?

PR #500, opened 2018-09-27
Fix Unit Test failure due to bad URL

andlar:
[#OCD-1310]


kekey1:
It's hard to see the relevant changes among the checkstyle changes but if it fixes a unit test, I approve!

andlar:
I know. It's tricky because I can't see the code without fixing the checkstyle stuff, as it highlights SO MANY LINES in Eclipse for me. This one was failing because the 2015 upload test resource had "www.epic.com" as an URL, but that doesn't pass validation any more...

PR #504, opened 2018-10-03
Remove invalid test functionality; change error to warning

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/TestFunctionalityReviewer.java
```diff
@@ -42,34 +43,40 @@ public void review(final PendingCertifiedProductDTO listing) {
...
-                    for (PendingCertificationResultTestFunctionalityDTO crtf : cr.getTestFunctionality()) {
-                        listing.getErrorMessages().addAll(
-                                getTestingFunctionalityErrorMessages(crtf, cr, listing));
+                    Iterator<PendingCertificationResultTestFunctionalityDTO> crtfIter =
+                            cr.getTestFunctionality().iterator();
+                    while (crtfIter.hasNext()) {
+                        PendingCertificationResultTestFunctionalityDTO crtf = crtfIter.next();
+                        TestFunctionalityDTO tf = getTestFunctionality(crtf.getNumber());
+                        if (tf == null) {
+                            listing.getWarningMessages().add(
```
andlar: I think this should be an error. My understanding is that warnings = "we think there's something wrong, so we did something or are telling you about it" and errors = "something in your upload file is _wrong_; don't do that again". It's true that all of our other errors, if unaddressed, would cause the confirmation to fail, and this one we fix, but we're actually deleting data in their upload, and I think we want them to be very aware of that. My opinion can change, at the direction of the product owner, but...

kekey1: I am happy to change it back to an error! I made it a warning because I thought that's what we said in Teams. When I looked at fuzzy matching, I saw that is a warning and I thought we intended for this to work similarly. I'll check with Heather.

PR #507, opened 2018-10-04
Ocd 2429 - chart data job

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/CertificationEditionDAOImpl.java
```diff
@@ -18,6 +19,7 @@
 public class CertificationEditionDAOImpl extends BaseDAOImpl implements CertificationEditionDAO {
 
     @Override
+    @Transactional
```
andlar: Why are all of these transactional? Is this as a result of the need to do the "delete and save" inside a single transaction? If so, these methods probably don't need this annotation

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/chartdata/NonconformityTypeChartCalculator.java
```diff
@@ -0,0 +1,67 @@
...
+        return dtos;
+    }
+
+    @Transactional
+    @Async
+    private void deleteExistingNonconformityStatistics() throws EntityRetrievalException {
+        nonconformityTypeStatisticsDAO.deleteAllOldNonConformityStatistics();
+    }
+
+    @Transactional
```
andlar: This is the one that's supposed to be a single transaction. What's the @Async annotation for?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/chartdata/ParticipantAgeStatisticsCalculator.java
```diff
@@ -99,22 +94,16 @@ private void initialize() {
     }
 
     private void save(final List<ParticipantAgeStatisticsEntity> entities) {
```
andlar: Shouldn't this method be @Transactional too? This is the method that does the delete & save calls, right? So this should be done as a single transaction

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/chartdata/SedDataCollector.java
```diff
@@ -79,6 +88,7 @@
         Date startTime = new Date();
         for (Future<CertifiedProductSearchDetails> future : futures) {
             try {
+                CertifiedProductSearchDetails currDetails = future.get();
```
andlar: currDetails isn't used anywhere, right? It shouldn't be needed.

chpl/chpl-service/src/main/resources/jobs.xml
```diff
@@ -432,7 +432,25 @@
                 </entry>
             </job-data-map>
         </job>
-
+        <!-- Chart data creator -->
```
andlar: Also needs to be in the other jobs.xml files, otherwise it won't deploy on DEV/STG/PROD

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/chartdata/ChartDataCreatorJob.java
```diff
@@ -0,0 +1,145 @@
...
+            props.load(in);
+            in.close();
+        }
+        return props;
+    }
+
+    @Override
+    @Transactional
+    public void execute(JobExecutionContext arg0) throws JobExecutionException {
+        LOGGER.info("*****Chart Data Generator is startin now.*****");
```
andlar: startin -> starting

chpl/chpl-api/src/main/resources/log4j2.xml
```diff
@@ -28,6 +28,15 @@
                 <TimeBasedTriggeringPolicy interval="1" modulate="true"/>
             </Policies>
         </RollingFile>
+        <RollingFile name="chartDataCreatorJob" fileName="${logDir}/scheduler/chartDataCreatorJob.log"
```
kekey1: Also make sure to add this to the log4j configs in dev/stg/prod environments

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/chartdata/ParticipantAgeStatisticsCalculator.java
kekey1: The job's execute method is annotated as transactional and I think Spring's default behavior is to join an existing transaction if there is one otherwise start a new one... so all these transactional methods (I think) will be part of the top-level transaction. If that's not desired then some re-organization of the annotations should be done. Generally we try not to make DAO methods transactional but leave it up to the user of the DAO (a manager or calculator or something) to determine what should happen atomically.

brianlindsy: The reason I added those Transactional tags to the DAO methods is becuase I was getting errors like the following "javax.persistence.TransactionRequiredException: no transaction is in progress" when the entityManager.flush(); methods were getting called within those methods, in the ListingCountStatisticsDAOImpl for example. Now, Im not entirely sure why those transactional annotations worked but I just tried to take those annotations away from the DAO methods and added it to the top level save method inside ListingCountsStatisticsCalculator and got the same error as above.

kekey1: Ok, I don't think I understand how all of the Spring stuff works in the jobs vs how it works in our webapp classes because I thought none of that autowired stuff should be working either when classes are created with new Something() but it works. Maybe @tmy1313  has some insight? Or just leave the annotations and forget about it since it's working?

chpl/chpl-service/src/main/resources/jobs.xml
```diff
@@ -432,7 +432,25 @@
...
+            <recover>false</recover>
+        </job>
+        <trigger>
+            <cron>
+                <name>generateChartData</name>
+                <group>chartDataCreatorTrigger</group>
+                <job-name>chartDataCreator</job-name>
+                <job-group>systemJobs</job-group>
+                <misfire-instruction>MISFIRE_INSTRUCTION_DO_NOTHING</misfire-instruction>
+                <cron-expression>0 28 11 * * ?</cron-expression> <!-- At 0230 UTC every day -->
```
andlar: Comment does not match cron expression

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/chartdata/ParticipantAgeStatisticsCalculator.java
kekey1: I'm fine with moving forward with this as well!

andlar:
All of my code specific comments have been addressed, but not the PR wide one from my initial review:
> Needs RELEASE_NOTES, should remove the "generateChartData.sh" script, and should add something to the wiki about turning off cron on the various environments.

brianlindsy:
Removed generateChartData cron job from DEV and STG

andlar:
The latest commit message says "...release note..." but I don't see any changes in the commit itself

PR #508, opened 2018-10-04
Update script to use folder as parameter

andlar:
[#OCD-2446]


andlar:
Yes, but it won't be run via cron after I get all of these things changed. The Bamboo plan that will run this is already set up with the right arguments. Also, bonus, no need to copy that environment.properties file any more (I think)

PR #523, opened 2018-10-16
Ocd-2488 - add contact and developer info to 2014/2015 download files

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/presenter/CertifiedProductCsvPresenter.java
```diff
@@ -107,6 +117,31 @@ public Logger getLogger() {
         result.add(data.getCertifyingBody().get("name").toString());
         result.add(data.getOtherAcb());
         result.add(data.getDeveloper().getName());
+        if (data.getDeveloper().getAddress().getLine2() != null) {
```
andlar: Should this be "isEmpty"?

Also, for the Developer components, if they're empty (maybe no contact information, maybe something else) does the add that you have here correctly put a blank value in the csv? Or, if one of them is null, does it just not add the column at all?

brianlindsy: According to the definition files, those fields are required, which I think means that they will always have info?

andlar: They might be required, but if the Developer doesn't have any contact information (like for Developers who only have 2011 listings, or only 2014 listings that haven't been updated, or have new ones) then there won't be data to show

brianlindsy:
Ignore that latest push...

andlar:
Might need to merge development back into this branch to get down to just the changes that are relevant. I'm seeing a lot of MUU stuff...

PR #531, opened 2018-10-25
Fix issue with Questionable Activity threshold

tmy1313:
OCD-2475

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
@@ -13,16 +13,14 @@ import org.springframework.stereotype.Component;
 
 import gov.healthit.chpl.auth.Util;
 import gov.healthit.chpl.dao.CertifiedProductDAO;
+import gov.healthit.chpl.domain.CertificationResult;
+import gov.healthit.chpl.domain.CertifiedProductSearchDetails;
 import gov.healthit.chpl.domain.concept.ActivityConcept;
 import gov.healthit.chpl.dto.DeveloperDTO;
 import gov.healthit.chpl.dto.ProductDTO;
 import gov.healthit.chpl.dto.ProductVersionDTO;
 import gov.healthit.chpl.manager.QuestionableActivityManager;
 
-/**
- * @author TYoung
- * Checks for and adds Questionable Activity when appropriate.
- */
 @Component
 @Aspect
 public class QuestionableActivityListener implements EnvironmentAware {
@@ -33,16 +31,9 @@ public class QuestionableActivityListener implements EnvironmentAware {
 
     private long listingActivityThresholdMillis = -1;
     private static final long MILLIS_PER_DAY = 24 * 60 * 60 * 1000;
-
-    /**
-     * Autowired constructor for dependency injection.
-     * @param env - Environment
-     * @param listingDao - CertifiedProductDAO
-     * @param questionableActivityManager - QuestionableActivityManager
-     */
+    
     @Autowired
-    public QuestionableActivityListener(final Environment env, final CertifiedProductDAO listingDao,
-            final QuestionableActivityManager questionableActivityManager) {
+    public QuestionableActivityListener(Environment env, final CertifiedProductDAO listingDao, final QuestionableActivityManager questionableActivityManager) {
         this.env = env;
         this.listingDao = listingDao;
         this.questionableActivityManager = questionableActivityManager;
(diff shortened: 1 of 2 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
```diff
@@ -13,16 +13,14 @@
...
+import gov.healthit.chpl.domain.CertifiedProductSearchDetails;
 import gov.healthit.chpl.domain.concept.ActivityConcept;
 import gov.healthit.chpl.dto.DeveloperDTO;
 import gov.healthit.chpl.dto.ProductDTO;
 import gov.healthit.chpl.dto.ProductVersionDTO;
 import gov.healthit.chpl.manager.QuestionableActivityManager;
 
-/**
- * @author TYoung
- * Checks for and adds Questionable Activity when appropriate.
```
andlar: Why remove the javadoc? Here and in the rest of the file?

chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
```diff
@@ -33,16 +31,9 @@
...
-     * Autowired constructor for dependency injection.
-     * @param env - Environment
-     * @param listingDao - CertifiedProductDAO
-     * @param questionableActivityManager - QuestionableActivityManager
-     */
+    
     @Autowired
-    public QuestionableActivityListener(final Environment env, final CertifiedProductDAO listingDao,
-            final QuestionableActivityManager questionableActivityManager) {
+    public QuestionableActivityListener(Environment env, final CertifiedProductDAO listingDao, final QuestionableActivityManager questionableActivityManager) {
```
andlar:
```suggestion
    public QuestionableActivityListener(final Environment env, final CertifiedProductDAO listingDao, final QuestionableActivityManager questionableActivityManager) {
```

chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
```diff
@@ -58,16 +49,70 @@ public void setEnvironment(final Environment e) {
...
+     * @param originalData
+     * @param newData
+     * @param reason
      */
+    @After("execution(* gov.healthit.chpl.manager.impl.ActivityManagerImpl.addActivity(..)) && "
+            + "args(concept,objectId,activityDescription,originalData,newData,reason,..)")
+    public void checkQuestionableActivityWithReason(final ActivityConcept concept,
+            final Long objectId, String activityDescription, final Object originalData,
+            final Object newData, final String reason) {
+        LOGGER.info("Called QuestionableActivityAspect2.checkQuestionableActivityWithReason()");
```
andlar: What's `QuestionableActivityAspect2`?

chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
```diff
@@ -58,16 +49,70 @@ public void setEnvironment(final Environment e) {
...
+        if (originalData == null || newData == null
+                || !originalData.getClass().equals(newData.getClass())
+                || Util.getCurrentUser() == null) {
```
andlar: Is there a way for there to be no current user? If so, does that mean it won't count as questionable activity? That seems off

chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
```diff
@@ -58,16 +49,70 @@ public void setEnvironment(final Environment e) {
...
+        Date activityDate = new Date();
+        Long activityUser = Util.getCurrentUser().getId();
+
+        if (originalData instanceof CertifiedProductSearchDetails
+                && newData instanceof CertifiedProductSearchDetails) {
+            CertifiedProductSearchDetails origListing = (CertifiedProductSearchDetails) originalData;
+            CertifiedProductSearchDetails newListing = (CertifiedProductSearchDetails) newData;
+
+            //look for any of the listing questionable activity
+            questionableActivityManager.checkListingQuestionableActivity(origListing, newListing, activityDate, activityUser, reason);
```
andlar: Does this check also check against the "grace period window"? Or does the grace period check need to be before this call?

chpl/chpl-service/src/main/java/gov/healthit/chpl/aspect/QuestionableActivityListener.java
tmy1313: This is existing logic.  I removed this method thinking it was the cause of some failing tests.  Before I moved the methods to this class it existed in the QuestionableActivityAspect class.  After further research found that the problem existed in the QuestionableActivityManager.  I'm not 100 sure about this code, but since it had been working, I simply put it back in.

tmy1313:
The release notes do not need to be updated.  This is all functionality that was working before I made these changes.  When I broke out the QuestionableActivityAspect into two separate classes (one for API, one for service), I broke it.

PR #532, opened 2018-10-26
Remove unneccessary request parameter

andlar:
[#OCD-2277]


andlar:
While I think it technically would break backwards-compatibility, it's only in one endpoint, that's behind security, and is fairly new. I am 99% confident no one with an account has written code to update products using a PUT method yet, so I didn't see a need to mark this as a major version bump for something no one's using. Similar reason for nothing in the RELEASE NOTES; this is an invisible change for probably 100% of our users

PR #543, opened 2018-11-05
Ocd 2544: update contact email throughout api

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/SwaggerConfig.java
```diff
@@ -37,31 +38,31 @@
...
     private ApiInfo apiInfo() {
-        LOGGER.info("get ApiInfo");
+        Contact contact = new Contact("CHPL Development Team. "
+                + "Please submit any questions using the Health IT Feedback Form and "
+                + "select the \"Certified Health IT Products List (CHPL)\" category.",
+                "https://www.healthit.gov/form/healthit-feedback-form",
+                "");
         return new ApiInfo("CHPL", "Certified Health IT Product Listing", "18.0.0", "http://terms/of/service.url",
-                "CHPL@ainq.com", "License Text", "https://github.com/chpladmin/chpl-api/blob/staging/LICENSE");
+                contact, "License Text", "https://github.com/chpladmin/chpl-api/blob/staging/LICENSE");
```
andlar: Should we/can we remove the terms of service link while we're in here? Or link to something that actually exists?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -337,7 +338,9 @@ public UserInvitation inviteUser(@RequestBody final UserInvitation invitation)
                 + "which will allow you to manage certified product listings on the CHPL. "
                 + "Please click the link below to create or update your account: <br/>"
                 + env.getProperty("chplUrlBegin") + "/#/registration/create-user/" + createdInvite.getInviteToken()
-                + "</p>" + "<p>If you have any questions, please contact the ONC CHPL Team at onc_chpl@hhs.gov.</p>"
+                + "</p>"
+                + "<p>If you have any issues completing the registration, "
+                + "please contact the ONC CHPL Team at onc_chpl@hhs.gov.</p>"
```
andlar: These _are_ HTML emails; is it worth it to make these mailto: links?

chpl/chpl-api/src/main/java/gov/healthit/chpl/SwaggerConfig.java
kekey1: We can! Do we have terms of service online or should I just remove the link?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
kekey1: We can if you want. I've only ever sent these emails to my Ai account and looked at them in Outlook but in there they appear as mailto links anyway. I think it recognizes email addresses and displays them as such.

chpl/chpl-api/src/main/java/gov/healthit/chpl/SwaggerConfig.java
andlar: I don't think we have any TOS anywhere. Unless ONC has some I'd suggest we just remove it

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
andlar: Hmm. I hadn't noticed that. I forwarded one to my gmail account and it does that as a link too. I think it'd be a good idea to be explicit about it, but could go either way

PR #552, opened 2018-11-14
2015 functionality tested should be restricted by criteria

tmy1313:
OCD-2532


chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/TestFunctionalityDTO.java
```diff
@@ -77,17 +74,16 @@ public void setPracticeType(final PracticeTypeDTO practiceType) {
...
-        return certificationCriterion;
-    }
-
-    /**
-     * @param certificationCriterion the certificationCriterion to set
-     */
-    public void setCertificationCriterion(final CertificationCriterionDTO certificationCriterion) {
-        this.certificationCriterion = certificationCriterion;
+    @Override
+    public boolean equals(Object o) {
```
kekey1: I may be dating myself here, but last time I looked if you override equals then you also were supposed to override hashcode. Is this still a thing?

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/TestFunctionalityCriteriaMapEntity.java
```diff
@@ -0,0 +1,110 @@
...
+    @JoinColumn(name = "criteria_id", insertable = false, updatable = false)
+    private CertificationCriterionEntity criteria;
+    
+    @Basic(optional = false)
+    @OneToOne(optional = false, fetch = FetchType.LAZY)
+    @JoinColumn(name = "test_functionality_id", insertable = false, updatable = false)
+    private TestFunctionalityEntity testFunctionality;
+    
+    @Basic(optional = false)
+    @Column(name = "creation_date", nullable = false)
```
kekey1: Can you make creation date and last modified date also insertable = false and updatable = false? The older entities aren't written that way but most(?) newer ones are since these values are set by the db and triggers rather than through the API.

chpl/chpl-service/pom.xml
```diff
@@ -319,120 +319,122 @@
...
-                    <execution>
-                        <id>schemagen</id>
-                        <goals>
-                            <goal>schemagen</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
-        </plugins>
+        <pluginManagement>
```
andlar: What's `<pluginManagement>`?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/TestFunctionalityCriteriaMapDTO.java
```diff
@@ -0,0 +1,91 @@
+package gov.healthit.chpl.dto;
+
+import java.util.Date;
+
+import gov.healthit.chpl.entity.TestFunctionalityCriteriaMapEntity;
+
```
andlar: JavaDoc? I know checkstyle would have complained about this

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/TestingFunctionalityManagerImpl.java
```diff
@@ -1,84 +1,108 @@
...
+    private Map<String, List<TestFunctionalityDTO>> testFunctionalityByCriteria2014 = 
+            new HashMap<String, List<TestFunctionalityDTO>>();
+    
+    @Autowired
+    public TestingFunctionalityManagerImpl(TestFunctionalityDAO testFunctionalityDAO) {
+        this.testFunctionalityDAO = testFunctionalityDAO;
+    }
+    
+    @Override
+    public void onApplicationEvent(ContextRefreshedEvent event) {
```
kekey1: When does this get called?

PENDING_RELEASE_NOTES.md
```diff
@@ -4,6 +4,7 @@ _Date TBD_
 
 ### New Features
 * Update ONC contact information for CHPL API and on user invitation emails.
+* 2015 Functionality testing is restricted by criteria.
```
andlar: This says 2015, but there's code below that talks about 2014. Does it also restrict 2014 Listings?

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
```diff
@@ -12,8 +12,15 @@
...
 public class ErrorMessageUtil {
 
-    @Autowired private MessageSource messageSource;
-
+    private MessageSource messageSource;
+    
+    public ErrorMessageUtil() {}
+    
+    @Autowired 
+    public ErrorMessageUtil(MessageSource messageSource) {
```
andlar: Why did this change?

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
```diff
@@ -12,8 +12,15 @@
 @Component
 public class ErrorMessageUtil {
 
-    @Autowired private MessageSource messageSource;
-
+    private MessageSource messageSource;
```
kekey1: Is there an advantage to using the autowired constructor vs just having autowired on the variable? Most of the code has it the latter way, just wondering why change it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/TestFunctionalityReviewer.java
```diff
@@ -55,36 +71,54 @@ public void review(final PendingCertifiedProductDTO listing) {
...
 
         Set<String> errors = new HashSet<String>();
-        PracticeTypeDTO pt = practiceTypeDAO.getByName(cp.getPracticeType());
 
-        if (!isTestFunctionalityPracticeTypeValid(pt.getId(), tf)) {
-            errors.add(getTestFunctionalityPracticeTypeErrorMessage(crtf, cr, cp));
+        CertificationEditionDTO edition = getEditionDTO(getEditionFromListing(listing));
+        TestFunctionalityDTO tf = getTestFunctionality(crtf.getNumber(), edition.getId());
+        
+        //Only validate practice type for 20146
```
andlar: 20146?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/TestFunctionalityReviewer.java
```diff
@@ -22,12 +31,31 @@
  *
  */
 @Component("testFunctionalityReviewer")
-public class TestFunctionalityReviewer implements Reviewer {
+@Transactional
+public class TestFunctionalityReviewer implements Reviewer, ApplicationListener<ContextRefreshedEvent> {
     private static final String EDITION_2014 = "2014";
+    private static final String EDITION_2015 = "2015";
```
kekey1: I think there is a package for 2014-only reviewers and another package for 2015-only reviewers. Could you break up this class to have 2014 logic in the one and 2015 logic the other? Would also require a change to the validators to add the appropriate reviewer.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/TestFunctionalityReviewer.java
```diff
@@ -24,19 +29,30 @@
  *
  */
 @Component("pendingTestFunctionalityReviewer")
-public class TestFunctionalityReviewer implements Reviewer {
+public class TestFunctionalityReviewer implements Reviewer, ApplicationListener<ContextRefreshedEvent> {
+    private static final String EDITION_2014 = "2014";
+    private static final String EDITION_2015 = "2015";
```
kekey1: Same comment about breaking up the logic.

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
tmy1313: https://www.vojtechruzicka.com/field-dependency-injection-considered-harmful/

chpl/chpl-service/pom.xml
tmy1313: By wrapping the <plugins> tag with the <pluginsManagement> I was able to get rid pom.xml error that Eclipse has flagged in my project  since I started on the team.

PENDING_RELEASE_NOTES.md
tmy1313: 2014 listings were previously restricted by criteria.  It just so happened that for 2014 there was only one criteria for each test functionality.  The change for 2014 test functionalities is transparent to the users.  The data that had to be migrated for the 2014 to work with the new table structure has been noted in the release notes for the data model project.

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
tmy1313: Reverted...

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/TestFunctionalityDTO.java
tmy1313: Done

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/TestFunctionalityCriteriaMapEntity.java
tmy1313: Done

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/TestingFunctionalityManagerImpl.java
tmy1313: I wanted to run this code in a @PostConstruct, but not all of the beans that I needs to use in the logic were initialized.  This is an event that is raised by the spring context when all of the beans have been initialized.  It does not need to be explicitly called. 
But in some of the unit tests that are not dependent on the spring context, it does need to be called explicitly.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/TestFunctionalityCriteriaMapDTO.java
tmy1313: Added

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
kekey1: I'm ok with this then.. if we want to make this part of our coding style and change classes as we touch them going forward that is fine. Can you mention it at the Thursday meeting so that we are all on the same page? If it's already been mentioned I've obviously forgotten or missed it so please repeat. :)

andlar: If we're going to move towards constructor based initialization, then doesn't the constructor without the dependency need to be removed? Otherwise we lose out on the biggest advantage of constructor based vs. setter based initialization, in that a "new ErrorMessageUtil()" will have an NPE as soon as someone tries to "getMessage()"

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/TestingFunctionalityManagerImpl.java
kekey1: Okay... I guess I'm asking because we haven't done this anywhere else in the code. I have seen PostConstruct used and don't understand why it doesn't work here... which beans are not initialized? Also have seen in CertifiedProductDetailsManagerImpl in the Autowired constructor after each member of the class is set some data is loaded (loadCqms and loadCriteriaMacraMeasures are called). Could that work here instead? It doesn't seem right to have to call this from the unit tests or anywhere.

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
kekey1: Well you should never call new ErrorMessageUtil() if you want to use it a spring component. It should always be autowired into a class in some way otherwise you lose all the springyness of it. Not sure if the default constructor has to be removed but I could get on board with that if it works.

andlar: Everything I read said one of the reasons you do this with a constructor is so that, when someone instantiates the class not in a spring-y way, they'll be required to pass in the dependency. I think that means we'd need to remove the "non-dependency injected" constructors. As long as it compiles and works without that default one I think we should do that, if we're going down this path

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/TestingFunctionalityManagerImpl.java
tmy1313: For some reason, when I wrote the code the the DAO was not initialized when the @PostConstruct ran.  I'm going to give this a try:  https://nixmash.com/post/using-postconstruct-and-dependson-in-spring

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
```diff
@@ -1,4 +1,4 @@
-package gov.healthit.chpl.util;
+    package gov.healthit.chpl.util;
```
kekey1: No indent needed here

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
andlar: @tmy1313  did you undo the changes you made here? I think we got to the point where moving to constructor-injection sounded like the right idea

andlar: I'm good with this one. @kekey1 are you satisfied with these changes wrt this conversation?

kekey1: Yep!

PENDING_RELEASE_NOTES.md
andlar: This needs to be in PENDING_PENDING

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,8 @@
+
+## Version 19.2.0
+_Date TBD_
+
+### New Features
+* 2015 Functionality testing is restricted by criteria.
+
+---
```
andlar: Needs a blank line after the `---`

PR #554, opened 2018-11-15
Ocd 2524 - password reset

brianlindsy:


chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/dao/impl/UserResetTokenDAOImpl.java
```diff
@@ -0,0 +1,89 @@
...
+@Repository(value="userResetTokenDAO")
+public class UserResetTokenDAOImpl extends BaseDAOImpl implements UserResetTokenDAO {
+    private static final Logger logger = LogManager.getLogger(UserResetTokenDAOImpl.class);
+
+    public UserResetTokenDTO create(String resetToken, Long userId) {
+        UserResetTokenEntity entity = new UserResetTokenEntity();
+        entity.setUserResetToken(resetToken);
+        entity.setUserId(userId);
+        entity.setCreationDate(new Date());
+        entity.setLastModifiedDate(new Date());
```
kekey1: Creation date and last modified date should not need to be set here since the database sets them. You may need to mark them as insertable = false, updatable = false in the entity.

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/entity/UserResetTokenEntity.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.persistence.Table;
+
+import gov.healthit.chpl.auth.user.User;
+
+
+@Entity
+@Table(name = "user_reset_token")
+public class UserResetTokenEntity {
+
+    private static final long serialVersionUID = 1L;
```
kekey1: Does any other class have the UID of 1? Seems like a potential conflict... it's usually a weird randomly generated long value.

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/UserManager.java
```diff
@@ -71,4 +70,10 @@
      * @return a Strength object with password strength information
      */
     Strength getPasswordStrength(UserDTO user, String password);
+
+    UserResetTokenDTO createResetUserPasswordToken(String username, String email) throws UserRetrievalException;
```
kekey1: Can you mark this method as public if it is?

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/impl/UserManagerImpl.java
```diff
@@ -275,16 +327,17 @@ public String getEncodedPassword(final UserDTO user) throws UserRetrievalExcepti
...
     @Override
     public Set<UserPermissionDTO> getGrantedPermissionsForUser(final UserDTO user) {
         return securedUserManager.getGrantedPermissionsForUser(user);
     }
 
-
     @Override
     public UserDTO getByName(final String userName) throws UserRetrievalException {
-        return securedUserManager.getBySubjectName(userName);
+        loginAdminUser();
```
kekey1: Where is this being called from in the new code that is requiring a user to be logged in when it's called? Could the place that's calling it just go to the DAO instead of the manager?

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/impl/UserManagerImpl.java
```diff
@@ -231,37 +257,62 @@ public void updateFailedLoginCount(final UserDTO userToUpdate) throws UserRetrie
...
 
-        return password;
+        return userResetToken;
+    }
+
+    // checks that the token was made in the last 24 hours
+    private boolean isTokenValid(UserResetTokenDTO userResetToken) {
+        Date checkDate = userResetToken.getCreationDate();
+        Instant now = Instant.now();
+        return (!checkDate.toInstant().isBefore(now.minus(24, ChronoUnit.HOURS)))
```
andlar: 24 should be a property, not a magic number

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/dao/impl/UserResetTokenDAOImpl.java
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            entity.setLastModifiedUser(userId);
+        }
+        entity.setDeleted(false);
+        
+        create(entity);
+        return new UserResetTokenDTO(entity);
+    }
+    
+    public void deletePreviousUserTokens(Long userId) {
```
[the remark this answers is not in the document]
brianlindsy: In the method UserManager.createResetUserPasswordToken this method is called as well.

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/impl/UserManagerImpl.java
```diff
@@ -231,37 +258,62 @@ public void updateFailedLoginCount(final UserDTO userToUpdate) throws UserRetrie
...
         if (foundUser == null) {
             throw new UserRetrievalException("Cannot find user with name " + username + " and email address " + email);
         }
 
-        //create new password
+        // create user password reset token
         char[] buf = new char[GENERATED_PASSWORD_LENGTH];
 
         for (int idx = 0; idx < buf.length; ++idx) {
             buf[idx] = SYMBOLS[random.nextInt(SYMBOLS.length)];
```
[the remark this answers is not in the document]
brianlindsy: I can use RandomStringGenerator instead, also, thoughts on how long this should be? Is 15 characters long enough?

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/impl/UserManagerImpl.java
brianlindsy: Its being called in the AutheticationController.resetPassword method. Because people are resetting passwords they probably aren't logged in (at least I don't think that's possible). So we need to login temporarily so that we can get past the authentication in the securedUserManager class. I never though about calling DAO's directly though....

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/impl/UserManagerImpl.java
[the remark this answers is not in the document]
andlar: What about a random UUID?

https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#randomUUID()

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/impl/UserManagerImpl.java
```diff
@@ -231,37 +259,62 @@ public void updateFailedLoginCount(final UserDTO userToUpdate) throws UserRetrie
...
 
-        return password;
+        return userResetToken;
+    }
+
+    // checks that the token was made in the last 24 hours
+    private boolean isTokenValid(UserResetTokenDTO userResetToken) {
+        Date checkDate = userResetToken.getCreationDate();
+        Instant now = Instant.now();
+        return (!checkDate.toInstant().isBefore(now.minus(HOURS_IN_DAY, ChronoUnit.HOURS)))
```
andlar: I didn't want the number moved to a static variable. I wanted it moved to a property in the properties file. We don't know what the right duration of the link should be. It could be 30 minutes, or 2 hours, or three days. I'd like the property to be something like "resetLinkExpirationTimeInHours" and use that value here

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/manager/impl/UserManagerImpl.java
kekey1: It just seems like this method is one that could be (is?) called from elsewhere in the application and if it was the behavior could be odd. Like, what happens if you're logged in as someone (not admin) and do some action that calls this code (I don't know if that is currently possible, but maybe)? Are you then logged out if that happens? I think this needs to be changed around in some way so that the login and logout lines are not a part of this method. Maybe that's calling the DAO from the controller - which we don't normally do but maybe it's ok in this case? - or maybe that's adding some other non-secured manager method.

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/entity/UserResetTokenEntity.java
brianlindsy: I dont know why Github still shows this code but I did update the entity to not have a serialUID because most others dont

PR #556, opened 2018-11-16
Add Reason to Questionable Activity when changing Developer status

tmy1313:
OCD-2560

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/questionableActivity/QuestionableActivityDeveloperEntity.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/questionableActivity/QuestionableActivityDeveloperEntity.java
@@ -44,6 +44,9 @@ public class QuestionableActivityDeveloperEntity implements QuestionableActivity
     @Column(name = "after_data")
     private String after;
     
+    @Column(name = "reason")
+    private String reason;
+    
     @Column(name = "activity_date")
     private Date activityDate;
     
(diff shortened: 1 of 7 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/questionableActivity/QuestionableActivityDeveloperEntity.java
```diff
@@ -177,5 +180,22 @@ public DeveloperEntity getDeveloper() {
...
+    public String getReason() {
+        return reason;
+    }
+
+    public void setReason(String reason) {
+        this.reason = reason;
+    }
+
+    @Override
+    public String toString() {
```
andlar: Love this. Can you do it for the DTO too? Unless it's already there, in which case, can you update it so we're moving to the same format for all `toString()`s

tmy1313: The associated DTO is a little trickier since it extends a base QuestionableActivityDTO class.  I'll give it a try, though.

andlar: Even if it doesn't have the elements from the base class, and just has the ones that are in the extension, it'd still be better than the default toString

tmy1313: Eclipse IDE allowed my access inherited methods.  I left the default notation of using the method name "getId()=xxx" to help indicate that the value is coming from the base object and is not from the current object.

PR #557, opened 2018-11-19
Fix jUnit tests

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/util/ErrorMessageUtil.java
```diff
@@ -19,6 +19,10 @@ public ErrorMessageUtil(final MessageSource messageSource) {
         this.messageSource = messageSource;
     }
 
+    public ErrorMessageUtil() {
```
andlar: I had thought part of the reason we were going to constructor based injection was to remove the default constructors to make those injections explicit. Is that not the case? Or is there something else going on here?

PR #564, opened 2018-11-26
Require password change on next login

andlar:


chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/controller/AuthenticationController.java
```diff
@@ -57,32 +61,55 @@
     @Autowired
     private UserManager userManager;
 
+    @Autowired
+
```
tmy1313: Extra line??

PR #566, opened 2018-11-26
Ocd 2565

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/StatusController.java
```diff
@@ -42,24 +45,27 @@
...
      */
     @ApiOperation(
-            value = "Check if the basic search cache has completed loading. "
-                    + "{ status: 'OK', age: long } is returned if it's finished and "
-                    + "{ status: 'INITIALIZING' } is returned if not. "
-                    + "Age indicates the number of miliseconds since the cache "
-                    + "was last refreshed.",
+            value = "Check the status of every cache. "
+                    + "{ status: 'OK' } is returned if all caches are loaded and "
+                    + "{ status: 'PENDING' } is returned if not. ",
```
andlar: Is this enough of a change to the API that we need a major version bump? If someone was using "INITIALIZING" for something they'd have to update, right? Would it be better to not change the word, even though the logic behind it is different?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/StatusController.java
```diff
@@ -42,24 +45,27 @@
...
-    public @ResponseBody String getCacheStatus() {
+    public @ResponseBody CacheStatus getCacheStatus() {
+        CacheStatus response = new CacheStatus();
         CacheManager manager = CacheManager.getInstance();
-        Cache basicCache = manager.getCache(CacheNames.COLLECTIONS_LISTINGS);
-        if (basicCache == null || basicCache.getKeysNoDuplicateCheck().size() == 0) {
-            return "{\"status\": \"" + CacheStatus.INITIALIZING + "\"}";
+        boolean anyPending = false;
+        List<String> cacheNames = CacheInitializor.getPreInitializedCaches();
+        for(int i = 0; i < cacheNames.size(); i++) {
```
andlar: I know it's minor, but checkstyle complains about `if(` instead of `if (` and similarly with for/try/catch/etc.

chpl/chpl-resources/src/main/resources/environment.properties.template
```diff
@@ -70,11 +70,6 @@ inheritanceReportEmailWeeklyFileName=inheritance-validation-weekly-report.csv
 inheritanceReportEmailWeeklyNoContent=<p>No listings were found that do not meet ICS requirements.</p>
 inheritanceReportEmailAcbWeeklyHtmlMessage=<h3>ONC-ACB Weekly ICS Validation Report</h3>
 
-#cache status maximum age email properties
```
andlar: Should make a note to remove these from deployment jobs during push, but not before. Probably in the wiki?

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/ListingsCollectionCacheUpdater.java
```diff
@@ -35,8 +35,6 @@ public synchronized void refreshCache() {
         CacheManager manager = CacheManager.getInstance();
         prefetchedCaches.loadPreFetchedBasicSearch();
         LOGGER.debug("Adding cache generation time to prefetched cache.");
```
andlar: This line isn't happening anymore, right?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheStatusName.java
```diff
@@ -0,0 +1,5 @@
+package gov.healthit.chpl.domain;
+
+public enum CacheStatusName {
+    OK, PENDING;
```
andlar: I saw something in a comment recently about enums and hibernate (I think). We needed to do something to make enums work correctly somewhere? Does that ring a bell?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/StatusController.java
kekey1: I agree - I'll change back to Initializing. I was just going by the language in the ticket.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CacheStatusName.java
kekey1: There is something you have to do to make an enum work with hibernate and we have done it with a few enums in the past but this class is in the domain package (used to send data back and forth to the front end) and won't ever touch the database.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/StatusController.java
andlar: You know as well as I do the language in the ticket is always...umm...never more than 95% correct 😉

PR #596, opened 2018-12-18
Handle duplicate values in criteria detail fields in upload template

tmy1313:
OCD-2593


PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -1,9 +1,12 @@
 
-## Version 20.0.1
+## Version 20.1.0
 _Date TBD_
 
 ### Bugs Fixed
 * Save "Developer identified targeted users" on upload/confirm
 * Update invalid error and warning messages for upload file
 
+### New Features
```
andlar: Request putting this section above the Bugs section. Usually have the sections in order of major-minor-bug

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -1,9 +1,12 @@
...
+* When validating pending listings, check for duplicate values, remove the duplicate value, and provide a warning message that the duplicate was removed
```
andlar:
```suggestion
 * When validating pending listings, check for duplicate values, remove the duplicate value(s), and provide a warning message about the removal
```

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -266,6 +266,25 @@ listing.upload.couldNotParse=Could not read uploaded file: "%s".
 listing.chplProductNumber.notUnique=The expected CHPL Product Number %s must be unique among all other certified products but one already exists with this value.
 listing.chplProductNumber.changedNotUnique=The CHPL Product Number has changed. The new CHPL Product Number "%s" must be unique among all other certified products but one already exists with this value.
 
+listing.duplicateTestingLab=Listing contains duplicate Testing Lab: '%s'.  The duplicates have been removed.
```
andlar: Our other errors messages don't have the `  ` double space after a period. While I know they tend to get ignored in most displays, I'd like to stay consistent

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/duplicate/TestStandardDuplicateReviewer.java
```diff
@@ -0,0 +1,114 @@
...
+                if (dto1.getName() != null && dto2.getName() != null) {
+                    return dto1.getName().equals(dto2.getName());
+                } else {
+                    return false;
+                }
+            }
+        };
+    }
+
+    //    public void review(PendingCertifiedProductDTO listing, final PendingCertificationResultDTO certificationResult) {
```
andlar: Probably don't need all this commented out stuff, right?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/duplicate/TestToolDuplicateReviewer.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return warnings;
+    }
+
+    private BiPredicate<PendingCertificationResultTestToolDTO, PendingCertificationResultTestToolDTO> getPredicate() {
+        return new BiPredicate<PendingCertificationResultTestToolDTO, PendingCertificationResultTestToolDTO>() {
+            @Override
+            public boolean test(final PendingCertificationResultTestToolDTO dto1,
+                    final PendingCertificationResultTestToolDTO dto2) {
+                if (dto1.getName() != null && dto2.getName() != null
```
andlar: These are all a bunch of `&&`,s right? So it's:
```
if (a & b & c & d)
return e & f
else
return false
```
Why not
```
 return (a & b & c & d & e & f)
```
where each of a, b, c, ... are the clauses you have?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/edition2014/duplicate/QmsStandard2014DuplicateReviewer.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private BiPredicate<PendingCertifiedProductQmsStandardDTO, PendingCertifiedProductQmsStandardDTO> getPredicate() {
+        return new BiPredicate<PendingCertifiedProductQmsStandardDTO, PendingCertifiedProductQmsStandardDTO>() {
+            @Override
+            public boolean test(final PendingCertifiedProductQmsStandardDTO dto1,
+                    final PendingCertifiedProductQmsStandardDTO dto2) {
+                if (dto1.getName() != null && dto2.getName() != null) {
+                    return dto1.getName().equals(dto2.getName());
+                } else {
+                    return false;
```
andlar: Similar comment to above.

I skipped the add'l s/w one as there was an if - else - else format, and while that might be consolidated I wasn't sure.

In any case, these straightforward ones could be combined; won't be mentioning them going further (if there are any)

chpl/chpl-resources/src/main/resources/errors.properties
kekey1: I'm sure we were all taught to double-space after a period but that's not how the cool kids do it anymore!

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/duplicate/TestToolDuplicateReviewer.java
tmy1313: I typically leave the null checks separate from the return value simply for readability, but it doesn't really matter to me.

What would be your thoughts on using this Java 7 Objects.equals method.  https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object)

I need to make sure the logic is the same, but thought it was worth a look.

andlar: It looks like the Object.equals(a, b), when neither a nor b are null, simply calls a.equals(b). If that's the case then I think that could work here, but we'd probably need to make sure our "a.equals" override, if it exists, is correct, and if it doesn't, works. Also, Object.equals(a, b), for both a and b as null, would return true. Would we want that here? If we can guarantee that at least one of a or b is non-null, and the a.equals() method works, then I think that would be okay here, but I can't tell that from here

kekey1: I generally very much like the idea of overriding equals (and hashCode). The only thing I would worry about in this case is are there times when just the test tool name is filled in or just the test tool id? The answer might be no, but just make sure you test various cases of a user uploading a file with a bad test tool name, good test tool name, and then when a user goes through the confirm dialog and adds a test tool are the id and name both filled in or just the id? I think we ran into something similar with the macra measures recently where the name or id was null in one use case but not another. Also just want to make sure there are no unintended consequences when overriding equals. It is possible that method gets called on objects a very low level so if for instance we have a Set<PendingCertifiedProductTestToolDTO> anywhere, the equals() behavior could affect that. All that to say - I like the idea, go for it, be careful!

PENDING_RELEASE_NOTES.md
```diff
@@ -3,6 +3,7 @@ _Date TBD_
 
 ### New Features
 * Change surveillance and sed report filenames to include timestamp when the file was last generated.
+* When validating pending listings, check for duplicate values, remove the duplicate value, and provide a warning message that the duplicate was removed
```
andlar: Shouldn't this be in PENDING_PENDING?

PR #597, opened 2018-12-31
Ocd 2474 - update error and warning messages

brianlindsy:


chpl/chpl-api/src/test/java/gov/healthit/chpl/listener/CacheRefreshTest.java
```diff
@@ -348,11 +348,7 @@ public void testUpdateListingStatusRefreshesCache() throws EntityRetrievalExcept
 
         ListingUpdateRequest updateRequest = new ListingUpdateRequest();
         updateRequest.setListing(listingToUpdate);
-        try {
-            cpController.updateCertifiedProduct(updateRequest);
-        } catch (ValidationException e) {
-            assertEquals(e.getErrorMessages().size(), 3);
```
andlar: This test was previously expecting to have some errors on the upload call, and then just assert that they were the same as expected, and continue, right? Given that the test doesn't really care about the errors on the update then that made sense. Now, though, won't it fail when it has any errors on the update? If so, then the test will fail for reasons that have nothing to do with what it's testing. Why not, instead of throwing a ValidationException, just catch it and ignore it?

Same comments with below changes related to the update calls

chpl/chpl-service/src/test/java/gov/healthit/chpl/manager/impl/ActivityManagerTest.java
```diff
@@ -87,7 +87,7 @@ public static void setUpClass() throws Exception {
 
     @Test
     @Transactional
-    @Rollback
+    //@Rollback
```
andlar: Why take out the Rollback? And why leave it commented out?

chpl/chpl-service/pom.xml
```diff
@@ -424,14 +424,6 @@
                             <source>src/main/java/gov/healthit/chpl/domain/TransparencyAttestationMap.java</source>
                         </sources>
                     </configuration>
-                    <executions>
```
kekey1: Don't we need schemagen to generate the download file XSD? Why remove it?

brianlindsy:
Is there documentation somewhere on how to format commit messages? I've been adding --no-verify to my commit messages because git commit -m '' does not work anymore

chpl/chpl-service/pom.xml
```diff
@@ -425,13 +425,13 @@
...
                     </configuration>
                     <executions>
-                        <execution>
-                            <id>schemagen</id>
-                            <goals>
-                                <goal>schemagen</goal>
-                            </goals>
-                        </execution>
-                    </executions>
+						<execution>
```
andlar: Looks like a mix of tabs and spaces in this file

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,10 @@
+## Version 20.1.0
+_Date TBD_
+
+### New Features
+* Update invalid input fields
```
andlar: This is changing some errors to warnings, or warnings to errors, and maybe fixing some typos, right? RELEASE NOTES are meant to be read by the public so should have a description of the changes that a layperson would understand. Also shouldn't have an empty Bugs Fixed section

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,8 @@
+## Version 20.1.0
```
andlar: There should be an empty line at the top of this file

chpl/chpl-service/pom.xml
andlar: I've lost track. Are we still using schemagen? Or is that no longer needed due to the modifications we made to the XML generation system?

kekey1: We still use it to generate the XSD. I confirmed by checking Bamboo and see this line in one of the steps:
`cp chpl/chpl-service/target/generated-resources/schemagen/schema1.xsd /opt/tomcat8/tomcat-latest/files
`

chpl/chpl-api/src/test/java/gov/healthit/chpl/listener/CacheRefreshTest.java
brianlindsy: I think that's the way most of the tests with update calls are written, the test will fail because of any validation error that occurs during the update call. If something happens inside that method that causes the exception to be thrown then the update will not happen and the test will fail anyway, whether or not we catch and ignore it inside the unit test because of assertions farther down the test. Maybe I'm misunderstanding your comment though.

andlar: Line 354 expects to find three errors in the update that happened on line 352, right? And then it won't throw any other errors or whatever, and processing will continue with the rest of the method. But if you remove that try/catch block, then when the update happens, any error in the update will be thrown, which causes the test to fail, but we're not testing the update in this test; we're testing the caching refresh thing.

We want to make our tests only fail when the thing we're testing goes wrong. One way to do that would be to put the update in a try block, catch any validationerrors and ignore them, and then continue. Another could be to mock the update call, though if the update call is supposed to do the cache refresh then that wouldn't work.

Does the cache refresh happen as part of the update?

brianlindsy: That try catch block on 354 was something I added because of the warnings that got turned into error messages. I originally thought that I could just catch those error messages and just move on with the rest of the test. For some reason, the test initially passed on my local machine with the try-catch, but when I pushed to DEV that test failed because that update to the certification status to Retired from Active did not happen because the validation error was thrown inside of the update call before the actual update. So, I then made a change to the test data to include UCD Processes so that the validation errors would not be thrown and therefore would not need to catch those errors, which I think is the right way to do this.

andlar: Okay. If the tests are passing now I don't want to hold this up any longer. I'm going to resolve this conversation and approve the PR. It'll be up to @kekey1 to resolve her conversation then

PR #598, opened 2019-01-03
Ocd 2635 - Show better error message when user uploads Test Tasks with IDs that are too long

brianlindsy:
Includes changes from OCD-2474...


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/FieldLengthReviewer.java
```diff
@@ -26,6 +28,14 @@ public void review(CertifiedProductSearchDetails listing) {
         if (listing.getVersion() != null && !StringUtils.isEmpty(listing.getVersion().getVersion())) {
             checkField(listing, listing.getVersion().getVersion(), "productVersion");
         }
+        if (listing.getSed() != null && listing.getSed().getTestTasks() != null) {
+            for(TestTask task : listing.getSed().getTestTasks()) {
```
andlar: Checkstyle should complain about not having a space between `for` and `(`

chpl/chpl-service/src/test/java/gov/healthit/chpl/validation/listing/reviewer/ValidDataReviewerTest.java
```diff
@@ -33,12 +43,17 @@
             + PrivacyAndSecurityFrameworkConcept.getFormattedValues();
     private static final String BAD_ADDL_SOFTWARE_ERROR =
             "No CHPL product was found matching additional software CHP-12345 for " + D_1;
+    private static final String BAD_LENGTH_PARTICIPANT_ID = "You have exceeded the max length for the Participant Identifier.";
+    private static final String BAD_LENGTH_TASK_ID = "You have exceeded the max length for the Task Identifier.";
```
andlar: Are these what the errors are that are shown to the user? I don't see a change to the errors.properties, so am assuming you're using what was already in there. In any case, given the fact that users sometimes have a lot of tasks and/or participants I think it'd be useful to add, for each error, the ID that was bad as well as the maximum length value that's allowed. Something like `You have exceeded the maximum length, 20 characters, for the Task Identifier "A task that has a lot of characters in the name"`

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -186,51 +186,51 @@ listing.criteria.testFunctionalityCriterionMismatch=In Criteria %s, Test Functio
...
+listing.vendorName.maxlength=You have exceeded the max length, %s characters, for the Vendor Name %s.
+listing.productName.maxlength=You have exceeded the max length, %s characters, for the Product Name %s.
+listing.vendorVersion.maxlength=You have exceeded the max length, %s characters, for the Product Version %s.
+listing.certificationEdition.maxlength=You have exceeded the max length, %s characters, for the Certification Edition %s.
+listing.acbCertificationId.maxlength=You have exceeded the max length, %s characters, for the ACB Certification ID %s.
+listing.certifyingAcb.maxlength=You have exceeded the max length, %s characters, for the Certifying ACB %s.
+listing.testingAcl.maxlength=You have exceeded the max length, %s characters, for the Testing ACL %s.
+listing.vendorStreetAddress.maxlength=You have exceeded the max length, %s characters, for the Vendor Street Address %s.
+listing.vendorStreetAddressTwo.maxlength=You have exceeded the max length, %s characters, for the Vendor Street Address Two %s.
+listing.vendorState.maxlength=You have exceeded the max length, %s characters, for the Vendor State %s.
```
kekey1: I know the code says "Vendor" sometimes but it should actually be "Developer".

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/FieldLengthReviewer.java
```diff
@@ -26,27 +30,36 @@ public void review(CertifiedProductSearchDetails listing) {
...
+                }
+            }
+        }
     }
 
     private void checkField(final CertifiedProductSearchDetails product, final Object field, final String errorField) {
         if (field instanceof Long) {
             Long fieldCasted = (Long) field;
             if (fieldCasted.toString().length() > getMaxLength("maxLength." + errorField)) {
                 CertifiedProductSearchDetails productCasted = (CertifiedProductSearchDetails) product;
```
kekey1: I know this isn't new code, but just wondering why the cast to CertifiedProductSearchDetails is being done since the passed-in "product" is already of that type? Can you clean this up a bit?

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -187,23 +187,23 @@ listing.criteria.testFunctionalityNotFoundAndRemoved=Criteria %s contains an inv
...
-listing.vendorStreetAddressTwo.maxlength=You have exceeded the max length, %s characters, for the Vendor Street Address Two %s.
-listing.vendorState.maxlength=You have exceeded the max length, %s characters, for the Vendor State %s.
-listing.vendorCity.maxlength=You have exceeded the max length, %s characters, for the Vendor City %s.
-listing.vendorZip.maxlength=You have exceeded the max length, %s characters, for the Vendor Zip %s.
-listing.vendorWebsite.maxlength=You have exceeded the max length, %s characters, for the Vendor Website %s.
-listing.vendorEmail.maxlength=You have exceeded the max length, %s characters, for the Vendor Email %s.
-listing.vendorPhone.maxlength=You have exceeded the max length, %s characters, for the Vendor Phone %s.
-listing.vendorContactName.maxlength=You have exceeded the max length, %s characters, for the Vendor Contact Name %s.
-listing.vendorState.maxlength=You have exceeded the max length, %s characters, for the Vendor State %s.
+listing.DeveloperStreetAddress.maxlength=You have exceeded the max length, %s characters, for the Developer Street Address %s.
```
kekey1: When I looked at the code before, I thought that the error message was looked up dynamically using a property of the class like "listing." + propertyName + ".maxlength" and I didn't like that because I thought what if you change the property name then the error message will stop showing up and it would be super easy to forget to change the error message name. This is a good illustration of how that can happen. I don't think this change will end with the error message actually showing up because the property on the java object isn't called "DeveloperStreetAddress" (at a minimum the case is not correct, not sure if it's called developer or vendor in there though). I don't have a better suggestion other than to not look up the error property names dynamically.... I don't think doing it that way is beneficial for any reason aside from less typing.

brianlindsy: Hm, I see what you're saying. I did run the tests locally and they passed, not sure if they'll pass on Jenkins, though. Even if they are not dynamically generated, wouldn't we have to go into the code and change the name as well?

kekey1: If this isn't a good time/ticket for making a refactoring type change I think at least if you're changing the error message names you have to change where they are referenced in the java code. I didn't see anything changed where "vendorWebsite" was referenced, for example, but it got changed to "DeveloperWebsite" below.

kekey1: And I don't see any unit tests written for gov.healthit.chpl.validation.pendingListing.reviewer.FieldLengthReviewer which is where some of these are referenced.

brianlindsy: Fixed the property names in the Java codes and added the field length reviewer tests, all tests pass on Jenkins as well.

PR #603, opened 2019-01-09
Ocd 2648 - display better error when user uploads bad targeted user data

brianlindsy:


chpl/chpl-service/src/test/java/gov/healthit/chpl/validation/listing/reviewer/FieldLengthReviewerTest.java
```diff
@@ -91,4 +94,26 @@ public void testBadDeveloperMaxLength_HasErrors_Pending() {
...
+    }
+    
+    @Test
+    public void testTargetedUserMaxLength_HasNoErrors_Pending() {
+        PendingCertifiedProductDTO listing = mockUtil.createPending2014Listing();
+        PendingCertifiedProductTargetedUserDTO tu = new PendingCertifiedProductTargetedUserDTO();
+        tu.setName("This name is not too long");
+        listing.getTargetedUsers().add(tu);
+        fieldLengthReivewerPending.review(listing);
+        assertFalse(listing.getWarningMessages().contains(BAD_LENGTH_TARGETED_USER));
```
kekey1: Thanks for adding tests! If there were to be a warning about this targeted user, it would not be the one being tested for.. it would be something like "You have exceeded the max length, 300 characters, for the Targeted User This name is not too long, which has been deleted". Can you change to test for the actual expected warning message?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/FieldLengthReviewer.java
```diff
@@ -1,68 +1,99 @@
...
-        checkField(listing, listing.getProductName(), "productName");
-        checkField(listing, listing.getProductVersion(), "productVersion");
+        checkField(listing, listing.getCertificationEditionId(), "certificationEdition", ERROR);
+        checkField(listing, listing.getAcbCertificationId(), "acbCertificationId", ERROR);
+        checkField(listing, listing.getCertificationBodyId(), "certifyingAcb", ERROR);
+        ArrayList<PendingCertifiedProductTargetedUserDTO> toRemove = new ArrayList<PendingCertifiedProductTargetedUserDTO>();
+        for (PendingCertifiedProductTargetedUserDTO tu : listing.getTargetedUsers()) {
+            checkField(listing, tu.getName(), "targetedUser", WARNING);
+            if (listing.getWarningMessages().contains(msgUtil.getMessage("listing.targetedUser.maxlength",
+                    String.valueOf(getMaxLength("maxLength.targetedUser")), tu.getName()))) {
```
kekey1: I am assuming this works when you upload something and that it was tested locally within the app... it just looks to me like it might not because the pending_certified_product_targeted_user table has a character limit of 300 on the targeted_user text so I'm confused as to how/when the targeted user is getting removed. It can't be put in the table if length > 300 characters so I feel like a db exception would have been thrown before this code is executed. I could be thinking about the workflow incorrectly so if it does work, that's fine. Otherwise I'm thinking there might also need to be a change to the pending table in the db to make the targeted user a text column that doesn't restrict length.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/FieldLengthReviewer.java
```diff
@@ -1,68 +1,99 @@
...
-    private void checkField(final PendingCertifiedProductDTO listing, final Object field, final String errorField) {
-        if (field instanceof Long) {
-            Long fieldCasted = (Long) field;
-            if (fieldCasted.toString().length() > getMaxLength("maxLength." + errorField)) {
-                PendingCertifiedProductDTO productCasted = (PendingCertifiedProductDTO) listing;
-                productCasted.getErrorMessages().add(msgUtil.getMessage("listing." + errorField + ".maxlength",
-                        String.valueOf(getMaxLength("maxLength." + errorField)), fieldCasted));
+    private void checkField(final PendingCertifiedProductDTO listing, final Object field, final String errorField,
+            String type) {
+        if (type.equals(ERROR)) {
```
andlar: The "if long vs if string" stuff is duplicated. Why not generate the message first, independent of whether it's an ERROR or a WARNING, and only at the end add an:
```
if type.equals(ERROR)
  listing.geterrormessages.add(message)
else
  listing.getwarningmessages.add(message)
```

PR #610, opened 2019-01-18
Ocd-2635 - upload error for too long test task and participant ids

brianlindsy:


chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2015Version1.java
```diff
@@ -377,6 +381,10 @@ protected int parseTestTasks(final CSVRecord record,
 
         PendingTestTaskEntity task = new PendingTestTaskEntity();
         task.setUniqueId(record.get(colIndex++).trim());
+        if (task.getUniqueId().length() > msgUtil.getMaxLength("maxLength.taskIdentifier")) {
+            pendingCertifiedProductEntity.getErrorMessages().add(msgUtil.getMessage("listing.taskIdentifier.maxlength",
```
andlar: Will these kinds of errors cause the upload to fail? Or appear as errors in the workflow?

brianlindsy: If the error is in the handler file, they will cause the upload to fail

andlar: Cool

PR #618, opened 2019-01-30
Ocd 2634

tmy1313:


PENDING_RELEASE_NOTES.md
```diff
@@ -3,7 +3,12 @@
 _Date TBD_
 
 ### New Features
+* Added ROLE_ONC user role. ROLE_ONC replaces the existing ROLE_ADMIN and all relevant accounts were converted to the new role.
```
andlar: Weren't these first few new lines already deployed to PROD?

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/scheduler/GetAllActionPermissions.java
```diff
@@ -0,0 +1,68 @@
...
+            }
+        } catch (Exception e) {
+            LOGGER.error(e);
+            return false;
+        }
+    }
+
+    private Boolean doesUserHavePermissionToJob(final ChplJob job) {
+        //Get the authorities from the job
+        if (job.getGroup() != null && job.getGroup().equals("chplJobs")) {
```
andlar: This method is only called from the `hasAccess` method, right? And from that method, only after we've already determined that the group == "chplJobs", so this if isn't needed, correct?

tmy1313: True - change made.

PENDING_RELEASE_NOTES.md
tmy1313: Update notes

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/schedule/ChplOneTimeTrigger.java
```diff
@@ -0,0 +1,27 @@
+package gov.healthit.chpl.domain.schedule;
+
+import java.io.Serializable;
+
+public class ChplOneTimeTrigger implements Serializable {
```
kekey1: Why doesn't this class need a name or group like the ChplTrigger class has? Purely based on the class names, it looks like it should be more of a hierarchy with parent class ChplTrigger and subclasses RepeatableChplTrigger and OneTimeChplTrigger or something like that. Would any refactoring like that be appropriate here or is it not really the right thing to do? I'm not as familiar with the objects as you are.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/SchedulerManagerImpl.java
```diff
@@ -166,20 +197,34 @@ public ChplTrigger updateTrigger(final ChplTrigger trigger) throws SchedulerExce
...
-                }
+            for (JobKey jobKey : scheduler.getJobKeys(groupEquals(group))) {
+                JobDetail jobDetail = scheduler.getJobDetail(jobKey);
+                jobs.add(new ChplJob(jobDetail));
+            }
+        }
+
+        //TODO: There is a problem using @PostFilter - some sort of ACL issue
+        //Until that is figured out...
+        Iterator<ChplJob> iterator = jobs.iterator();
```
kekey1: I thought you had a problem with PostFilter for the surveillance stuff but figured it out. Am I remembering that wrong or is this a different problem?

tmy1313: I completely forgot!  I'll update this to work with @PostFilter.

PR #625, opened 2019-02-05
Add JaCoCo plugin for test coverage

kekey1:
Also requires some Jenkins configuration to show coverage report in our individual jobs. See kekey_chpl-api "Post Build Actions" for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -260,6 +260,7 @@ pip-log.txt
 /chpl/deploy-webapp.sh
 /chpl/.recommenders/
 *-api.sh
+push-*.sh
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -260,6 +260,7 @@ pip-log.txt
 /chpl/deploy-webapp.sh
 /chpl/.recommenders/
 *-api.sh
+push-*.sh
```
andlar: These files aren't needed any more, so shouldn't be in the gitignore thing. All of the push scripts were merged into the single push.sh one

PR #628, opened 2019-02-06
feat: change InfoGard's name to UL LLC

andlar:
[#OCD-2340]


chpl/chpl-resources/src/main/resources-dev/jobs.xml
```diff
@@ -303,7 +303,7 @@
                     </entry>
                     <entry>
                         <key>acb</key>
-                        <value>Drummond Group☺ICSA Labs☺InfoGard☺SLI Compliance</value>
+                        <value>Drummond Group☺ICSA Labs☺UL LLC☺SLI Compliance</value>
```
kekey1: It's interesting that you had to change these jobs files. The change you made in the website project pretty much gets us to the point where the website will keep working fine and display data as expected (barring stale cache data) if an ACB changes their name. The scheduled jobs however would not keep working. I think it's a little dangerous to even let them change their name via the UI edit dialog because of that. Suggest revisiting how we handle ACB permissions in the scheduled jobs (in a different and larger ticket probably) since theoretically any ACB could go in and change their name at any time.. even if something small like adding punctuation.  We could live with old ACB names in comments and unit tests because that wouldn't actually break anything but it would here in the jobs. Another note - as I was thinking about this I looked at the quartz jobs tables and noted that the ACB name is appended to the trigger name, so we have triggers in prod like "stephanie_eckgren@ul_comInfoGard". Does that need to be changed? Is it used anywhere in any way that it would matter?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/SchedulerManagerImpl.java
```diff
@@ -238,6 +238,22 @@ public void retireAcb(final String acb) throws SchedulerException, ValidationExc
         }
     }
 
+    @Override
+    @PreAuthorize("hasAnyRole('ROLE_ADMIN', 'ROLE_ONC', 'ROLE_ACB')")
```
tmy1313: Should this use the new security model?

andlar: It certainly could be, but I wasn't prepared to make an entire transition of the manager. There's another discussion we need to have regarding Quartz jobs and how those are handled by security too, and I'm hoping both of those can be resolved at the same time

PR #635, opened 2019-02-18
Ocd 2636 - Disable "No-Cache" rule on appropriate API endpoints

tmy1313:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SchedulerController.java
```diff
@@ -96,10 +105,11 @@ public void deleteTrigger(@PathVariable("triggerGroup") final String triggerGrou
             notes = "Security Restrictions: ROLE_ADMIN, ROLE_ONC, or ROLE_ACB and have administrative "
                     + "authority on the specified ACB.")
     @RequestMapping(value = "/triggers", method = RequestMethod.GET, produces = "application/json; charset=utf-8")
-    public @ResponseBody ScheduleTriggersResults getAllTriggers() throws SchedulerException {
+    public @ResponseBody ScheduleTriggersResults getAllTriggers(final HttpServletResponse response) throws SchedulerException {
```
kekey1: Why was the response variable added? The diff doesn't show it being used in the method.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
```diff
@@ -561,6 +578,7 @@ public void streamSEDDetailsDocumentContents(final HttpServletResponse response)
                     + "authority on the ACB that certified the product.")
     @RequestMapping(value = "/{certifiedProductId}", method = RequestMethod.PUT,
     produces = "application/json; charset=utf-8")
+    @CacheControl(policy = CachePolicy.PUBLIC, maxAge = CacheMaxAge.TWELVE_HOURS)
```
kekey1: Is this right for an update to be cached? Seems a little odd. If it's right, I think that the update listing method adds a custom header which indicates to the UI that the CHPL ID may have changed and causes the little orange popup in the top right of the screen to show up. I was just wanting to ensure you tested that still works since the CacheControl annotation is fussing with the response headers.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SchedulerController.java
tmy1313: Good catch - that was left from a previous attempt at a fix and should have been removed.  I'll take remove of it.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
tmy1313: This was a mistake - copied the code to the wrong method...

PR #638, opened 2019-02-20
Ocd 2559 - add error message for unparseable cert date

andlar:


andlar:
I looked at the possibility of adding unit tests, but we don't currently have _any_ tests of the upload manager. To do it right we'd probably have to build a CSV in memory and pass that to the handler, and that seemed like it was way outside of scope. Maybe if @tmy1313's builder things were working that would be possible

kekey1:
That's fine. We have some tests in the API project that actually upload CSV files but that might be exercising a lot more code than is needed.

andlar:
Yeah, it seems odd to write tests that are targeted the chpl-api project but exercising code in the depths of the chpl-service. Especially if we're trying to call them "unit tests". ;)

PR #639, opened 2019-02-21
Ocd 2697 - round some values

andlar:
Might change content based on PO's comments, but at least the rounding is there


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/edition2015/RequiredData2015Reviewer.java
```diff
@@ -323,10 +323,18 @@ public void review(final PendingCertifiedProductDTO listing) {
                                     try {
                                         Integer.valueOf(task.getTaskPathDeviationObserved());
                                     } catch (final Exception e) {
-                                        listing.getErrorMessages().add(
-                                                msgUtil.getMessage("listing.criteria.badTestTaskNumber",
-                                                        task.getUniqueId(), "Task Path Deviation Observed",
-                                                        task.getTaskPathDeviationObserved()));
+                                        try {
```
kekey1: Something about nested try/catch that I don't love.. I'm not sure if there is a java best-practice type reason against it other than maybe readability though. Would you consider changing this to have the outer try/catch where you try to parse the value as an Integer, then following the catch a check if the task.getTaskPathDeviationObserved value is null? If it's still null you could assume that the value didn't parse correctly as an integer and inside that if have the rest of the code that tries to parse as float, round, give warning.

chpl/chpl-service/src/test/java/gov/healthit/chpl/validation/pendingListing/reviewer/PendingListingRequiredData2015ReviewerTest.java
```diff
@@ -291,6 +305,29 @@ public void testNullNumbersInTestTask() {
         assertTrue(hasTestTaskNumberMissingMessage(listing, "Task ID", "170.315 (a)(1)", "Task Time Standard Deviation"));
     }
 
+    @Test
+    public void testRoundedNumbersInTestTask() {
```
kekey1: Thanks for adding a test!

chpl/chpl-service/src/test/java/gov/healthit/chpl/validation/pendingListing/reviewer/PendingListingRequiredData2015ReviewerTest.java
```diff
@@ -322,6 +371,10 @@ private Boolean hasGenericTestTaskErrorMessage(final PendingCertifiedProductDTO
         return false;
     }
 
+    private String formatMessage(final String message, final String a, final String b, final String c, final String d) {
```
kekey1: Instead of another quite similar method here could you use java's variable arguments thing? It's like formatMessage(String... args) I have done something similar to what I'm thinking in the ErrorMessageUtil class. If it's not feasible that's fine but as I'm glancing over this it seems like it could work.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/edition2015/RequiredData2015Reviewer.java
andlar: So the logic flow you're talking about is sort of
```
try parseInt // if successful, done
catch exception
  if val is null
    give "missing value error"
  else
    try round(parseFloat)
    give "rounding warning"
    catch exception
      give "could not parse error"
```


Our usual check workflow is sort of

```
if null
  give "missing value error"
else
  check logic of data
```

What I'm doing is expanding on that slightly

```
if null
  give "missing value error"
else
  try parseInt // if successful, done
  catch exception
    try round(parseFloat(
    give "rounding warning"
    catch exception
      give "could not parse error"
```

I think that's the same structure, but it moves the null check after the first parseInt thing. Given that we usually do null checks first I'd lean away from that strategy.

If I'm misinterpreting your logic flow please let me know

kekey1: I definitely should have used pseudocode.
`if null
  give "missing value error"
else
  try parseInt // if successful, done
  catch exception
      log error could not parse value as int

  if value is still null //it was not able to be parsed as an integer
      try round(parseFloat(
          give "rounding warning"
      catch exception
          give "could not parse error"`

andlar: Ah, I see. The only issue with that option is that I'm not actually saving the value that is returned by the parseInt. I _could_, but it's not used anywhere else, so I didn't see a reason to. Without saving it I'm basically just jumping directly into the round(parseFloat step

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/edition2015/RequiredData2015Reviewer.java
```diff
@@ -337,10 +345,18 @@ public void review(final PendingCertifiedProductDTO listing) {
...
+                                                    msgUtil.getMessage("listing.criteria.roundedTestTaskNumber",
+                                                            task.getUniqueId(), "Task Path Deviation Optimal",
+                                                            task.getTaskPathDeviationOptimal(), String.valueOf(val)));
+                                        } catch (final Exception ex) {
+                                            listing.getErrorMessages().add(
+                                                    msgUtil.getMessage("listing.criteria.badTestTaskNumber",
+                                                            task.getUniqueId(), "Task Path Deviation Optimal",
+                                                            task.getTaskPathDeviationOptimal()));
+                                        }
                                     }
```
tmy1313: Is there anyway to convert this to a function, to reduce the repetitive code?  
```
private void functionName(String origValue, String badErrorMessage, String uniqueId, String description, String roundedErrorMessage)
```
If you don't think so, I'll approve it, since it is correct.

PR #640, opened 2019-02-25
Ocd 2652 - allow impersonation

andlar:
Haven't implemented the new security stuff on the /impersonate calls, because I didn't want to make the merge with the updated chpl-auth project any harder than it's already going to be. Would like review on this anyway, though it won't be ready to deploy until I get the UI part ready too


kekey1:
Stil reviewing but one question I have is how this will work if a user is impersonating another user then kicks off some job that runs in the background (meaningful use upload or surveillance upload). There is some code in RunnableJobFactory that runs a background job as a specific user. Not sure if we care, if it matters, etc, but probably is worth at least testing one of those jobs to see how it works if kicked off by an impersonating user.

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/authentication/UserAuthenticator.java
```diff
@@ -346,4 +339,55 @@ public void setJwtAuthor(final JWTAuthor jwtAuthor) {
...
+    /**
+     * Compare acting user's role with targeted user's role.
+     * ROLE_ADMIN > all other roles
+     * ROLE_ADMIN is not > ROLE_ADMIN
+     * ROLE_ONC > all roles except ROLE_ADMIN and ROLE_ONC
+     * @param a acting user's ROLE
+     * @param t targeted user's ROLE
+     * @return true iff a > t
+     */
+    private Boolean isGreaterRole(final String a, final String t) {
```
kekey1: Could you make these variable names more descriptive? Maybe actingUserRole and targetedUserRole or something?

andlar: Good news. This entire set of methods were removed with the transition to the new security thing

andlar:
@kekey1 I finally got around to testing the upload of surveillance. It does do it as the impersonated user, so that user's name/email is used for the notification of the job being done.

That said, the "last modified user" stored in the database is the impersonating user, as we expect.

I _think_ this could count as acceptable behavior, but am up for a discussion if you'd like

PR #645, opened 2019-02-28
Ocd 2562

andlar:


andlar:
I don't care which way we go with it, but all of the other emailBuilder.config.SOMETHING parameters have periods separating the items, and that's what I have in my tomcat (not sure why)

kekey1:
> I don't care which way we go with it, but all of the other emailBuilder.config.SOMETHING parameters have periods separating the items, and that's what I have in my tomcat (not sure why)

Ok - I'll approve it if you want to create an MR and make a note to change the setenv.sh script on each system!

chpl/chpl-auth/src/main/java/gov/healthit/chpl/auth/EmailOverrider.java
```diff
@@ -121,6 +121,6 @@ private String getEmailDomain(final String someEmail) {
     }
 
     private String getForwardToEmail() {
-        return env.getProperty("emailBuilder_config_forwardAddress");
+        return env.getProperty("emailBuilder.config.forwardAddress");
```
tmy1313: The "_" was used because on the servers we set these values in the `setenv.sh`.  I found that environment variables with "." do not work in the *nix environments.
https://unix.stackexchange.com/questions/93532/exporting-a-variable-with-dot-in-it

andlar: Well drat. I was hoping there wasn't a good reason for this...

PR #646, opened 2019-02-28
OCD-2727 Split Developer

kekey1:


kekey1:
I'm open to that... looking at the split method in the DeveloperController, my intention of the validation that is happening first is to ensure someone calling the endpoint has provided all of the necessary fields (otherwise it's a Bad Request which I don't really think the manager should know about?). The next bit of code converts the domain object that both the UI and API endpoints know about into a DTO which is what the manager/service should know about. After that it's returning a response. 
The bit that I do think could be moved into the manager is this
`        //check developer fields
        Set<String> devErrors = creationValidator.validate(splitRequest.getNewDeveloper());
        if (devErrors != null && devErrors.size() > 0) {
            throw new ValidationException(devErrors, null);
        }`
so I can work on moving that specific validation. If you had other specific ideas in mind for what to move let me know!

kekey1:
@tmy1313 I refactored as I described.. please re-review when you get a moment.

PR #652, opened 2019-03-06
Ocd 2652 - Retire test tools

tmy1313:
QA bug fix

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/TestToolReviewer.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/TestToolReviewer.java
@@ -60,11 +60,6 @@ public class TestToolReviewer implements Reviewer {
                                 listing.getWarningMessages()
                                         .add(msgUtil.getMessage("listing.criteria.retiredTestToolNotAllowed",
                                                 testTool.getTestToolName(), cert.getNumber()));
-                            } else {
-                                listing.getErrorMessages()
-                                        .add(msgUtil.getMessage("listing.criteria.testToolNotFoundAndRemoved",
-                                                cert.getNumber(), testTool.getTestToolName()));
-                                testToolIter.remove();
                             }
                         }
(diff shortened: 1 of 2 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/TestToolReviewer.java
```diff
@@ -60,7 +60,7 @@ public void review(final CertifiedProductSearchDetails listing) {
                                 listing.getWarningMessages()
                                         .add(msgUtil.getMessage("listing.criteria.retiredTestToolNotAllowed",
                                                 testTool.getTestToolName(), cert.getNumber()));
-                            } else {
+                            } else if (tt == null) {
```
andlar: I'm confused. Assuming this is a "non-pending" reviewer, then if the TT is `!null and retired`, we give a warning; great. When would we give an error? If it's null? Are TTs required? If that's the case, then the error message with the name "testToolNotFoundAndRemoved" is either the wrong message or named badly.

If this _is_ a pending reviewer, then we're still not removing TTs; we should just give them an error and make them remove it themselves, which also doesn't match the error name

tmy1313: It's based on the original logic that was in the reviewer.  I have no idea why there was existing logic to look the test tool up by it's name.  I can remove it, but since I wasn't sure why it was there, I didn't remove it.

kekey1: I'm just trying to think why this might have been there... if someone edited a listing via the API and only supplied the test tool name (no id) maybe? I guess we could reasonably expect them to provide an ID so we wouldn't have to look it up in this way. Or could it be leftover from when we were inserting test tools if they weren't already found (if we ever did that? I know we inserted some things but I'm not sure if test tools were one of them)

PR #659, opened 2019-03-12
Ocd 2683 - CHPL managed user/ATL relationships

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/UserTestingLabMapDAOImpl.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        if (entity.getDeleted() == null) {
+            entity.setDeleted(false);
+        }
+        entityManager.persist(entity);
+        return entity;
+    }
+
+    private UserTestingLabMapEntity update(UserTestingLabMapEntity entity) {
+        entity.setLastModifiedUser(getUserId(User.SYSTEM_USER_ID));
```
kekey1: Should this be the system user id? Is it possible there is an actual logged in user updating this relationship and their ID should be here but default to the system id if null?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/UserTestingLabMapDAOImpl.java
```diff
@@ -0,0 +1,163 @@
...
+
+        String queryStr = "SELECT atl from TestingLabEntity atl " + "LEFT OUTER JOIN FETCH atl.address "
+                + "WHERE (atl.id = :entityid)" + " AND (atl.deleted = false)";
+
+        Query query = entityManager.createQuery(queryStr, TestingLabEntity.class);
+        query.setParameter("entityid", entityId);
+        List<TestingLabEntity> result = query.getResultList();
+
+        if (result == null || result.size() == 0) {
+            String msg = String.format(messageSource.getMessage(new DefaultMessageSourceResolvable("acb.notFound"),
```
kekey1: Could you use the ErrorMessageUtil here? Also should the message be something about atl not found instead of acb.notFound?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/UserTestingLabMapDTO.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.Serializable;
+
+import gov.healthit.chpl.auth.dto.UserDTO;
+import gov.healthit.chpl.entity.UserTestingLabMapEntity;
+
+public class UserTestingLabMapDTO implements Serializable {
+    private static final long serialVersionUID = -4237553934302170475L;
+
+    private Long id;
+    private gov.healthit.chpl.auth.dto.UserDTO user;
```
kekey1: Just curious why the whole package name is needed... it looks like it's being imported above.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/InvitationManagerImpl.java
```diff
@@ -376,7 +367,8 @@ private void handleInvitation(final InvitationDTO invitation, final UserDTO user
         }
         // give them access to the invited atl
         if (userAtl != null) {
-            atlManager.addPermission(userAtl, user.getId(), BasePermission.ADMINISTRATION);
+            // atlManager.addPermission(userAtl, user.getId(), BasePermission.ADMINISTRATION);
```
kekey1: remove commented out code

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/UserPermissionsManagerImpl.java
```diff
@@ -120,4 +187,23 @@ public boolean evaluate(Object object) {
         return dtos.size() > 0;
     }
 
+    private Boolean doesUserTestingLabMapExist(Long atlId, Long userId) {
+        List<UserTestingLabMapDTO> dtos = userTestingLabMapDAO.getByUserId(userId);
+
+        if (dtos == null || dtos.size() == 0) {
+            // TODO: throw an exception throw exception...
+            LOGGER.error("Could not locate the UserTestingLabyMap object for Userid: " + userId + ", ATL: " + atlId);
```
kekey1: Do you need to throw an exception?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/UserTestingLabMapDAOImpl.java
tmy1313: This function gets the current user, but if there isn't a current for some reason (maybe running in the context of a Quartz job running), it will return `User.SYSTEM_USER_ID` .

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/TestingLabManagerImpl.java
```diff
@@ -75,8 +61,8 @@ public TestingLabDTO create(final TestingLabDTO atl)
         // Create the atl itself
         TestingLabDTO result = testingLabDAO.create(atl);
 
-        // Grant the admin user administrative permission to the ATL
-        addPermission(result, User.ADMIN_USER_ID, BasePermission.ADMINISTRATION);
+        // Grant the user administrative permission to the ATL
+        userPermissionsManager.addAtlPermission(result, Util.getCurrentUser().getId());
```
andlar: With the new impersonation thing in place, this should probably use `Util.getAuditId()` because as it stands it'd grand the permission to the Impersonated user, not the one doing the impersonation. When ROLEs change in an impersonation state, the impersonating user is the one the ROLEs apply to, so I think the authority on ATL (& ACB if it's not right) should be set up the same way

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/UserTestingLabMapDAOImpl.java
```diff
@@ -0,0 +1,171 @@
...
+
+        UserEntity user = null;
+
+        Query query = entityManager.createQuery(
+                "from UserEntity where (NOT deleted = true) " + "AND (user_id = :userid) ", UserEntity.class);
+        query.setParameter("userid", userId);
+        List<UserEntity> result = query.getResultList();
+
+        if (result == null || result.size() == 0) {
+            String msg = String.format(messageSource.getMessage(new DefaultMessageSourceResolvable("user.notFound"),
```
kekey1: Didn't notice this one before.. can you double check that errorMessageUtil is used everywhere in this class and the messageSource.getMessage thing is not used anywhere?

PR #663, opened 2019-03-18
OCD-2769: New activity metadata API endpoints

kekey1:
There are two TODOs left in the code that are there because I need the User/ATL relationship PR merged in order to use some of that code. Will not merge this in until that has happened. PRing now for comments on the rest.


chpl/chpl-resources/src/test/resources/META-INF/persistence.xml
```diff
@@ -148,7 +149,7 @@
             <property name="hibernate.format_sql" value="false" />
             <property name="hibernate.driver_class" value="org.postgresql.Driver" />
             <property name="hibernate.connection.datasource"
-                value="java:comp/env/jdbc/openchpl_test" />
+                value="java:comp/env/jdbc/openchpl" />
```
andlar: Shouldn't the test system connect to the test database?

chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/ActivityMetadataBuilderFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import gov.healthit.chpl.dto.ActivityDTO;
+
+/**
+ * Creates an appropriate metadata buidler object for the type of activity.
+ * @author kekey
+ *
+ */
+@Component
+public class ActivityMetadataBuilderFactory {
+    @Autowired
```
andlar: Should these be autowired in the constructor?

chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/ListingActivityMetadataBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+            return;
+        }
+        ListingActivityMetadata listingMetadata = (ListingActivityMetadata) metadata;
+
+        //parse listing specific metadata
+        CertifiedProductSearchDetails origListing = null;
+        if (dto.getOriginalData() != null) {
+            try {
+                origListing =
+                    jsonMapper.readValue(dto.getOriginalData(), CertifiedProductSearchDetails.class);
```
andlar: What happens if someone tries to get activity from a Listing that's really old, and has a fairly significant change in structure?

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/ActivityConceptEntity.java
```diff
@@ -13,18 +13,18 @@
 import gov.healthit.chpl.util.Util;
 
 @Entity
-@Table(name = "activity_class")
+@Table(name = "activity_concept")
```
andlar: Does there need to be a data-model change too?

kekey1: Nope. That table doesn't exist. That entity wasn't referenced from anywhere which is why it never mattered but it happened to have the same class name that I wanted to use.

chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/ListingActivityMetadataBuilder.java
kekey1: So I tested this with all of the activity data in the database. The Jackson JSON deserializer is actually very generous and surprised me greatly. Mostly it tries its best to fill in the fields that it can and I added some annotations to a couple of the domain object classes that help it ignore fields that are troublesome. And we don't actually need ALL the fields filled in - just a few that we care about for displaying the metadata. So I made sure those are all filled in even for listing activity from 2016. In one case I had to add a field (product name i think?) so that it could be filled in for old listing activity since we want to display it in the metadata but I annotated that field so that it will be ignored in other circumstances... it's not present in the details response and is annotated to be ignored when the XSD is generated.

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/ActivityConceptEntity.java
andlar: Hah! Okay. Works for me :)

chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/ListingActivityMetadataBuilder.java
andlar: I saw some annotations after writing this comment that seemed likely to be related. Good to hear it's not as terrible as I was expecting

chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/ActivityMetadataBuilderFactory.java
kekey1: Yes, addressed.

chpl/chpl-resources/src/test/resources/META-INF/persistence.xml
kekey1: Yes, leftover from testing against all the listing activity. Addressed.

andlar (approved):

tmy1313 (approved):

PR #667, opened 2019-03-20
Ocd 2797 - Uploaded surveillance jobs don't start

tmy1313:


tmy1313:
> This PR has a bunch of commits from a different ticket

I think I fixed this.

PR #671, opened 2019-03-25
OCD-2808: Add pending listing metadata API call

kekey1:


PENDING_RELEASE_NOTES.md
```diff
@@ -4,5 +4,6 @@ _Date TBD_
 
 ### New features
 * Modify security to use CHPL managed user/ATL relationship
+* Add new API call /certified_products/pending/metadata for high-level information about pending listings. /certified_products/pending is now deprecated.
```
andlar: I don't think we can deprecate the /certified_products/pending endpoint, as I still use that to get the details for a pending listing, using /certified_products/pending/{ pending ID }. Should I be using something else?

kekey1: No that's the right one to use. I'm thinking of those as two completely different endpoints, as they are two separate methods in the controller and two different URLs... /certified_products/pending (to get all pending listings the user has access to) and /certified_products/pending/{id} (to get details about a specific pending listing). The second one there is not deprecated. Suggestion for different wording so it's not confusing?

andlar: I think I'd been thinking about /pending and /pending/{id} as the same endpoint, but with an optional thing. Now that I think about it, that's not really the same thing; if it were optional it'd be in a query parameter. The UI-Router component in the UI project does allow for those IDs to be optional, which is probably what confused me

PR #672, opened 2019-03-25
 fix: user deletion issues

tmy1313:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -408,22 +409,36 @@ private String delete(final Long userId)
...
+            userPermissionsManager.deleteAllAcbPermissionsForUser(userId);
+            userPermissionsManager.deleteAllAtlPermissionsForUser(userId);
+        } else if (resourcePermissions.isUserRoleAcbAdmin()) {
+            List<CertificationBodyDTO> targetUserAcbs = resourcePermissions.getAllAcbsForUser(userId);
+            for (CertificationBodyDTO dto : resourcePermissions.getAllAcbsForCurrentUser()) {
+                if (isAcbInList(dto, targetUserAcbs)) {
+                    userPermissionsManager.deleteAcbPermission(dto, userId);
+                }
+            }
+        } else if (resourcePermissions.isUserRoleAtlAdmin()) {
```
kekey1: If a user has ROLE_ACB and ROLE_ATL I think this won't work as desired? I know that's super unlikely. But if they do, I think the if/else if's need to be changed around a bit.

PR #683, opened 2019-04-04
Ocd 2797

tmy1313:


andlar:
Could try adding a minute or two instead of just 250 ms; that might alleviate the problem, if it's a thread related thing

PR #691, opened 2019-04-23
Ocd 2755 - Improve Reports responsiveness - save filter

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/FilterManagerImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    @Override
+    @Transactional
+    public FilterDTO getByFilterId(Long filterId) throws EntityRetrievalException {
+        return filterDAO.getById(filterId);
+    }
+
+    private Set<String> validateForCreate(FilterDTO filterDTO) {
+        if (filterDTO == null) {
+            Set<String> errors = new HashSet<String>();
+            errors.add("No filter object was present to be saved.");
```
andlar: Should this be in the errors.properties file?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/FilterController.java
```diff
@@ -0,0 +1,108 @@
...
+    @RequestMapping(value = "/{filterId}", method = RequestMethod.DELETE,
+    produces = "application/json; charset=utf-8")
+    public @ResponseBody String deleteFilter(@PathVariable("filterId") final Long filterId) {
+        try {
+            FilterDTO filterDTO = filterManager.getByFilterId(filterId);
+            if (filterDTO != null) {
+                filterManager.delete(filterDTO);
+                return "{\"success\" : true}";
+            } else {
+                return "{\"success\" : false}";
```
kekey1: I think if the code executes this line it would mean that the filterId passed in wasn't valid and I think we should return an error code in that case. Bad Request? Maybe either throw an exception here that will be caught and handled in that way by the controller advice class or just send back the Bad Request status code yourself here.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/FilterController.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+        } catch (Exception e) {
+            LOGGER.error("Could not delete filterId: " + filterId, e);
+            return "{\"success\" : false}";
```
kekey1: Same comment as above although it probably shouldn't be a Bad Request code.. maybe Internal Server Error since it's just that there was some exception thrown and we don't know what it is?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/FilterDAOImpl.java
```diff
@@ -0,0 +1,158 @@
...
+import gov.healthit.chpl.entity.FilterEntity;
+import gov.healthit.chpl.entity.FilterTypeEntity;
+import gov.healthit.chpl.exception.EntityRetrievalException;
+
+@Repository("filterDAO")
+public class FilterDAOImpl extends BaseDAOImpl implements FilterDAO {
+
+    @Override
+    public FilterDTO update(final FilterDTO dto) throws EntityRetrievalException {
+        FilterEntity entity = new FilterEntity();
```
kekey1: What is this updating? I thought for an update method we normally would get the FilterEntity by ID and then fill in the rest of the values as you've done but this is creating a new FilterEntity and never setting the ID field. Does it actually end up inserting a new row or does it update an existing one?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/FilterDAOImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private FilterTypeEntity getFilterTypeEntityById(final Long filterTypeId) throws EntityRetrievalException {
+        Query query = entityManager.createQuery(
+                "from FilterTypeEntity where (NOT deleted = true) " + "AND (id = :filterTypeId) ",
+                FilterTypeEntity.class);
+        query.setParameter("filterTypeId", filterTypeId);
+        List<FilterTypeEntity> result = query.getResultList();
+
+        if (result == null || result.size() == 0) {
+            String msg = String.format(messageSource.getMessage(new DefaultMessageSourceResolvable("user.notFound"),
```
kekey1: Should be a different error message

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/FilterDAOImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Override
+    public FilterDTO update(final FilterDTO dto) throws EntityRetrievalException {
+        FilterEntity entity = new FilterEntity();
+        entity.setFilter(dto.getFilter());
+        entity.setLastModifiedUser(Util.getAuditId());
+        entity.setFilterType(new FilterTypeEntity());
+        entity.getFilterType().setId(dto.getFilterType().getId());
+        entity.getFilterType().setName(dto.getFilterType().getName());
+        entity.setUser(getUserEntityById(dto.getUser().getId()));
```
kekey1: You shouldn't really have to set the UserEntity here. The entity could be set up so that it has a field that maps to the user_id column and an additional field, marked insertable=false and updatable=false that lazily fetches the UserEntity. The way SurveillanceEntity maps certifiedProductId and certifiedProduct fields is a good example. If the FilterEntity is set up that way then you only have to set the userId field here and shouldn't need the getUserEntityById method at all. I'm not 100% sure on this but I think the way it is currently might end up updating the row for this user in the user table (although would not change anything)?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/FilterManagerImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    @Override
+    @Transactional
+    public FilterTypeDTO getFilterType(Long filterTypeId) throws EntityRetrievalException {
+        return filterDAO.getFilterTypeById(filterTypeId);
+    }
+
+    @Override
+    @Transactional
+    public FilterDTO getByFilterId(Long filterId) throws EntityRetrievalException {
+        return filterDAO.getById(filterId);
```
kekey1: Since there is no security on this, could anyone calling this code see any other user's filter? I know we don't have the concept of public and private but I was thinking they were all sort of private. Should we have permissions that only let a user get to a filter if they are the user_id on it? If this was intentional then that's fine.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/FilterManagerImpl.java
```diff
@@ -0,0 +1,125 @@
...
+        if (filterDTO == null || filterDTO.getId() == null) {
+            errors.add("No filter object was present to be saved.");
+        }
+        return errors;
+    }
+
+    private Set<String> validate(FilterDTO filterDTO) {
+        Set<String> errors = new HashSet<String>();
+        if (StringUtils.isEmpty(filterDTO.getFilter())) {
+            errors.add("Filter cannot be empty.");
```
kekey1: Same comments about using errors.properties

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/FilterDAOImpl.java
tmy1313: The update method is there just for completeness of the DAO.  It is not being called, therefore it was nor really tested.  Method has been updated.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/FilterController.java
```diff
@@ -0,0 +1,99 @@
...
+    private FilterManager filterManager;
+    private UserManager userManager;
+    
+    @Autowired
+    public FilterController(final FilterManager filterManager, final UserManager userManager) {
+        this.filterManager = filterManager;
+        this.userManager = userManager;
+    }
+    
+    @ApiOperation(value = "List all filters based on the filter type for the current user.",
```
andlar: I don't see a call to get the list of available filter types. We'll need one

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/FilterDAOImpl.java
```diff
@@ -0,0 +1,179 @@
...
+
+        if (result.size() > 1) {
+            throw new EntityRetrievalException("Data error. Duplicate filter id in database.");
+        } else if (result.size() == 1) {
+            entity = result.get(0);
+        }
+        return entity;
+    }
+
+    // private UserEntity getUserEntityById(final Long userId) throws EntityRetrievalException {
```
kekey1: Last thing I see - could you just remove this?

PR #692, opened 2019-04-24
OCD-2781: update activity metadata to handle merge/split

kekey1:


PENDING_RELEASE_NOTES.md
```diff
@@ -4,5 +4,6 @@ _Date TBD_
 
 ### New features
 * Updated security for edit/split/merge of developers, products, and versions
+* Updated product and version activity metadata to parse activity for merges and splits; better parsing of product and developer names if they have been deleted.
```
andlar: Is this a backwards compatibility breaking change?

kekey1: No I don't think so.. the format of the request and response is the same for the activity metadata endpoints

andlar: Oh! I see. You're using the single vs. list to determine which one to pull data from. Got it. I was half reading the code and late for a meeting at the same time :)

PR #693, opened 2019-04-25
OCD-2785: ACB/ATL metadata activity

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/ActivityMetadataBuilder.java
```diff
@@ -58,6 +60,12 @@ private ActivityMetadata createMetadataObject(final ActivityDTO dto) {
         case VERSION:
             metadata = new VersionActivityMetadata();
             break;
+        case CERTIFICATION_BODY:
+            metadata = new CertificationBodyActivityMetadata();
+            break;
+        case ATL:
```
andlar: Can we either do ACB & ATL, or CERTIFICATION_BODY & TESTING_LAB?

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/activity/GetActivityMetadataByAcbActionPermissions.java
```diff
@@ -0,0 +1,50 @@
...
+            Long acbId = (Long) obj;
+            CertificationBodyDTO acb = null;
+            try {
+                acb = acbDao.getById(acbId);
+                if (acb != null && acb.isRetired()) {
+                    LOGGER.warn("Non-admin user " + Util.getUsername()
+                    + " tried to see activity for retired ACB " + acb.getName());
+                    return false;
+                }
+            } catch (EntityRetrievalException ex) { }
```
tmy1313: I know this would be a very odd situation, but if we can't test if the ACB is retired, should we just return false in the `catch` ?  I guess it depends on how important it is that we don't return retired ACB information...

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/activity/GetActivityMetadataByAtlActionPermissions.java
```diff
@@ -0,0 +1,50 @@
...
+            Long atlId = (Long) obj;
+            TestingLabDTO atl = null;
+            try {
+                atl = atlDao.getById(atlId);
+                if (atl != null && atl.isRetired()) {
+                    LOGGER.warn("Non-admin user " + Util.getUsername()
+                    + " tried to see activity for retired ATL " + atl.getName());
+                    return false;
+                }
+            } catch (EntityRetrievalException ex) { }
```
tmy1313: Same as above.

PR #694, opened 2019-04-25
Ocd 2885  (OCD-2820 - Resolve phantom logged in issue)

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/auth/jwt/JWTConsumerRsaJoseJImpl.java
```diff
@@ -13,41 +13,39 @@
...
+                .setExpectedAudience(env.getProperty("jwtAudience")) // to whom the JWT is intended for
+                .setVerificationKey(jwk.getKey()) // verify the signature with the public key
+                .build(); // create the JwtConsumer instance
+
+        try {
+            // Validate the JWT and process it
+            JwtClaims jwtClaims = jwtConsumer.processToClaims(jwt);
+            return jwtClaims.getClaimsMap();
+        } catch (InvalidJwtException e) {
+            logger.error("Invalid JWT", e);
```
andlar: Does this still log the entire stack trace? If so, can you stop it from doing that? It's not a real error and just makes our logs harder to read

andlar:
Is there a way to configure something so that flags are automatically off in PROD/STG, and automatically on in DEV/Local? The UI has that set up now using two different JSON source files, so wanted to ask about that from the API side

chpl/chpl-service/src/main/java/gov/healthit/chpl/FF4JConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+
+            return ff4j;
+        } catch (NamingException | SQLException e) {
+            return null;
+        }
+    }
+
+    private DataSource ff4jDataSource() throws NamingException, SQLException {
+        Context ctx = new InitialContext();
+        DataSource ds = (DataSource) ctx.lookup("java:/comp/env/jdbc/openchpl");
```
kekey1: We have a property for this: jndiName

chpl/chpl-service/src/main/java/gov/healthit/chpl/FF4JConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+public class FF4JConfiguration {
+    private static final String TABLE_PREFIX = "ff4j_";
+
+    @Bean
+    public FF4j getFF4j() {
+        try {
+            FF4j ff4j = new FF4j();
+
+            JdbcFeatureStore featureStore = new JdbcFeatureStore(ff4jDataSource());
+            featureStore.setQueryBuilder(new JdbcQueryBuilder(TABLE_PREFIX, "", BaseDAOImpl.SCHEMA_NAME));
```
kekey1: Does this add a ton of tables? Just wondering if we might rather have the tables in their own schema (like quartz is) rather than mixed in with all of our chpl/listing data.

tmy1313: There are 5 tables that are created.  We could add them to a different schema.  There are several issues we need to figure out how we want to handle going forward.  I'll add this to the list of questions I have,

kekey1: Sounds good!

PENDING_RELEASE_NOTES.md
```diff
@@ -1,8 +1,10 @@
 
-## Version 20.11.0
+## Version 20.12.0
```
kekey1: When you get to the point of resolving the merge conflict in this file, I don't think the version number should have changed. The last release was 20.10.0.

kekey1:
In addition to my comment on the release notes - when I pulled in this branch I got a compilation error on one unit test gov.healthit.chpl.auth.jwt.JwtAuthorTest. No change to the file so no other place to leave this comment. Needs fixed before approving.

PENDING_RELEASE_NOTES.md
```diff
@@ -3,6 +3,8 @@
 _Date TBD_
 
 ### New Features
+* Respond with a 401 error when the user's token has expired
```
kekey1: This is behind a flag and should go in FLAGGED_RELEASE_NOTES.md similar to how it's done in the UI. See chat in Teams for reference.

PR #695, opened 2019-04-28
OCD-2870: Fix issue with scheduled jobs when an ACB is renamed

kekey1:


kekey1:
I am not sure... in my testing I renamed an ACB multiple times and sometimes renamed it to a name it had been previously and when I would do that a job that I previously wasn't being shown (because of looking up permissions by ACB name) would then re-appear. Maybe something like that happened during testing and made it look like it was working? If we had put in ACB IDs for job data I think that would have been easier all around.

PR #697, opened 2019-05-01
Ocd 2821- Incorrect permissions for edit/split/merge of developer/product/version

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/ProductManagerImpl.java
```diff
@@ -319,12 +319,13 @@ private ProductDTO createProduct(final ProductDTO dto)
                     + dev.getName() + " cannot be determined.";
             LOGGER.error(msg);
             throw new EntityCreationException(msg);
-        } else if (!currDevStatus.getStatus().getStatusName().equals(DeveloperStatusType.Active.toString())) {
-            String msg = "The product " + dto.getName() + " cannot be created since the developer " + dev.getName()
-                    + " has a status of " + currDevStatus.getStatus().getStatusName();
-            LOGGER.error(msg);
-            throw new EntityCreationException(msg);
         }
+        // } else if (!currDevStatus.getStatus().getStatusName().equals(DeveloperStatusType.Active.toString())) {
```
kekey1: [possibly remove this unless you think it will need to go back in?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -246,48 +246,24 @@ public DeveloperDTO update(final DeveloperDTO updatedDev, final boolean doValida
             throw new EntityCreationException(msg);
         }
 
-        // determine if the status has been changed in most cases only allowed by ROLE_ADMIN but ROLE_ACB
-        // can change it to UnderCertificationBanByOnc
-        boolean currentStatusChanged = !currDevStatus.getStatus().getStatusName()
```
kekey1: I am just wondering where all this deleted code went... when I saw it initially I was expecting it to be moved into a permissions class. Did we just not need it?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/ProductManagerImpl.java
```diff
@@ -319,12 +319,13 @@ private ProductDTO createProduct(final ProductDTO dto)
...
+        // String msg = "The product " + dto.getName() + " cannot be created since the developer " + dev.getName()
```
andlar: Should be removed

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
tmy1313: I think i got a little over zealous removing code.  I reverted the commit and re-coded the change.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -256,10 +256,6 @@ public DeveloperDTO update(final DeveloperDTO updatedDev, final boolean doValida
                 && !resourcePermissions.isUserRoleAdmin() && !resourcePermissions.isUserRoleOnc()) {
             String msg = msgUtil.getMessage("developer.statusChangeNotAllowedWithoutAdmin");
             throw new EntityCreationException(msg);
-        } else if (!currDevStatus.getStatus().getStatusName().equals(DeveloperStatusType.Active.toString())
```
kekey1: I see that you reverted but I guess I'm still confused about this bit being deleted... the comment two lines down says "if the developer is not active and not going to be active only its current status can be updated" and then the updateStatusHistory method is called. Where does that happen in the updated code? Was it not supposed to be in there?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -427,22 +414,6 @@ public DeveloperDTO merge(final List<Long> developerIdsToMerge, final DeveloperD
             throw new ValidationException(getDuplicateChplProductNumberErrorMessages(duplicateChplProdNumbers), null);
         }
 
-        // check for any non-active developers and throw an error if any are found
-        for (DeveloperDTO beforeDeveloper : beforeDevelopers) {
```
kekey1: Is there an UpdateActionPermissions for Developers already in the code base that has this check? Or was part of the task to remove this security?

tmy1313: Moved this check to the MergeActionPermissions.

PR #698, opened 2019-05-03
fix: update permissions for merge of products and versions

tmy1313:
OCD-2821

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/product/MergeActionPermissions.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/product/MergeActionPermissions.java
@@ -26,11 +26,15 @@ public class MergeActionPermissions extends ActionPermissions {
             return true;
         } else if (getResourcePermissions().isUserRoleAcbAdmin()) {
             try {
-                // Need to make sure developer for products to be merged is active
+                // Need to make sure all developers for products to be merged are active
                 List<Long> productIds = (List<Long>) obj;
-                // All products will have the same developer, so we only need to check the first one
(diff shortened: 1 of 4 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/product/MergeActionPermissions.java
```diff
@@ -1,20 +1,42 @@
...
     @Override
     public boolean hasAccess(Object obj) {
-        return false;
+        if (getResourcePermissions().isUserRoleAdmin() || getResourcePermissions().isUserRoleOnc()) {
+            return true;
+        } else if (getResourcePermissions().isUserRoleAcbAdmin()) {
+            try {
+                // Need to make sure developer for products to be merged is active
+                List<Long> productIds = (List<Long>) obj;
+                // All products will have the same developer, so we only need to check the first one
```
kekey1: I guess I'm just wondering if this comment is true... I think probably the way the UI lets you do a product merge it is true but if I just call the API with some random product ids what would happen?

andlar: Good catch Katy. Definitely see that as a possible issue

tmy1313: I'll add to check each developer for a product.

kekey1: Sounds good.. make sure to do the same thing for the versions too as I think it had a similar comment.

PR #706, opened 2019-05-14
OCD-2893: Add version split endpoint

kekey1:
Depends on OCD-2889 which depends on OCD-2885. If you want to wait until that is merged in to review this that's fine.


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ProductVersionController.java
```diff
@@ -154,4 +167,80 @@
...
+                if (!chplProductNumberUtil.isLegacy(possibleChanged.getChplProductNumber())) {
+                    if (buf.length() > 0) {
+                        buf.append(",");
+                    }
+                    buf.append(possibleChanged.getChplProductNumber());
+                }
+            }
+            responseHeaders.set("CHPL-Id-Changed", buf.toString());
+        }
+        return new ResponseEntity<SplitVersionResponse>(response, responseHeaders, HttpStatus.OK);
```
tmy1313: Is it standard for us to return data via response headers?  If we need to this list back to the calling client, should it be part of the SplitVersionResponse object?

kekey1: There are a couple of other places where if something is edited the new-style CHPL ID could change and in all those places we send that information back in this CHPL-Id-Changed header. The UI looks for that and uses it to show one of those popups that tells you one or more CHPL IDs might have changed as a result of your action and changes could take a few minutes to show up. I just did this the same way... am open to discussion about a different/better way to do that, but I guess to answer your question it's "standard" in our CHPL code although I don't know if it's "standard" in a RESTful/best practices sense.

PR #716, opened 2019-05-29
Ocd 2820 - Resolve phantom logged in issue

tmy1313:
Remove flag


PENDING_RELEASE_NOTES.md
```diff
@@ -6,6 +6,7 @@ _Date TBD_
 * Add FF4j framework (feature flags)
 * Add endpoints to provide ability for users to save filters for admin reports
 * Updated product and version activity metadata to parse activity for merges and splits; better parsing of product and developer names if they have been deleted.
+* Respond with a 401 error when the user's token has expired
```
andlar: This line looks a lot like the Bug Fixed item from line 13. I don't think it should be in two places, right?

PR #718, opened 2019-05-31
Add FF4j endpoint

tmy1313:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/FeatureFlagController.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+
+@Api(value = "feature-flags")
+@RestController
+@RequestMapping("/feature_flags")
```
kekey1: I thought we were trying to have all URLs use dashes for consistency instead of underscores? I know that some older ones probably don't use dashes but I thought we were heading that way with new APIs. Also, do you want this to show up in the swagger doc? I'm thinking no, and that's what would happen now, but just checking in case we actually do - you'd need to edit another file to make it show up.

andlar: Oh, I forgot about that dash/underscore! I'd noticed that too but forgot to mention it

I'm also fine not having these show up in swagger

tmy1313: Can't say that I remember that decision.  I looked through all of our controller urls and only one endpoint uses `-` (/system-status).   I think saying that all new APIs have underscores creates more inconsistency since there is no plan to refactor all of the existing urls.

PR #719, opened 2019-06-05
OCD-2916: Basic surveillance report

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SurveillanceReportController.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        //create the excel workbook object
+        Workbook reportXlsx = reportManager.exportQuarterlyReport(quarterlyReportId);
+
+        //create a new file in the download directory
+        QuarterlyReportDTO report = reportManager.getQuarterlyReport(quarterlyReportId);
+        String filename = report.getQuarter().getName() + "-" + report.getAnnualReport().getYear()
+                    + "-" + report.getAnnualReport().getAcb().getName() + "-quarterly-report.xlsx";
+        File downloadFile = fileUtils.createDownloadFile(filename);
```
andlar: Does this mean the download folder will get new files in them permanently on each request for a quarterly surveillance report? That doesn't seem like a good idea. Can you create the Excel file as a temporary file and stream that instead? I remember doing something where we created a temporary file for some other thing in CHPL, though I have no idea where it is

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -402,6 +402,14 @@ surveillance.errorAdding=Error adding a new pending surveillance. Please make su
 surveillance.permissionError=User %s does not have access to add surveillance.
 surveillance.permissionErrorWithAcb=User %s does not have access to add surveillance to %s as the listing is owned by ONC-ACB %s.
 
+# surveillance reports
+report.quarterlySurveillance.missingYear=A year must be specified for the quarterly report.
+report.quarterlySurveillance.missingAcb=An ACB must be specified for the quarterly report.
```
andlar:
```suggestion
report.quarterlySurveillance.missingAcb=An ONC-ACB must be specified for the quarterly report.
```

Public facing text should always have the "ONC-" part in it

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/surveillance/report/package-info.java
```diff
@@ -0,0 +1,8 @@
+/**
+ *
```
andlar: There are a couple of these package-info files with duplicate /* */ blocks. Didn't mention it on the first one, but with more than one I felt the need to call it out

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SurveillanceReportController.java
```diff
@@ -0,0 +1,220 @@
...
+                    createRequest.setAcb(acb);
+                }
+            }
+        }
+
+        if (createRequest.getAcb() == null || createRequest.getAcb().getId() == null) {
+            throw new InvalidArgumentsException(msgUtil.getMessage("report.quarterlySurveillance.missingAcb"));
+        }
+
+        //create the report
```
tmy1313: I've always thought of controllers as being "dumb".  Do you think some of this logic and object building belongs in the Manager?

kekey1: I agree with you and am not sure how to proceed here. This bit of code is part of converting the request object into the DTO object that gets sent into the manager. If an ACB is provided it doesn't do anything, but if one isn't it's checking to see what ACBs the user has access to and if it's only one then it creates the report under that one. If the user has multiple ACBs then it throws an error. I was thinking of what I'm doing here as validation on the request object that's sent in. Do you think the controller is an appropriate place for a check like that or it really should all go into the manager? The manager just works with whatever ACB is passed in with the DTO object.

kekey1: @tmy1313 I've updated this as we discussed.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
```diff
@@ -822,6 +822,17 @@ private void validateOrderBy(final String orderBy) throws InvalidArgumentsExcept
         return dimensionalDataManager.getJobTypes();
     }
 
+    @Secured({
+        Authority.ROLE_ADMIN, Authority.ROLE_ACB
```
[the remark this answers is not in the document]
kekey1: Updated!

PR #721, opened 2019-06-06
Ocd 2912 - Create a basic complaint

tmy1313:


PENDING_RELEASE_NOTES.md
```diff
@@ -4,5 +4,6 @@ _Date TBD_
 
 ### New Features
 * Add endpoints to provide activity metadata for users, announcements, pending listings, corrective action plans, and pending surveillances
+* Add endpoints for managing surveillance endpoints: get all complaints, delete complaint, create complaint, update complaint
```
kekey1:
```suggestion
* Add endpoints for managing surveillance complaints: get all complaints, delete complaint, create complaint, update complaint
```

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ComplaintController.java
```diff
@@ -0,0 +1,102 @@
...
+import gov.healthit.chpl.exception.EntityRetrievalException;
+import gov.healthit.chpl.exception.ValidationException;
+import gov.healthit.chpl.manager.ComplaintManager;
+import gov.healthit.chpl.web.controller.results.ComplaintResults;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+
+@Api(value = "complaints")
+@RestController
+@RequestMapping("/complaints")
```
kekey1: Add to SwaggerConfig so these endpoints show up in the documentation

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ComplaintController.java
```diff
@@ -0,0 +1,102 @@
...
+    
+    @ApiOperation(value = "Save complaint for use in Surveillance Quarterly Report.",
+            notes = "")
+    @RequestMapping(value = "", method = RequestMethod.POST, produces = "application/json; charset=utf-8")
+    public @ResponseBody ComplaintDTO create(@RequestBody final ComplaintDTO complaint) throws EntityRetrievalException, ValidationException {
+        if (ff4j.check(FeatureList.COMPLAINTS)) {
+            ValidationException error = new ValidationException();
+            //Make sure there is an ACB
+            if (complaint.getCertificationBody() == null) {
+                error.getErrorMessages().add("Certification Body is required.");
```
kekey1: Add to errors.properties

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ComplaintController.java
```diff
@@ -0,0 +1,102 @@
...
+            return results;
+        } else {
+            throw new NotImplementedException();
+        }
+    }
+    
+    @ApiOperation(value = "Save complaint for use in Surveillance Quarterly Report.",
+            notes = "")
+    @RequestMapping(value = "", method = RequestMethod.POST, produces = "application/json; charset=utf-8")
+    public @ResponseBody ComplaintDTO create(@RequestBody final ComplaintDTO complaint) throws EntityRetrievalException, ValidationException {
```
kekey1: Normally we pass domain objects back and forth in the controllers rather than the DTOs. I know it's a lot of layers of classes and annoying if they DTO matches the domain object pretty closely. What do you think about changing this to use a domain object for consistency? Or moving forward do you think controllers should use DTOs? I like having domain objects because we could package up the domain package as its own jar with objects that one could use to communicate with the CHPL.

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
```diff
@@ -2,4 +2,6 @@
 
 public class FeatureList {
     public static final String BETTER_SPLIT = "better-split";
+    public static final String OCD_2820 = "ocd2820";
```
kekey1: Needs integration with upstream.. better split isn't there anymore

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/ComplaintDAOImpl.java
```diff
@@ -0,0 +1,258 @@
...
+        ComplaintEntity entity = null;
+
+        Query query = entityManager.createQuery("FROM ComplaintEntity c " + "JOIN FETCH c.certificationBody "
+                + "LEFT JOIN FETCH c.complaintType " + "LEFT JOIN FETCH c.complaintStatusType "
+                + "WHERE c.deleted = false " + "AND c.id = :complaintId", ComplaintEntity.class);
+        query.setParameter("complaintId", id);
+        List<ComplaintEntity> result = query.getResultList();
+
+        if (result.size() > 1) {
+            throw new EntityRetrievalException("Data error. Duplicate filter id in database.");
```
kekey1: wrong error text

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/ComplaintEntity.java
```diff
@@ -0,0 +1,225 @@
...
+    private Long id;
+
+    @Basic(optional = true)
+    @OneToOne(optional = true, fetch = FetchType.LAZY)
+    @JoinColumn(name = "certification_body_id", nullable = false, insertable = true, updatable = false)
+    private CertificationBodyEntity certificationBody;
+
+    @Basic(optional = true)
+    @OneToOne(optional = true, fetch = FetchType.LAZY)
+    @JoinColumn(name = "complaint_type_id", nullable = true, insertable = true, updatable = true)
```
kekey1: If this works, I guess it's fine, but normally in the entities when we are joining in another entity like this we have a @Column defined for the bigint foreign key value (complaint_type_id) and that column is allowed to be inserted and updated but this column that's going to hold the actual entity that gets joined in is not insertable/updatable. I think if you look at any of the Surveillance entities you will see a good example. I would be afraid this syntax would let the underlying ComplaintTypeEntity/ComplaintStatusTypeEntity/CertificationBodyEntity get updated.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/ComplaintDAOImpl.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    private List<ComplaintStatusTypeEntity> getComplaintStatusTypeEntities() {
+        Query query = entityManager.createQuery("from ComplaintStatusTypeEntity where (NOT deleted = true) ",
+                ComplaintStatusTypeEntity.class);
+        List<ComplaintStatusTypeEntity> result = query.getResultList();
+        return result;
+    }
+
+    private CertificationBodyEntity getAcbEntityById(final Long entityId) throws EntityRetrievalException {
```
kekey1: If you change the entity as mentioned, you would only need the ACB ID and wouldn't need this method.

PENDING_RELEASE_NOTES.md
andlar: Should be in FLAGGED, right?

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -494,3 +494,12 @@ filter.notPresent=No filter object was present to be saved.
 filter.empty=Filter cannot be empty.
 filter.noUser=The filter does not have an associated user.
 filter.noFilterType=The filter does not have an associated filter type.
+
+complaints.create.statusOpen=When creating a new complaint, the Complaint Status must be open.
```
andlar: Is this accurate?

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -494,3 +494,12 @@ filter.notPresent=No filter object was present to be saved.
...
+complaints.create.statusOpen.error=There was an error in validator: ComplaintStatusOpen
```
andlar: These error messages are supposed to be user friendly to non-developers. These messages can (maybe should?) go through the product owner for verification too. At least, that was one of the initial reasons for this file at all

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
andlar: What's 2820? I don't think this should be here

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/ComplaintEntity.java
tmy1313: I'll make this change, and I have seen it throughout the code.  But it feel like it's hack.  I have never heard or seen this done other places or in code examples on the web.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/ComplaintDAOImpl.java
tmy1313: Based on the comment above, I'll remove this method.

chpl/chpl-resources/src/main/resources/errors.properties
tmy1313: image

kekey1: Should the status be not null in the data model then?

PR #727, opened 2019-06-18
Ocd 2913 - Associate a listing to a complaint

tmy1313:


chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -522,3 +522,5 @@ complaints.complaintStatusType.notExists=The Complaint Status Type does not exis
 complaints.receivedDate.required=Received Date is required.
 complaints.acbComplaintId.required=ONC-ACB Complaint Id is required.
 complaints.summary.required=Summary is required.
+complaints.listingId.notFound=Unexpected error looking up certified product with id '%s'.
+complaints.listingId.doesNotMatchAcb=Certified product %s does not have the same certification body as the complaint.
```
kekey1: Should this read "ONC-ACB" instead of "certification body"? I think that's how we refer to them in other error messages.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/ComplaintDAO.java
```diff
@@ -11,7 +11,7 @@
 
     List<ComplaintStatusTypeDTO> getComplaintStatusTypes();
 
-    List<ComplaintDTO> getAllComplaints();
+    List<ComplaintDTO> getAllComplaints() throws EntityRetrievalException;
```
kekey1: Just wondering how/why this might throw this particular exception. If it's valid that's fine, I'm just thinking that I personally have mostly used it for like looking something up by ID but that ID doesn't exist in the database.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/ComplaintListingMap.java
```diff
@@ -0,0 +1,91 @@
...
+
+import org.springframework.beans.BeanUtils;
+
+import gov.healthit.chpl.dto.ComplaintListingMapDTO;
+
+public class ComplaintListingMap {
+    private Long id;
+    private Long complaintId;
+    private Long listingId;
+    private Date creationDate;
```
kekey1: Do we need creation date, last modified date, last modified user, and deleted fields in the domain object? I'm not sure others have them... maybe some but not all of those fields? If they need to be shown on the UI that's fine they are here of course.

PR #732, opened 2019-06-21
Ocd 2912 - Create a basic complaint

tmy1313:
rename complaint type -> complainant type


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
```diff
@@ -1226,12 +1226,12 @@ private FuzzyChoices updateFuzzyChoices(final FuzzyChoices fuzzyChoices)
         return result;
     }
     
-    @ApiOperation(value = "Get all possible complaint types in the CHPL")
-    @RequestMapping(value = "/data/complaint_types", method = RequestMethod.GET,
+    @ApiOperation(value = "Get all possible complainant types in the CHPL")
+    @RequestMapping(value = "/data/complainant_types", method = RequestMethod.GET,
```
andlar: Didn't notice this before, but could we use dashes instead of underscores? I think we'd talked about this, and I know there are other underscores in the API, but I'd like to move towards dashes everywhere we can

PR #734, opened 2019-06-26
OCD-2912 - Create a basic complaint

tmy1313:
OCD-2912


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/complaints/ComplainantTypeValidation.java
```diff
@@ -41,7 +41,8 @@ public boolean isValid(ComplaintValidationContext context) {
 
             // Complainant type other must exist if complaint type = other
             if (isComplainantTypeSetToOther(context.getComplaintDTO().getComplainantType())
-                    && context.getComplaintDTO().getComplainantTypeOther().trim() == "") {
+                    && (context.getComplaintDTO().getComplainantTypeOther() == null
+                            || context.getComplaintDTO().getComplainantTypeOther().trim() == "")) {
```
kekey1: Does the StringUtils.isEmpty function work here? I usually use that and expect it to handle both the null and empty cases (you'd still need to trim it I think... which brings up another issue I've been thinking about where we probably need to trim all input from the user but don't do it consistently).

PR #737, opened 2019-07-01
OCD-2915 - Associate a complaint with surveillance

tmy1313:
OCD-2915

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SurveillanceLite.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SurveillanceLite.java
@@ -1,0 +1,122 @@
+package gov.healthit.chpl.domain;
+
+import java.io.Serializable;
+import java.util.Date;
+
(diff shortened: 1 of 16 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/SurveillanceLite.java
```diff
@@ -0,0 +1,122 @@
...
+    private Long certifiedProductId;
+    private Date startDate;
+    private Date endDate;
+    private Long surveillanceTypeId;
+    private SurveillanceType surveillanceType;
+    private Integer numRandomizedSites;
+    private Long userPermissionId;
+    private String chplProductNumber;
+
+    public SurveillanceLite() {
```
andlar: We have a "CertifiedProductBasic" thing, I think, that's the Listing without the criteria/cqms/surveillance (IIRC). Would it make sense to keep a similar naming convention for a "SurveillanceBasic" object?

PR #740, opened 2019-07-03
OCD-2921: Add complaints to surveillance report

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/surveillance/SurveillanceProcessType.java
```diff
@@ -0,0 +1,65 @@
+package gov.healthit.chpl.domain.surveillance;
```
tmy1313: Are we actually including this data in the XML download?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/surveillance/SurveillanceOutcome.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.Serializable;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlType;
+
+import org.apache.commons.lang3.StringUtils;
+
+@XmlType(namespace = "http://chpl.healthit.gov/listings")
```
tmy1313: Are outputting this information to the download XML?  If not, I'm not sure we need the XML annotations.  Actually, I don't think we are using any of the annotations anymore, since the XML creation was re-written...

kekey1: I moved the two classes into a privileged package and removed the XML annotations. I actually am not using these classes yet and jumped the gun when adding them (changed so many files I forgot that I had added these) BUT we're going to need them in the next ticket so I was thinking I'd just leave them for now. If you're against that b/c they aren't really needed for this feature then I can delete them.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/surveillance/SurveillanceProcessType.java
kekey1: No! good catch! Removed annotations.

PR #745, opened 2019-07-16
OCD-2754 - Improve Reports responsiveness - Announcements

tmy1313:
QA Issues

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/activity/GetAnnouncementMetadataActionPermissions.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/activity/GetAnnouncementMetadataActionPermissions.java
@@ -22,7 +22,7 @@ public class GetAnnouncementMetadataActionPermissions extends ActionPermissions 
 
     @Override
     public boolean hasAccess() {
-        return getResourcePermissions().isUserRoleAdmin() || getResourcePermissions().isUserRoleOnc();
+        return false;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/activity/GetAnnouncementMetadataActionPermissions.java
```diff
@@ -22,7 +22,7 @@ public GetAnnouncementMetadataActionPermissions(final AnnouncementDAO announceme
 
     @Override
     public boolean hasAccess() {
-        return getResourcePermissions().isUserRoleAdmin() || getResourcePermissions().isUserRoleOnc();
+        return false;
```
andlar: Why is this changing? ADMIN and ONC can still see everything, while others can't, right? My understanding is that this method is used when checking for access "generically", so ADMIN & ONC could still get access through this one, right? Or do I not understand this correctly?

tmy1313: I thought we discussed that based on the ticket description, that anyone can get announcement activity via the API.  ANON will only see the Public announcements (via the PostFilter).  Then on the front end, if a logged in user navigates to the Announcements Report via the URL bar, we would show whatever data they were allowed to see, except for the ANON user, who would get the Log In page.

andlar: So before this returned true for ADMIN and ONC. Now it returns false for everyone. What functionality are ADMIN and ONC losing by changing this to return false for everyone?

kekey1: I think that this method isn't called anymore in favor of the method that's below this (but unchanged so doesn't show in the PR) in this file. If someone later wants to check access for announcement activity without passing in the activity object itself this would be the method to update but right now I don't think it's being used.

PR #753, opened 2019-07-30
feat: Add privileged surveillance data

kekey1:
* Add privileged surveillance data to the quarterly report+surveillance mapping
* Add feature to include copying previous quarterly report data when initiating a new report
* Add /data/ endpoints to return lookup values for surveillance outcomes and surveillance process types
* Add calculated surveillance summary counts to the Excel report

OCD-2920


chpl/chpl-service/src/main/java/gov/healthit/chpl/builder/ComplaintsWorksheetBuilder.java
```diff
@@ -54,7 +59,7 @@
     private static final int COL_CRITERIA_ID = 8;
     private static final int COL_CHPL_ID = 9;
     private static final int COL_SURV_ID = 10;
-    //columns 11 - 14 get hidden
+    //columns L-O get hidden
```
andlar: Would it make sense to go back to 11 - 14, set them as constants, and use them in the for loop a few sections down? Or maybe have an array of "hidden" booleans that we could use to flip column hiding stuff easily as needed? I'm thinking about that ONC review column thing...

PR #761, opened 2019-08-07
feat: remove address re-use between organizations

tmy1313:
OCD-3018


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/CertificationBodyDAOImpl.java
```diff
@@ -79,7 +76,7 @@ public CertificationBodyDTO update(final CertificationBodyDTO dto) throws Entity
 
         if (dto.getAddress() != null) {
             try {
-                entity.setAddress(addressDao.mergeAddress(dto.getAddress()));
+                entity.setAddress(addressDao.create(dto.getAddress()));
```
kekey1: This looks like the update method is always calling code to create a new address. I feel like it should call the saveAddress method you have in the developer dao?

PR #763, opened 2019-08-08
flag: support invitation of ROLE_DEVELOPER

andlar:
This is much of a work in progress PR, but does function, and provides the base for a few other tickets

[#OCD-2158]

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/UserDeveloperMapDAOImpl.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/UserDeveloperMapDAOImpl.java
@@ -64,7 +64,7 @@ public class UserDeveloperMapDAOImpl extends BaseDAOImpl implements UserDevelope
                         + "join fetch u.permission perm "
                         + "join fetch u.contact contact "
                         + "where (udm.deleted != true) "
-                        + "AND (cb.id = :developerId) ",
+                        + "AND (developer.id = :developerId) ",
                 UserDeveloperMapEntity.class);
         query.setParameter("developerId", developerId);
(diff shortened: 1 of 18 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/UserDeveloperMapDAOImpl.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public List<UserDeveloperMapDTO> getByDeveloperId(final Long developerId) {
+        Query query = entityManager.createQuery(
+                "from UserDeveloperMapEntity udm "
+                        + "join fetch udm.developer developer "
+                        + "join fetch udm.user u "
+                        + "join fetch u.permission perm "
+                        + "join fetch u.contact contact "
+                        + "where (udm.deleted != true) "
+                        + "AND (cb.id = :developerId) ",
```
kekey1: It says "cb.id" but I don't think "cb" is anything in this query. So I think either change the HQL or if the code is not being called maybe we don't need this method?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/UserDeveloperMapDTO.java
```diff
@@ -0,0 +1,53 @@
+package gov.healthit.chpl.dto;
+
+import java.io.Serializable;
+
+import gov.healthit.chpl.dto.auth.UserDTO;
+import gov.healthit.chpl.entity.UserDeveloperMapEntity;
+
+public class UserDeveloperMapDTO implements Serializable {
+    private static final long serialVersionUID = 1316069925338614100L;
```
kekey1: Can you change this value so it's something unique to this class?

PR #770, opened 2019-08-14
fix: Do not add any users to newly created acb/atl

kekey1:
OCD-3019


kekey1:
> Any unit tests that could have caught this?
There are no unit tests that I can find that test ACB creation. The unit tests that we have for ATL creation are mocked, so I would have to mock the method that returns users for the newly created ATL and obviously I would mock it to return 0. That's a confusion for me about how mocking works... I don't think it would be able to catch this issue. Do you want me to try to add some tests for this PR?

> Also, anything in the ROLE_DEVELOPER / Developer creation that might be similar?
No, there is not. There used to be code like "Add a Spring ACL for this new ACB/ATL" in the managers because the spring security didn't work otherwise. That code simply got migrated to "Add the user to our custom permissions tables for this new ACB/ATL". I just removed it because we didn't need it anymore. Developers never had that first part though with the spring tables because we never had permissions on them until now.

andlar:
I don't see a need for unit tests here, given your description of how it's currently running, so have nothing further to add. Continued approval :)

PR #775, opened 2019-08-21
feat: Allow retrieval of deleted API keys when retrieving all keys

tmy1313:
OCD-2748


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/ApiKeyDAOImpl.java
```diff
@@ -227,10 +227,12 @@ private void update(final ApiKeyEntity entity) {
 
     }
 
-    private List<ApiKeyEntity> getAllEntities() {
+    private List<ApiKeyEntity> getAllEntities(final Boolean includeDeleted) {
 
         List<ApiKeyEntity> result = entityManager
-                .createQuery("from ApiKeyEntity where (NOT deleted = true) ", ApiKeyEntity.class).getResultList();
+                .createQuery("from ApiKeyEntity where (deleted = :includeDeleted) ", ApiKeyEntity.class)
```
kekey1: I don't think this is quite right.. this would let you search for only deleted api keys if includeDeleted == true. I think what you want is some dynamic HQL that either includes "where deleted = false" or leaves off the where clause entirely.

PR #784, opened 2019-08-30
feat: Give errors on confirm if missing developer data

drbgfc:
[#OCD-3002]


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
```diff
@@ -11,6 +11,7 @@
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
```
andlar: I don't think this import is used

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
```diff
@@ -861,6 +871,32 @@ public void streamSEDDetailsDocumentContents(final HttpServletResponse response)
                 throw new ValidationException(pcpDto.getErrorMessages(), pcpDto.getWarningMessages());
             }
 
+            if (pendingCp.getDeveloper() != null && pendingCp.getCertifyingBody() != null
```
andlar: I'm pretty sure the CertifyingBody of a pendingCp can never be null, and if it was I don't think we'd even get this far. That said, I'm not sure it's actually wrong to perform the check. Up for discussion

The second part here (and related to the next line) is related to what happens if a Listing is being uploaded for a new Developer. In that case, there won't be a Developer in the system to check against, and can't be until the Listing is confirmed. I think this part might warrant a group discussion just so we're all on the same page as to what that means

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/developer/DeveloperInSystemIsSavedValidator.java
```diff
@@ -0,0 +1,114 @@
...
+import org.springframework.util.StringUtils;
+
+import gov.healthit.chpl.dto.DeveloperACBMapDTO;
+import gov.healthit.chpl.dto.DeveloperDTO;
+import gov.healthit.chpl.util.ErrorMessageUtil;
+
+/**
+ * Validate fields used when creating a new developer.
+ */
+@Component("developerInSystemIsSavedValidator")
```
andlar: This is the part I think I'd like to talk about. There's already a Validator for Developer creation, and one for Developer Update. I know the Update Validator can't require all data, due to legacy data, but the Create one can. The Create one is also the one that's used when a new Developer is created, which only happens on Listing upload, so maybe that Create one should be expanded / tweaked to basically be something like "DeveloperDuringUploadValidator". Maybe the "DeveloperCreateValidator" should actually go away, as we can't POST a new Developer at all anyway. There's no way for a Developer to exist in the CHPL without having a Product, that Product having a Version, and that Version having a Listing, so creating just a Developer on its own is impossible

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
drbgfc: That's correct. I should have removed it (and will update accordingly). Normally in the past I go ahead and run 'organize imports' automatically or as part of a team-based save-cleanup automation. I wasn't sure if I should run it here since others might not be doing that and it might cause the code to look very different. If we all did it, or it was automated on save for us all, we wouldn't ever have such issues or concerns.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
drbgfc: On the first part, I'd say unless we are 100% sure then the check should be made. But, I'm not dead set against removing the checks at all. You know the software better than me. My perspective is simply that I didn't know for sure so I checked.

As for what happens if we create a new developer - I would like to discuss that as well.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
andlar: I tend to do the Eclipse `Ctrl-Shift-O` command as I'm getting ready for PR. I _think_ that's the "organize imports" command ;)

If you can figure out how to automate that on save I'd be all about incorporating that into our processes. For that matter, any other automated way of cleaning up our code is welcomed and encouraged (with the caveat about too many "changed lines" makes it hard to parse a PR). That said, organizing the imports shouldn't be problematic

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/developer/DeveloperInSystemIsSavedValidator.java
drbgfc: Thanks for the review. A discussion on this would be great. Some good points. My thoughts since it's on my mind:
I designed DeveloperInSystemIsSavedValidator to only care about if the data in the system exists (with the exception of attestation), not that it is valid so much. DeveloperCreationValidator handles whether it's valid or not (well, a subset of it anyway). That's how I am able to use it that way, because I know if the data is there, it must be valid. So they have separate concerns. The classes are simple in that they just have one validate method so extending and overriding doesn't really give any code-reuse. So I separated. However, as for having one merged version of the two classes (e.g. DeveloperDuringUploadValidator) handle it all, that could certainly work, and would require less code and may be less confusing/less error-prone. So that seems like a good idea. Although, it would cause the missing info use case to check more than is needed. Again, we would know that stored data is valid as it's been through the process before. The performance though is negligible. and not an argument. I see your point. 

The main reasons that I went with my own class were: 
1: The requirements were different than what I had to support. For example, DeveloperCreationValidator does not require a website. It doesn't require a transparency attestation or validate it in any way. So this is by far the number 1 reason because I didn't think I was allowed to enforce things that were not already enforced. But it appears I am in this case? If so, merging into one class is likely the best solution.
2: I wanted to make it obvious through my own message properties that the error was related to bad data that is already in the system vs trying to store brand new bad data. Of course, for point 2, I could find another way to implement such a distinction but would need to avoid the possible pitfall of it being messy since confined to one class. Or, I could just not provide the distinction at all as the importance of separate properties (excepting the literal extra ones) is probably quite low.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
drbgfc: Yeah that's the hotkey. I would suggest that the team add organize imports (it's a checkbox) to the save action in Preferences->Editor->Save Actions. Another great one to add for this project's requirements specifically is 'Remove trailing white spaces on all lines.' For that one you would want to go in and edit out everything there by default except that. And in the future maybe there are other items we can automate that are universally understood as a team requirement so they won't cause mass changes. I am certainly all about the computer spending time caring about blank spaces vs me.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/DeveloperManager.java
```diff
@@ -17,6 +18,10 @@
 import gov.healthit.chpl.exception.ValidationException;
 
 public interface DeveloperManager {
+    String NEW_DEVELOPER_CODE = "XXXX";
+    String SYSTEM_DEV_VALIDATION_ERROR = "Unable to validate system developer as the pending ACB Name is null or its String representation "
+            + "is null or empty";
+
```
tmy1313: This error message should probably be in the errors.properties.

drbgfc: Thanks, good point, updated accordingly

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -78,10 +78,15 @@ developer.contact.emailRequired=Developer contact email address is required.
...
 developer.contact.nameRequired=Developer contact name is required.
 developer.websiteRequired=Developer website is required.
-developer.websiteIsInvalid=Developer's website is improperly formed.
+developer.websiteIsInvalid=Developer website is improperly formed.
 developer.status.noCurrent=The developer must have a current status specified.
 developer.status.duplicateStatus=The status '%s' cannot be listed twice in a row.
 
+#system-specific developer errors
+system.developer.transparencyAttestationIsNullOrEmpty=The developer transparency attestation in the system is null or empty.
+system.developer.transparencyAttestationNotMatching=A transparency attestation does not exist in the system where the ACB name matches the pending certifying body name.
```
andlar: All public facing mentions of ACBs need to refer to them as the full "ONC-ACB". Similar with "ONC-ATL"

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -775,4 +779,85 @@ public String toString() {
...
+                LOGGER.warn("Skipping system validation due to null pending developer or a null system developer");
+            }
+        } else {
+            LOGGER.info("Skipping system validation due to new developer code '" + NEW_DEVELOPER_CODE + "'");
+        }
+    }
+
+    private boolean isNewDeveloperCode(final String chplProductNumber) {
+        String devCode = chplProductNumberUtil.getDeveloperCode(chplProductNumber);
+        return !StringUtils.isEmpty(devCode) && devCode.contentEquals(NEW_DEVELOPER_CODE);
```
andlar: I think we usually use something like `StringUtils.equals(devCode, NEW_DEVELOPER_CODE)`, which handles the case where devCode might be null, in places like this

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
```diff
@@ -109,12 +110,12 @@
     @Autowired
     private CertifiedProductManager cpManager;
 
-    @Autowired 
+    @Autowired
```
Plow74: Could we start moving  our field injections to constructor injections so we are following Spring's preferred method?

drbgfc: I can update the existing clas sure

PR #787, opened 2019-09-05
qa: Not allow the last logged in date to be wiped out on edit of user

tmy1313:
OCD-3013


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -370,6 +366,8 @@ public User updateUserDetails(@RequestBody final User userInfo)
         toUpdate.setSignatureDate(before.getSignatureDate());
         toUpdate.setSubjectName(before.getSubjectName());
         toUpdate.setTitle(userInfo.getTitle());
+        //Client should be able to change this value, so we'll always use the on from the DB
```
andlar:
```suggestion
        //Client should not be able to change this value, so we'll always use the one from the DB
```

PR #793, opened 2019-09-25
qa: Fix issues found during testing

tmy1313:
OCD-2159

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/validation/DeveloperActiveValidation.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/validation/DeveloperActiveValidation.java
@@ -24,7 +24,6 @@ public class DeveloperActiveValidation extends ValidationRule<ChangeRequestValid
                 return true;
             }
         } else {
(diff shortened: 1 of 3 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/validation/DeveloperActiveValidation.java
```diff
@@ -2,21 +2,37 @@
...
+        ChangeRequest crToTest = null;
+        if (context.getChangeRequest() != null && context.getChangeRequest().getId() != null) {
+            try {
+                crToTest = context.getChangeRequestDAO().get(context.getChangeRequest().getId());
+            } catch (EntityRetrievalException e) {
+                // This should be caught be ChangeRequestExistenceValidation
+                return true;
+            }
+        } else {
+            getMessages().add(getErrorMessage("changeRequest.developer.notActive"));
```
kekey1: Is this the right error message for when context.getChangeRequest is null? The "if" part doesn't seem to have to do with the status of the developer, but I could be missing something!

PR #796, opened 2019-09-30
flag: Add ability for ROLE_ACB to manage developer change requests

tmy1313:
OCD-3093


chpl/chpl-resources/src/main/resources-dev/environment.properties
```diff
@@ -146,4 +146,10 @@ chplUrlBegin=https://chpl.ahrqdev.org
...
 asyncListingDetailsEnabled=false
-asyncCacheRefreshEnabled=true
\ No newline at end of file
+asyncCacheRefreshEnabled=true
+
+#Website Change Request Email
+changeRequest.website.approval.subject=Website URL change request approved
+changeRequest.website.approval.body=Hello,<br/><br/>The website change request that was submitted on %s to change the website URL from %s to %s has been approved by %s.  The updated website URL will now be reflected in CHPL.<br/><br/>Thanks,<br/>The CHPL Team 
+changeRequest.website.pendingDeveloperAction.subject=Website URL change request requires further action
+changeRequest.website.pendingDeveloperAction.body=Hello,<br/><br/>The website change request that was submitted on %s to change the website URL from %s to %s requires further action.  %s has sent this change request back you for updating with the following comment:<br><ul><li>%s</li></ul><br/><br/>Thanks,<br/>The CHPL Team
```
andlar: The "from %s to %s" part gives me pause. I noticed that the "from" value is being pulled from the Developer's current data, which means that if someone were to submit a CR, and a day or two later that Developer's website was updated some other way (upload, manual edit, whatever), then the email would have content that suggested the change was from the new data to the requested data, not the old data. I know it's an edge case, but if the text was just changed to read "change the website URL to %s" and just drop the old data then it wouldn't be an issue

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/domain/service/ChangeRequestWebsiteService.java
```diff
@@ -0,0 +1,207 @@
...
+@Component
+public class ChangeRequestWebsiteService implements ChangeRequestDetailsService<ChangeRequestWebsite> {
+
+    private ChangeRequestDAO crDAO;
+    private ChangeRequestWebsiteDAO crWebsiteDAO;
+    private DeveloperDAO developerDAO;
+    private DeveloperManager developerManager;
+    private UserDeveloperMapDAO userDeveloperMapDAO;
+    private ActivityManager activityManager;
+    private Environment env;
```
Plow74: Consider removing the Environment autowire and replacing env.getProperty(s) with @Value like the other fields.

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/dao/DeveloperCertificationBodyMapDAOImpl.java
```diff
@@ -0,0 +1,47 @@
...
+                .createQuery(hql, DeveloperCertificationBodyMapEntity.class)
+                .setParameter("developerId", developerId)
+                .getResultList().stream()
+                .map(item -> new CertificationBody(item.getCertificationBody()))
+                .collect(Collectors.<CertificationBody> toList());
+    }
+
+    @Override
+    public List<Developer> getDevelopersForCertificationBody(final Long certificationBodyId) {
+        String hql = "FROM DeveloperCertificationBodyMapEntity "
```
kekey1: I was expecting to see "FROM DeveloperCertificationBodyMapEntity main " like you have in the method above. (ie. I think "main" is missing)

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/domain/service/ChangeRequestStatusService.java
```diff
@@ -0,0 +1,176 @@
...
+    private boolean isStatusChangeValid(final Long previousStatusTypeId, final Long newStatusTypeId) {
+        // Does this status type id exist?
+        try {
+            crStatusTypeDAO.getChangeRequestStatusTypeById(newStatusTypeId);
+        } catch (EntityRetrievalException e) {
+            return false;
+        }
+
+        // Cannot change status is Cancelled
+        if (previousStatusTypeId == this.cancelledByRequesterStatus) {
```
kekey1: I think with "L"ong (as opposed to "l"ong) you should use .equals instead of ==. Not 100% on that but I feel like I've been burned before by using ==.

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/manager/ChangeRequestManagerImpl.java
```diff
@@ -121,26 +123,15 @@ public ChangeRequest createChangeRequest(final ChangeRequest cr)
...
-        cr = changeRequestDAO.get(changeRequestId);
-        return populateChangeRequestData(cr);
+        return changeRequestDAO.get(changeRequestId);
     }
 
     @Override
     @Transactional(readOnly = true)
-    @PreAuthorize("@permissions.hasAccess(T(gov.healthit.chpl.permissions.Permissions).CHANGE_REQUEST, "
-            + "T(gov.healthit.chpl.permissions.domains.ChangeRequestDomainPermissions).GET_ALL)")
+    @PostFilter("@permissions.hasAccess(T(gov.healthit.chpl.permissions.Permissions).CHANGE_REQUEST, "
```
kekey1: Should we still have the PreAuthorize here as well so that users who would never be allowed to see change requests get an Access Denied error rather than an empty list (which is what I think the postfilter will do)?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/impl/DeveloperDAOImpl.java
```diff
@@ -754,4 +762,18 @@ private DeveloperStatusEntity getStatusByName(final String statusName) {
...
         return statuses.get(0);
     }
+
+    private List<DeveloperEntity> getEntitiesByCertificationBodyId(final List<Long> certificationBodyIds) {
+        String hql = "SELECT DISTINCT dev "
+                + "FROM CertifiedProductEntity cp "
+                + "JOIN FETCH cp.productVersion pv "
+                + "JOIN FETCH pv.product prod "
+                + "JOIN FETCH prod.developer dev "
+                + "WHERE cp.certificationBody.id IN :certificationBodyIds";
```
kekey1: I think I usually put parenthesis around the parameters in the "IN" clause, like "IN (:certificationBodyIds)" but I'm not sure if they are actually necessary or not. It seems like something hibernate might handle for you if they are left out. If you know this HQL runs then don't bother, but just wanted to note it in case.

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/domain/service/ChangeRequestStatusService.java
andlar: I remember checkstyle reporting me as doing a bad thing by not using the `.equals()` in that case too

PR #798, opened 2019-10-03
fix: API response for basic certified product info removes non-basic details from response

Plow74:
OCD-3101


Plow74:
Hmm, I didn't do a usage report for this class in my IDE, assuming it was just used in the controller. I see now that it seems the CertifiedProductSearchDetails object is used a LOT of places. The lists that I'm explicitly nulling-out are instantiated with empty lists when the object is created. Also, this annotation is only effects CertifiedProductSearchDetails null field members, meaning that a field member that's an object, the Developer object for example, would still show 'null' values. However, if the Developer field **is** null, then it would NOT be in the returned object. For example, a non-null developer with null fields would be returned as (note the 'null' values).

> "developer": {
>         "developerId": 378,
>         "developerCode": "1377",
>         "name": "DrScribe, Inc.",
>         "website": null,
>         "address": null,
>         "contact": null,
>         "lastModifiedDate": null,
>         "deleted": null,
>         "transparencyAttestations": [],
>         "statusEvents": [
>             {
>                 "id": null,
>                 "developerId": 378,
>                 "status": {
>                     "id": 1,
>                     "status": "Active"
>                 },
>                 "statusDate": 1459484374776,
>                 "reason": null
>             }
>         ],
>         "status": {
>             "id": 1,
>             "status": "Active"
>         }
>     }

There are several ways to handle this

1. It's a non issue - it's OK that null class-level fields are excluded from the returned object
2. We instantiate the member fields with default values, like we do with the lists, and explicitly null-out when and where needed
3. The CertifiedProductSearchDetails object is mapped to a NEW object that does not have those fields and is returned by the controller

I'm absolutely cool with whatever approach makes sense. Now that I see this object has a bigger reach then I knew, I might lean towards mapping to a new object. But, if this implementation of using the Jackson annotation makes sense, then that's cool too. Thanks!

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
```diff
@@ -346,6 +346,8 @@ public CertifiedProductController(final CertifiedProductUploadHandlerFactory upl
 
         certifiedProduct = validateCertifiedProduct(certifiedProduct);
```
andlar: Why validate this product when the next line removes the warning/error messages?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
```diff
@@ -370,9 +372,18 @@ public CertifiedProductController(final CertifiedProductUploadHandlerFactory upl
...
 
+        removeNonBasicCertifiedProductSearchDetails(certifiedProduct);
+
         return certifiedProduct;
```
andlar: If we're still returning the same type of object, won't it still have the keys that we want removed?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
```diff
@@ -30,6 +32,7 @@
 @XmlType(namespace = "http://chpl.healthit.gov/listings")
 @XmlAccessorType(XmlAccessType.FIELD)
 @JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(Include.NON_NULL)
```
andlar: There are some fields that have null values for certified products but are expected by the front end; will this annotation remove those?

commitlint.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
andlar: This file should be removed. It was here when we were using commitlint as a hook, but we got rid of that when we went to squash merge PRs

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertifiedProductController.java
Plow74: No, because of the Jackson annotation on the DTO tells the mapper to exclude the class field members that are null.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
Plow74: Yeah, probably. I wasn't aware that this object is used in many places, not just the controller. So, with that in mind, I think I'm going to refactor this to use a new object that doesnt have those fields and have the controller methods in question return that object.

Plow74:
> I see two places where those data points are removed, but I think there are three endpoints that have the basic data. You're missing public @responsebody CertifiedProductSearchDetails getCertifiedProductByIdBasic(

GET:certified_products/{chplPrefix}-{identifier}
GET:/certified_products/{certifiedProductId}
GET:/certified_products/{year}.{testingLab}.{certBody}.{vendorCode}.{productCode}.{versionCode}.{icsCode}.{addlSoftwareCode}.{certDateCode}

I do not see the second one in the controller, "GET:/certified_products/{certifiedProductId}". There is a 'PUT' for this endpoint, but not a 'GET'.

andlar:
Lines 274-289 in my local controller:
```java
    @ApiOperation(value = "Get all basic information for a specified certified product.  Does not include "
            + "the CQM results and certification results.",
            notes = "Returns basic information in the CHPL related to the specified certified product.  "
                    + "The results will not include the CQM results and certification results.")
    @RequestMapping(value = "/{certifiedProductId:^-?\\d+$}",
    method = RequestMethod.GET,
    produces = "application/json; charset=utf-8")
    @CacheControl(policy = CachePolicy.PUBLIC, maxAge = CacheMaxAge.TWELVE_HOURS)
    public @ResponseBody CertifiedProductSearchDetails getCertifiedProductByIdBasic(
            @PathVariable("certifiedProductId") final Long certifiedProductId) throws EntityRetrievalException {

        CertifiedProductSearchDetails certifiedProduct = cpdManager.getCertifiedProductDetailsBasic(certifiedProductId);
        certifiedProduct = validateCertifiedProduct(certifiedProduct);

        return certifiedProduct;
    }
```
It looks like lines 283-298 or so in this PR

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchBasicDetails.java
```diff
@@ -0,0 +1,752 @@
...
+import javax.xml.bind.annotation.XmlType;
+
+/**
+ * Certified Product Search Basic Details entity.
+ */
+@XmlType(namespace = "http://chpl.healthit.gov/listings")
+@XmlAccessorType(XmlAccessType.FIELD)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CertifiedProductSearchBasicDetails implements Serializable {
+    private static final long serialVersionUID = 2903219171127034775L;
```
kekey1: Can you change this value? Looks like it's the same as the one in CertifiedProductSearchDetails

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchBasicDetails.java
```diff
@@ -0,0 +1,752 @@
...
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlElementWrapper;
+import javax.xml.bind.annotation.XmlTransient;
+import javax.xml.bind.annotation.XmlType;
+
+/**
+ * Certified Product Search Basic Details entity.
+ */
+@XmlType(namespace = "http://chpl.healthit.gov/listings")
```
kekey1: All these XML-related annotations probably do not need to be in this class. The original CertifiedProductSearchDetails gets written out nightly to an XML file which is why all the annotations are here with details comments, but I don't think this class will need any of that.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchBasicDetails.java
```diff
@@ -0,0 +1,752 @@
...
+@XmlAccessorType(XmlAccessType.FIELD)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CertifiedProductSearchBasicDetails implements Serializable {
+    private static final long serialVersionUID = 2903219171127034775L;
+    public static final String ACB_ID_KEY = "id";
```
kekey1: I don't think this class needs any of these constants as they are all currently referenced from the other class.

PR #817, opened 2019-10-24
feat: Add endpoint to retrieve API Key activity metadata

Plow74:
OCD-2960


Plow74:
> Two things:
> 
> The title of the PR needs to have the right tense / format to match what's in our RELEASE_NOTES. The way I think of it is filling in the blanks with the phrase, "If applied, this PR will `insert action here`". It also should be specific to the project. In this case, the fact that the UI uses the new endpoint to make a report has nothing to do with the API. I'd phrase this title as "feat: Add endpoint to retrieve API Key activity metadata", and probably put a bullet in the contents with something like "* Deprecate /api_keys activity call". Also, watch out when merging this eventually, because when you go into the "PR Merge" screen it usually changes the title on you; I'm not sure why
> 
> Second, we'd talked about removing `impl` files as we come across them. Do you want to do that in this PR for the ActivityMetadataManager? Not requesting it specifically, as it's probably a bigger lift, but...

Changed the title - for some reason the title made sense when I was creating this in regards to, "If applied, this PR will `insert action here`" but I can see that it doesn't since you pointed it out.

As far as the impl refactor - after talking with Todd, I thought I was going to be creating new classes and those new classes would not follow the impl pattern. Since I did not create any new classes, I decided to keep the pattern in place.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ActivityController.java
```diff
@@ -809,6 +810,18 @@ public ActivityDetails activityById(@PathVariable("id") final Long id)
...
+    @ApiOperation(value = "Get auditable data for all API keys",
+        notes = "Users must specify 'start' and 'end' parameters to restrict the date range of the results."
+            + "Security Restrictions: Only ROLE_ADMIN or ROLE_ONC")
+    @RequestMapping(value = "/metadata/api-keys", method = RequestMethod.GET, produces = "application/json; charset=utf-8")
+    public List<ActivityDetails> metadataForApiKeys(@RequestParam final Long start,
+        @RequestParam final Long end) throws JsonParseException, IOException, ValidationException {
+        Date startDate = new Date(start);
+        Date endDate = new Date(end);
+        validateActivityDatesAndDateRange(start, end);
+        return activityManager.getApiKeyActivity(startDate, endDate);
```
kekey1: Should the new getApiKeyActivityMetdata method be called here? And ActivityMetadata object be returned?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/ActivityMetadataManager.java
```diff
@@ -49,4 +49,6 @@
     List<ActivityMetadata> getAnnualReportActivityMetadata(Date startDate, Date endDate) throws IOException;
 
     List<ActivityMetadata> getChangeRequestActivityMetadata(Date startDate, Date endDate) throws IOException;
+
+    List<ActivityMetadata> getApiKeyManagementMetaData(Date startDate, Date endDate) throws IOException;
```
kekey1: Can you make this method name capitalization consistent with the others? (Lowercase 'd" in Metadata)

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ActivityController.java
```diff
@@ -809,6 +810,18 @@ public ActivityDetails activityById(@PathVariable("id") final Long id)
         return activityManager.getApiKeyActivity(startDate, endDate);
     }
 
+    @ApiOperation(value = "Get auditable data for all API keys",
+        notes = "Users must specify 'start' and 'end' parameters to restrict the date range of the results."
+            + "Security Restrictions: Only ROLE_ADMIN or ROLE_ONC")
+    @RequestMapping(value = "/metadata/api-keys", method = RequestMethod.GET, produces = "application/json; charset=utf-8")
+    public List<ActivityDetails> metadataForApiKeys(@RequestParam final Long start,
```
tmy1313: Based on our phone conversation...  This needs to be updated to return an "ApiActivityMetadata" object.

PR #818, opened 2019-10-25
flag: prevent ROLE_ACB from editing 2014 Edition

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/Permissions.java
```diff
@@ -125,4 +125,12 @@ public boolean hasAccess(final String domain, final String action, final Object
         }
     }
 
+
+    public boolean hasAccess(final String domain, final String action, final Object obj, final Object obj2) {
+        if (domainPermissions.containsKey(domain)) {
+            return domainPermissions.get(domain).hasAccess(action, obj, obj2);
+        } else {
```
drbgfc: This may be pedantic, and therefore is not something you _need_ to change, but, the else statement is not required in this case

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/DomainPermissions.java
```diff
@@ -22,6 +22,14 @@ public boolean hasAccess(final String action, final Object obj) {
         }
     }
 
+    public boolean hasAccess(final String action, final Object obj, final Object obj2) {
+        if (getActionPermissions().containsKey(action)) {
+            return getActionPermissions().get(action).hasAccess(obj, obj2);
+        } else {
```
drbgfc: Same thing here - could just say 'return false'

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/Permissions.java
andlar: While that's true, it's also consistent with the way the other `hasAccess` methods are structured, so if I _did_ do it that way I'd feel required to change all of the other ones too

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/certifiedproduct/UpdateActionPermissions.java
```diff
@@ -1,29 +1,58 @@
...
+        } else if (getResourcePermissions().isUserRoleAcbAdmin()) {
+            Long acbId = (Long) obj;
+            if (ff4j.check(FeatureList.EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK)) {
+                CertifiedProductSearchDetails listing = (CertifiedProductSearchDetails) listingObj;
+                return !listing.getCertificationEdition().get(CertifiedProductSearchDetails.EDITION_NAME_KEY)
+                        .toString().equalsIgnoreCase("2014") && isAcbValidForCurrentUser(acbId);
+            } else {
+                return isAcbValidForCurrentUser(acbId);
+            }
+        } else {
```
drbgfc: This else is also not needed. I won't mention it in future PRs if you'd prefer as it appears to be your style/a choice you made. I'm sure the compiler weeds it out anyway. And you may prefer it readability wise, I just prefer not having it I suppose because it's faster to read and not necessary. Anyway, not important...

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/Permissions.java
drbgfc: Yeah I see that now too - approved the PR (not that I would have held it back over this anyway)

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
```diff
@@ -15,7 +15,7 @@ private FeatureList() {}
     /** Features that will be turned on as of the effective rule date. */
     public static final String EFFECTIVE_RULE_DATE = "effective-rule-date";
     /** Features that will be turned on as of one week past the effective rule date. */
-    public static final String EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK = "effective-rule-date+1-week";
+    public static final String EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK = "effective-rule-date-plus-on-week";
```
drbgfc: If this is the latest - should be 'one' vs 'on'

PR #820, opened 2019-10-30
feat: Add job to mark specific 2015 criteria as removed

kekey1:
* Include "removed" field in XSD, XML file generation
* Include "removed" field in /details response
* Include criteria (with "removed" field) in application /data/* API calls.

OCD-3148


chpl/chpl-resources/src/main/resources-dev/jobs.xml
```diff
@@ -706,5 +706,25 @@
...
+            <description>Sets the removed flag to 'true' for specific 2015 criteria</description>
+            <job-class>gov.healthit.chpl.scheduler.job.RemoveCriteriaJob</job-class>
+            <durability>true</durability>
+            <recover>false</recover>
+            <job-data-map>
+                <entry>
+                    <key>authorities</key>
+                    <value>ROLE_ADMIN;ROLE_ONC</value>
+                </entry>
+                <entry>
```
andlar: Shouldn't need this entry; this is a one time trigger, the frequency key is only for recurring ones

andlar: Probably. Worth checking with @tmy1313 

I'm pretty sure it's only the jobs we have set up as cron jobs that can use this field at all

kekey1: Confirmed with Todd and he will remove that field as part of the 2014 update ticket since that's using the job. Made the change for this job.

PR #821, opened 2019-10-31
flag: remove "surveillance-reporting" flag

andlar:
[#OCD-3113]

### DIFF
--- a/chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ActivityController.java
+++ b/chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ActivityController.java
@@ -108,10 +108,10 @@ public class ActivityController {
 
     @ApiOperation(value = "Get detailed audit data for a specific activity event.",
             notes = "Security Restrictions: ROLE_ADMIN and ROLE_ONC may view any activity event. "
-                + "Other users may be restricted in what they can see.")
+                    + "Other users may be restricted in what they can see.")
     @RequestMapping(value = "/{id}", method = RequestMethod.GET, produces = "application/json; charset=utf-8")
     public ActivityDetails activityById(@PathVariable("id") final Long id)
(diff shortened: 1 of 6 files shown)


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ActivityController.java
```diff
@@ -1210,8 +1203,8 @@ public ActivityDetails activityById(@PathVariable("id") final Long id)
 
     @Deprecated
     @ApiOperation(value = "DEPRECATED. Track the actions of all users in the system",
-            notes = "Users must specify 'start' and 'end' parameters to restrict the date range of the results."
```
drbgfc: No reason to hold back a merge but a note: The formatting changes outweigh the update in this specific file which makes it hard to see the relevant changes. Due to that, I'm wondering, as per our documentation for the selected eclipse formatting, https://git.ainq.com/chpl/resources/blob/master/eclipse/eclipse_style.xml - is your new version more correct? Using both that style guide in my case reverts to the prior indentation. Maybe I'm doing something wrong.

andlar: I'm pretty sure I have the style guide set up correctly. In this file I was tired of manually indenting things, so I did a Ctrl-A Ctrl-I to auto-indent stuff, and this is what it outputted. If it's not the right style for our team then I don't know what to do about it

If it helps, when reviewing the PR, there's a setting at the top of the Files Changed section where you can hide whitespace-only changes. I don't use that much, but for something like this it's definitely useful

drbgfc: That's a **great** feature. Good to know. Thanks.
Is your formatter active profile set to CHPL style (the imported XML)? I'm not sure how directly importing the same file could have 2 different outcomes. Kind of defeats the purpose.

PR #827, opened 2019-11-14
feat-flag: Add new error to Inheritance Errors Report when ICS=0 and Gap=1 for Any Criteria and Cert Date after Cures rule effective date

tmy1313:
OCD-3152


chpl/chpl-service/src/main/java/gov/healthit/chpl/CHPLServiceConfig.java
```diff
@@ -169,6 +182,15 @@ public LocaleChangeInterceptor localeInterceptor() {
         return interceptor;
     }
 
+    @Bean
+    public InternalResourceViewResolver viewResolver() {
```
kekey1: Do we need this?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/InheritanceErrorsReportCreatorJob.java
```diff
@@ -60,131 +57,111 @@
...
-        }
-        return props;
-    }
-
-    private List<CertifiedProductSearchDetails> retrieveData() {
+        inheritanceErrorsReportDAO.deleteAll();
         List<CertifiedProductFlatSearchResult> listings = certifiedProductSearchDAO.getAllCertifiedProducts();
         List<CertifiedProductFlatSearchResult> certifiedProducts = filterData(listings);
-        LOGGER.info("2015 Certified Product Count: " + certifiedProducts.size());
+        certifiedProducts = certifiedProducts.subList(0, 100);
```
kekey1: This seems like something that was for testing and should be removed

chpl/chpl-service/src/main/java/gov/healthit/chpl/CHPLServiceConfig.java
tmy1313: I honestly do not know...  This was pulled over from the AppConfig.java which was deleted.  This was the only bean in that config file that was not in CHPLServiceConfig.java., so I thought I should bring it over just in case.  I think we need to look at our Spring config files - there are 4 or so different files and I think we can combine them.

PR #828, opened 2019-11-15
feat-flag!: Restrict ACBs from altering Transparency Attestation

drbgfc:
[#OCD-3164]


chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
```diff
@@ -1,15 +1,23 @@
...
+ * All relevant FF4J feature flags.
  * @author alarned
  *
  */
-public class FeatureList {
+public final class FeatureList {
+    private FeatureList() {}
+
+    /** Feature related to submitted Complaints. */
+    public static final String COMPLAINTS = "complaints";
```
andlar: I think the complaints flag has been deployed to STG, so shouldn't 'be here

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
```diff
@@ -1,15 +1,23 @@
...
     public static final String CHANGE_REQUEST = "change-request";
     /** Features that will be turned on as of the effective rule date. */
     public static final String EFFECTIVE_RULE_DATE = "effective-rule-date";
+    /** Features that will be turned on as of one week past the effective rule date. */
+    public static final String EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK = "effective-rule-date-plus-one-week";
     /** Ability to have Developers as User. */
     public static final String ROLE_DEVELOPER = "role-developer";
-}
+    /** Function for ONC-ACBs to generate their required quarterly and annual surveillance reporting. */
+    public static final String SURVEILLANCE_REPORTING = "surveillance-reporting";
```
andlar: Same with this one

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -221,6 +225,17 @@ public DeveloperDTO update(final DeveloperDTO updatedDev, final boolean doValida
         }
 
         DeveloperDTO beforeDev = getById(updatedDev.getId());
+
+        // ROLE_ACB cannot edit Transparency Attestation as per OCD-3164
+        if (ff4j.check(FeatureList.EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK)) {
```
andlar: A lot of our validation stuff has been foisted off on validation classes, and removed from the managers themselves. I don't know if this needs to be placed there too, but wanted to mention it

I think the idea is to try to move validation / security out of managers into reviewers, but, TBH, a lot of the lines are blurring, so I don't know for sure. @kekey1 @tmy1313 , thoughts?

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
drbgfc: I thought this would be best to reduce merge conflicts - let's talk Monday

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
drbgfc: Interesting. I hadn't considered this a validation but instead just authentication and I wanted the authentication to be pre-validated so it was working with valid info. I consulted the current flow of the existing code which follows this pattern. Scrolling down in the same method shows various snippets which generally do similar things in the same method (check for authentication of some role and/or if values have changed and throw EntityCreationExceptions),. So I'm thiking if we wanted to move my new code to validation, that's fine, but we'd probably want to move the rest too.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
```diff
@@ -780,6 +795,33 @@ public String toString() {
...
+                                    changedMapping.getTransparencyAttestation())) {
+                        return true;
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    private static boolean isBasicPropertyChanged(final Object originalProperty, final Object otherProperty) {
```
kekey1: There is an ObjectUtils.equals or maybe it's isEqual method in I think either a Spring or Apache library and I wonder if that does something similar to this and you could use it instead? If not, may want to consider moving this method into a Utils class of our own as it seems like a utility

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
kekey1: I'd be fine with moving this into a permissions class and refactoring other code in the same method to be in there as well but would like to be conscious of not letting it get to be too much for this one ticket. Consider how much regression testing would have to happen if we move the code (I don't know how much it is to be moved) and update test cases accordingly if we decide to go this route.

drbgfc: I see now you said slash security. If there is some externalized security area maybe we move it and the other code there.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
drbgfc: I will do one of those two things.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
drbgfc: In response to Katy, I agree on the possible issue with ticket size i this case because it is a 2 part ticket and has a front-end side as well that must be done after the backend is tested. That being said, I don't personally object, it will just take time. And I don't know that there is any basis for  a security/auth/permissions class yet or if it would be something brand new? Sounds new?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/DeveloperManagerImpl.java
drbgfc: Thanks - I ended up using the JDK7 Objects.equals method for this as you were likely referencing. There is a Spring one too called nullSafeEquals that is the same but can't really beat using the JDK from a dependency standpoint. All I had to do was reverse the result since I was checking for a change not equality.

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
drbgfc: removed

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
drbgfc: removed

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
```diff
@@ -1,15 +1,19 @@
...
     /** System for Developers to submit change requests. */
     public static final String CHANGE_REQUEST = "change-request";
     /** Features that will be turned on as of the effective rule date. */
     public static final String EFFECTIVE_RULE_DATE = "effective-rule-date";
+    /** Features that will be turned on as of one week past the effective rule date. */
+    public static final String EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK = "effective-rule-date-plus-one-week";
     /** Ability to have Developers as User. */
     public static final String ROLE_DEVELOPER = "role-developer";
-}
+}
```
andlar: I think checkstyle doesn't like files that don't end with a newline, but I could be wrong

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperEditTransparencyAttestationValidation.java
```diff
@@ -0,0 +1,63 @@
...
+    private ResourcePermissions resourcePermissions;
+
+    public DeveloperEditTransparencyAttestationValidation(final FF4j ff4j,
+            final ResourcePermissions resourcePermissions) {
+        this.ff4j = ff4j;
+        this.resourcePermissions = resourcePermissions;
+    }
+
+    /**
+     * ROLE_ACB cannot edit Transparency Attestation as per OCD-3164
```
andlar: I don't like this comment. OCD-3164 is not visible to the public, but our code is, and the comment itself doesn't describe the function accurately, in that the function also depends on a flag state, whereas the comment has no mention of that. I'd prefer just getting rid of it entirely, especially as the function itself is small enough that it's self documenting

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperEditTransparencyAttestationValidation.java
```diff
@@ -0,0 +1,63 @@
...
+                    getMessages().add(msgUtil.getMessage("developer.transparencyAttestationEditNotAllowedForRoleACB"));
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    private static boolean isTransparencyAttestationUpdated(final DeveloperDTO original, final DeveloperDTO changed) {
+        if ((original.getTransparencyAttestationMappings() != null
```
andlar: What if we changed "getTransparencyAttestationMappings" to return a list that was empty if there weren't any, instead of null? Then this entire set of things could be reduced to `if (original.getmaps.size != changed.getmaps.size)` and we don't have to worry about all the null checks?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperPriorStatusActiveValidation.java
```diff
@@ -0,0 +1,41 @@
...
+public class DeveloperPriorStatusActiveValidation extends ValidationRule<DeveloperValidationContext> {
+    private static final Logger LOGGER = LogManager.getLogger(DeveloperPriorStatusActiveValidation.class);
+    private ResourcePermissions resourcePermissions;
+
+    public DeveloperPriorStatusActiveValidation(final ResourcePermissions resourcePermissions) {
+        this.resourcePermissions = resourcePermissions;
+    }
+
+    /**
+     * If the before status is not Active and the user is not ROLE_ADMIN then
```
andlar: ROLE_ONC is missing from this comment. Suggest removing the comment entirely

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperStatusChangedValidation.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public DeveloperStatusChangedValidation(final DeveloperManagerImpl developerManagerImpl,
+            final ResourcePermissions resourcePermissions, final DeveloperDAO developerDao) {
+        this.resourcePermissions = resourcePermissions;
+    }
+
+    /**
+     * determine if the status has been changed in most cases only allowed by
+     * ROLE_ADMIN but ROLE_ACB can change it to UnderCertificationBanByOnc
```
andlar: ROLE_ONC missing from comment as well

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperStatusChangedValidation.java
```diff
@@ -0,0 +1,59 @@
...
+        if (currentStatusChanged
+                && !newDevStatus.getStatus().getStatusName()
+                        .equals(DeveloperStatusType.UnderCertificationBanByOnc.toString())
+                && !resourcePermissions.isUserRoleAdmin() && !resourcePermissions.isUserRoleOnc()) {
+            String msg = msgUtil.getMessage("developer.statusChangeNotAllowedWithoutAdmin");
+            getMessages().add(msg);
+            return false;
+        }
+        // Check to see that the Developer's website is valid.
+        if (!StringUtils.isEmpty(updatedDev.getWebsite())) {
```
andlar: What does the website have to do with the status change? Should be extracted

andlar: In `public ValidationRule<DeveloperValidationContext> getRule(String name) {` there's already a case for "Website required"

chpl/chpl-service/src/test/java/gov/healthit/chpl/manager/impl/DeveloperStatusTest.java
```diff
@@ -76,6 +77,9 @@
     @Autowired
     private DeveloperValidationFactory developerValidationFactory;
 
+    @Autowired
+    private FF4j ff4j;
```
andlar: I don't see this used anywhere in the file

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
drbgfc: Agreed - noticed that but left as is to match the way it was in the remote repo (no newline) so there is no merge conflict since there are changes from different people within the file.  I suppose that was pointless as git should be able to handle that.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperEditTransparencyAttestationValidation.java
drbgfc: Remove the comment and pushed change

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
andlar: But you didn't leave it the way it was, because git is showing it as a change from having a newline to not having a newline. Otherwise it wouldn't have shown up as a changed line...

drbgfc: Added and pushed

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperEditTransparencyAttestationValidation.java
drbgfc: Seems like a good idea. I really like appreciate these kinds of comments the most, but:
-This isn't my code nor is it directly related to the ticket, thus, my focus was on refactoring the code into using the validation factory paradigm, and at the same time, on not changing the actual implementation. This way, code we know works, should still work in regression testing, especially since these changes are out of scope.
-What if something else calls getTransparencyAttestationMappings and depends on how it is now? What about implied verbal contracts of getters in general?
--To expand on my own questions: For the first one it's probably not the case that something depends on null, and if it does, maybe that's a problem in some circles. And the seconds one, that's a philosphy type thing too.

That being said, as usual, I'm not against it. We are going to regression test anyway. I will just open the discussion up to the second reviewer before I implement.

drbgfc: Note: This code length could probably also be reduced using Objects.equals() and that would add the case of double null as well

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperPriorStatusActiveValidation.java
drbgfc: Left as is as figured writer of the impl knew better than me. I lke the idea of fixing the comment (adding ROLE ONC) but since there is a property referenced and one can look up the error, I'll ditch it as requested.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperStatusChangedValidation.java
drbgfc: Removed and pushed

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperStatusChangedValidation.java
drbgfc: Yeah I agree - I don't know why it was there.  It used to be an an else condition so intended to be a part of the same routine by original writer so I left together (but removed else as redundant in return/exception type situation). Assuming there is no valid reason for it, If that exact validation is already run prior, then I won't call again. If it is not run, I'll call it externally. Either way, will remove from here.

chpl/chpl-service/src/test/java/gov/healthit/chpl/manager/impl/DeveloperStatusTest.java
drbgfc: Good catch - was part of an older impl. Will remove. Although, won't be surpised if it needs to be added back for new tests.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperStatusChangedValidation.java
drbgfc: There's 2 different website validations. Well-formed and required. Well-formed is called by update where as only system validations use required. This extra code represents the well-formed pattern and validates of the same version (updated) of the developer and therefore can be removed altogether/doesn't need to be repeated/externalized. I will remove it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperEditTransparencyAttestationValidation.java
andlar: I'm fine not changing this for this PR, but I don't know if we'll have any other ticket that'd cover this section of the code any time soon. Your ticket is related to Transparency Attestations, so it's at least tangentially appropriate here.

Part of our process is to improve things as we come across them; maybe see how many references there are to this method in the code base, and if it's, like 1 or 2 that'd be a reasonable thing to update, but if it's 5+ that might make a difference

kekey1: If it's quick and the test cases will already regression test whatever is affected then I'm all for changing it. Agree if there's a lot of references or if it turns into one thing after another that gets changed then I wouldn't bother.

drbgfc: Good points - sure thing, thanks

drbgfc: There are 15 other independent calls to getTransparencyAttestationMappings() in the codebase so now I'm a bit hesitant to override it? As per your comment @andlar "maybe see how many references there are to this method in the code base, and if it's, like 1 or 2 that'd be a reasonable thing to update, but if it's 5+ that might make a difference" maybe we shouldn't change its impl then.
I looked into Objects.equals for the null parts and it seemed like it could work on this object (since it's a list) if we also overrode the equals method for the type (DeveloperACBMapDTO) to basically just chek size. Then we could use that as a one liner for the first set of checks. But, that's a no go because it also returns true if both are null which is great for property checks but would cause our loop to run on double null unless we add a condition to the else (which would we could to solve it). But, i'm not sure this is any more clear then the current impl. It's kind of convoluted at that point. 
I also looked into simplifyihng the else condition into one line of CollectionUtils.isEqualCollection(original, changed). The only concern there is that for now we are only checking getAcbName and getTransparencyAttestation where as isEqualCollection would check every property, and that probably will not get the result we like (and may be overkill). So would need testing. So anyway, I know I just said a lot but I'm not so sure any changes are neccessary or better at this point. But here's the options (besides overriding the getter) What do you guys think? I can look into/test if you think it's worth the time.

```
    private static boolean isTAUpdatedAlt(final DeveloperDTO original, final DeveloperDTO changed) {
        // override equals to just check size
        if (!Objects.equals(original.getTransparencyAttestationMappings(),
                changed.getTransparencyAttestationMappings())) {
            return true;
        }
        if (original.getTransparencyAttestationMappings() != null
                && changed.getTransparencyAttestationMappings() != null) {
            for (DeveloperACBMapDTO originalMapping : original.getTransparencyAttestationMappings()) {
                for (DeveloperACBMapDTO changedMapping : changed.getTransparencyAttestationMappings()) {
                    if (!Objects.equals(originalMapping.getAcbName(), changedMapping.getAcbName())
                            || !Objects.equals(originalMapping.getTransparencyAttestation(),
                                    changedMapping.getTransparencyAttestation())) {
                        return true;
                    }
                }
            }
        }
        return false;
    }

    private static boolean isTAUpdatedAlt2(final DeveloperDTO original, final DeveloperDTO changed) {
        if ((original.getTransparencyAttestationMappings() != null
                && changed.getTransparencyAttestationMappings() == null)
                || (original.getTransparencyAttestationMappings() == null
                        && changed.getTransparencyAttestationMappings() != null)
                || (original.getTransparencyAttestationMappings().size() != changed.getTransparencyAttestationMappings()
                        .size())) {
            return true;
        }
        if (original.getTransparencyAttestationMappings() != null
                && changed.getTransparencyAttestationMappings() != null) {
            return CollectionUtils.isEqualCollection(original.getTransparencyAttestationMappings(),
                    changed.getTransparencyAttestationMappings());
        }
        return false;

    }

    private static boolean isTAUpdatedAltMixed(final DeveloperDTO original, final DeveloperDTO changed) {
        // override equals to just check size
        if (!Objects.equals(original.getTransparencyAttestationMappings(),
                changed.getTransparencyAttestationMappings())) {
            return true;
        }
        if (original.getTransparencyAttestationMappings() != null
                && changed.getTransparencyAttestationMappings() != null) {
            return CollectionUtils.isEqualCollection(original.getTransparencyAttestationMappings(),
                    changed.getTransparencyAttestationMappings());
        }
        return false;
    }
```

andlar: Looks like the best path forward is to do nothing with this set of things just yet. That's too many other places, and I don't see a lot of need in overriding all these other things either. I'm going to resolve this conversation and approve the PR as it stands so we can think about getting this part into the testing state

drbgfc: Here's what the getter version would look like (I did this in like 5 secs so might be a type):
    private static boolean isTAUpdatedGetterAlt(final DeveloperDTO original, final DeveloperDTO changed) {

>         // override getTransparencyAttestationMappings to return an empty list
>         // vs null
>         if (original.getTransparencyAttestationMappings().size() != changed.getTransparencyAttestationMappings()
>                 .size()) {
>             return true;
>         }
>         if (original.getTransparencyAttestationMappings() != null
>                 && changed.getTransparencyAttestationMappings() != null) {
>             for (DeveloperACBMapDTO originalMapping : original.getTransparencyAttestationMappings()) {
>                 for (DeveloperACBMapDTO changedMapping : changed.getTransparencyAttestationMappings()) {
>                     if (!Objects.equals(originalMapping.getAcbName(), changedMapping.getAcbName())
>                             || !Objects.equals(originalMapping.getTransparencyAttestation(),
>                                     changedMapping.getTransparencyAttestation())) {
>                         return true;
>                     }
>                 }
>             }
>         }
>         return false;
>     }

drbgfc: Sure thing - good liittle thought excercise if nothing else

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperEditTransparencyAttestationValidation.java
```diff
@@ -46,12 +47,14 @@ private static boolean isTransparencyAttestationUpdated(final DeveloperDTO origi
...
-                            || !Objects.equals(originalMapping.getTransparencyAttestation(),
-                                    changedMapping.getTransparencyAttestation())) {
-                        return true;
-                    }
+                Optional<DeveloperACBMapDTO> matchingMapping =
+                        changed.getTransparencyAttestationMappings().stream()
+                    .filter(changedMapping -> StringUtils.equals(originalMapping.getAcbName(), changedMapping.getAcbName()))
+                    .findFirst();
+                if (!matchingMapping.isPresent()
+                        || !StringUtils.equals(originalMapping.getTransparencyAttestation(),
```
drbgfc: I believe there was a reason I used Objects.equals vs StringUtils.equals in the way null comparisons were done (specifically when both items were null). Do the relevant JUnits still pass with this code? Specifically here: https://github.com/chpladmin/chpl-api/pull/828/commits/6c63b0dcb6f432eed3b0b882f84e7197a08d4a87#diff-c352826099c6157d2b675926ef7bf2f7

PR #830, opened 2019-11-21
feat!: Update Questionable Activity Report to support a user defined range of days

tmy1313:
OCD-3199


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/impl/SchedulerManagerImpl.java
```diff
@@ -15,6 +15,7 @@
 import java.util.Set;
 
 import org.quartz.CronTrigger;
+import org.quartz.JobDataMap;
```
andlar: Should this "Impl" code be put into the "non-Impl" version?

chpl/chpl-resources/src/main/resources-production/jobs.xml
```diff
@@ -27,6 +27,111 @@
                 </entry>
             </job-data-map>
         </job>
+
+        <!-- Questionable URL Data Collector -->
+        <!-- URL Status Data Collector -->
+        <!-- Job used to change status of a set listings -->
+        <job>
```
andlar: Why are all of these in this PR? Shouldn't they either be in a separate PR, or if already on staging, not in a PR at all?

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -394,9 +393,16 @@
                     <key>frequency</key>
                     <value>WEEKLY</value>
                 </entry>
+                <entry>
+                    <key>range</key>
+                    <value>7</value>
+                </entry>
+                <entry>
+                    <key>parameters</key>
+                    <value>[{"label": "Range (in days)", "key": "range", "type": "int", "min": "7"} ]</value>
```
andlar: No specific change request, but why have a min but no max? I thought the ticket talked about a max value?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -133,6 +134,14 @@ public void execute(final JobExecutionContext jobContext) throws JobExecutionExc
         LOGGER.info("********* Completed the Questionable Activity Email job. *********");
     }
 
+    private Integer getRangeInDays(final JobExecutionContext jobContext) {
+        Integer range = 7;
+        if (jobContext.getMergedJobDataMap().containsKey("range")) {
+            range = jobContext.getMergedJobDataMap().getInt("range");
+        }
+        return range * -1;
```
andlar: There's no validation in here. I'm not sure if there should be, but we can't assume the UI will provide "valid" numbers. My guess is that we should have a min value and a max value, defined as environment.properties values, and ensure that whatever the end user passes in falls within that range, or is constrained to that range by the code, if we want to allow that

chpl/chpl-resources/src/main/resources-production/jobs.xml
kekey1: Just to follow up on this - this file as it exists on staging right now has 2 jobs in it.

chpl/chpl-resources/src/main/resources/jobs.xml
tmy1313: I think the range should be 1-365.  Updated the code to reflect this.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
tmy1313: Updated the code to allow a range of 1-365.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -41,9 +41,9 @@
 import gov.healthit.chpl.util.Util;
 
 /**
- * The QuestionableActivityEmailJob implements a Quartz job and is available to ROLE_ADMIN. When
- * invoked it emails relevant individuals with the questionable activity that has occurred within
- * the last week.
+ * The QuestionableActivityEmailJob implements a Quartz job and is available to ROLE_ADMIN. When invoked it emails
+ * relevant individuals with the questionable activity that has occurred within the last week.
```
andlar: Comment is not correct any more

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -73,9 +73,14 @@
     private static final int ACTIVITY_CERT_STATUS_CHANGE_REASON_COL = 11;
     private static final int ACTIVITY_REASON_COL = 12;
 
+    private static final int MIN_RANGE_IN_DAYS = 1;
```
andlar: Is there any way to pull these values from the jobs.xml definition? If we ever change them (for whatever reason) we'd have to change them in two places. I'm also thinking about what happens with other jobs that we might add range values for; it'd be nice to have a paradigm for having those values defined in only one place

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -122,17 +133,33 @@ public void execute(final JobExecutionContext jobContext) throws JobExecutionExc
...
+                    .sendEmail();
 
         } catch (MessagingException e) {
             LOGGER.error(e);
         }
         LOGGER.info("********* Completed the Questionable Activity Email job. *********");
     }
 
+    private Integer getRangeInDays(final JobExecutionContext jobContext) {
+        Integer range = 7;
```
andlar: This looks like a candidate for a CONSTANT value to use as default if the range is missing

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -88,9 +93,15 @@ public void execute(final JobExecutionContext jobContext) throws JobExecutionExc
         LOGGER.info("********* Starting the Questionable Activity Email job. *********");
         LOGGER.info("Creating questionable activity email for: " + jobContext.getMergedJobDataMap().getString("email"));
 
+        Integer range = getRangeInDays(jobContext);
+        if (!isRangeValid(range)) {
+            LOGGER.error(
+                    "Range is invalid.  It must be between " + MIN_RANGE_IN_DAYS + " and " + MAX_RANGE_IN_DAYS + ".");
+            return;
```
andlar: Thinking about the user experience, if someone managed to create a job with an invalid range there's no way they'd ever know it, right? As in, when the job starts it'd have an invalid range and would then just end, without sending any message. Would it make sense to default to the DEFAULT_RANGE value and maybe even add a message to the email about that? We're defaulting them if they don't have a value at all - it might be reasonable to do the same thing if they have an invalid one

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -689,8 +674,28 @@ private Properties loadProperties() throws IOException {
         return props;
     }
 
-    private void setQuestionableActivityDAO(final QuestionableActivityDAO questionableActivityDao) {
-        this.questionableActivityDao = questionableActivityDao;
+    private void populateRangeDefaultsFromJobData(final JobExecutionContext context) {
```
andlar: What if this function returned a `org.apache.commons.lang3.Range` object, or something similar? Then you could make the current "minRangeInDays and maxRangeInDays actual final static constants, and use the Range object returned from this function in the calling function to do a ".contains(range)" for the actual value

chpl/chpl-resources/src/main/resources-production/jobs.xml
```diff
@@ -25,6 +52,18 @@
...
+                    <key>frequency</key>
+                    <value>WEEKLY</value>
+                </entry>
+                <entry>
+                    <key>range</key>
+                    <value>7</value>
+                </entry>
+                <entry>
+                    <key>parameters</key>
+                    <value>[{"label": "Range (in days)", "key": "range", "type": "int", "min": "1", "max": "365"} ]</value>
```
andlar: This job doesn't have these ranges, right?

PR #835, opened 2019-12-10
OCD-3098: Prevent ONC-ACBs from adding/editing surveillance referencing removed criteria

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/reviewer/SurveillanceDetailsReviewer.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Autowired
+    public SurveillanceDetailsReviewer(CertifiedProductDAO cpDao, SurveillanceDAO survDao,
+            ChplProductNumberUtil chplProductNumberUtil, ErrorMessageUtil msgUtil) {
+        this.cpDao = cpDao;
+        this.survDao = survDao;
+        this.chplProductNumberUtil = chplProductNumberUtil;
+        this.msgUtil = msgUtil;
+    }
+
```
drbgfc: This is just a preference note, not a required change. I'm not sure what we've decided as a team but: I like to see @Override annotations when a method is overridden such as here here, from "void review(Surveillance surveillance)." It helps identify the difference and makes reading/understanding the code faster.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/PendingSurveillanceManager.java
```diff
@@ -1,31 +1,685 @@
...
+        // to know if we handle it normally or as a background job
+        String data = fileUtils.readFileAsString(file);
+
+        // This is a container used for 2 different result types...
+        SurveillanceUploadResult uploadResult = new SurveillanceUploadResult();
+
+        int numSurveillance = survUploadHelper.countSurveillanceRecords(data);
+        if (numSurveillance < surveillanceThresholdToProcessAsJob) {
+            uploadResult = processAsFile(file);
+        } else { // process as job
```
drbgfc: Normally I'm not the type to advocate against comments, but in this case, the comment is the same as the method name so I think it should probably be removed
// process as job == processUploadAsJob(data);

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/surveillance/UpdateActionPermissions.java
```diff
@@ -14,13 +29,25 @@ public boolean hasAccess() {
...
+                }
+            } catch (EntityRetrievalException ex) {
+                LOGGER.error("Could not find listing with ID " + surv.getCertifiedProduct().getId());
+            }
+            if (listing != null && listing.getCertificationBodyId() != null
+                    && isAcbValidForCurrentUser(listing.getCertificationBodyId())) {
+                return true;
+            }
+            return false;
         } else {
```
drbgfc: Understood not part of your change, but if I'm reading this correctly, technically, this else is not needed. Returning false alone should suffice.

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/surveillance/AddDocumentActionPermissions.java
```diff
@@ -20,11 +36,35 @@ public boolean hasAccess(Object obj) {
...
+                return false;
+            }
         } else {
             return false;
         }
     }
 
+    private boolean isNonconformityForRemovedCriteria(SurveillanceEntity surv, Long nonconformityId) {
+        if (!ff4j.check(FeatureList.EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK)) {
+            return false;
```
andlar: I know the ticket is flagged with this flag, but should security care? If there were a removed criteria today we'd still want to restrict ACBs from adding/removing documents related to that criteria, right?

PR #837, opened 2019-12-12
Ocd 3147 - add questionable activity when b3 is added to listing

tmy1313:


chpl/chpl-resources/src/main/resources/environment.properties
```diff
@@ -23,6 +23,7 @@ keyLocation=SECRET
 uploadErrorEmailRecipients=SECRET
 #format MM/DD/YYYY
 cures.ruleEffectiveDate=07/01/2019
```
andlar: I know it's not your change, but shouldn't we default this date to "way in the future"?

chpl/chpl-service/src/main/java/gov/healthit/chpl/listener/QuestionableActivityListener.java
```diff
@@ -4,7 +4,6 @@
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.aspectj.lang.annotation.After;
```
andlar: If the @After annotation went away, how does the Questionable Activity Listener get triggered?

chpl/chpl-resources/src/main/resources/environment.properties
tmy1313: I'll change it - probably just a typo.  I'll make it 07/01/2020

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/concept/QuestionableActivityTriggerConcept.java
```diff
@@ -4,126 +4,110 @@
...
+            "Certification Criteria 170.315 (b)(3) added with ICS for current listing"
+    ),
+    CRITERIA_B3_ADDED_TO_EXISTING_LISTING_WITHOUT_ICS(
+            "Certification Criteria 170.315 (b)(3) added without ICS for current listing"
+    ),
+    CRITERIA_B3_ADDED_TO_NEW_LISTING(
+            "Certification Criteria 170.315 (b)(3) added without ICS for new listing"
+    ),
+    ICS_ADDED_TO_EXISTING_LISTING_WITH_CRITERIA_B3(
+            "ICS added when certification criteria 170.315(b)(3) is on a current listing"
```
andlar:
```suggestion
            "ICS added when Certification Criteria 170.315(b)(3) is on a current listing"
```

PR #840, opened 2019-12-19
Ocd 3143- Adding additional metrics to the CHPL Weekly Summary Statistics report

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/SummaryStatisticsEmailJob.java
```diff
@@ -44,9 +46,9 @@
  *
  */
 public class SummaryStatisticsEmailJob extends QuartzJob {
-    private static final Logger LOGGER = LogManager.getLogger("summaryStatisticsEmailJobLogger");
-    private static final int EDITION2014 = 2014;
-    private static final int EDITION2015 = 2015;
+    private static Logger LOGGER = LogManager.getLogger("summaryStatisticsEmailJobLogger");
+    private static int EDITION2014 = 2014;
```
andlar: This should still have final, right? As it's a constant?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/AsynchronousSummaryStatisticsInitializor.java
```diff
@@ -45,14 +47,18 @@
...
     /**
      * Actual call to get the statistics.
-     * @param dateRange range to find statistics in
+     * 
+     * @param dateRange
+     *            range to find statistics in
      * @return a Future of type Statistics
-     * @throws InterruptedException if retrieval is interrupted
-     * @throws ExecutionException if execution fails
+     * @throws InterruptedException
```
andlar: Could we just remove these javadocs? Like, all of them? We're not requiring them, and if the only changes to them are reformatting the EOLs I'd rather just see that change as a delete once and for all

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/AsynchronousSummaryStatistics.java
```diff
@@ -365,204 +425,471 @@
...
+                .flatMap(req -> req.getNonconformities().stream())
+                .filter(nc -> nc.getCapApproval() != null)
+                .distinct() // Want to figure out how to get rid of this
+                .collect(Collectors.toList());
+
+        Long totalDuration = nonconformities.stream()
+                .map(nc -> Math
+                        .abs(ChronoUnit.DAYS.between(
+                                nc.getCapApproval().toInstant(),
+                                nc.getCapEndDate() != null ? nc.getCapEndDate().toInstant() : Instant.now())))
```
kekey1: What happens if capEndDate is null? In the method "getAverageTimeToCloseSurveillance" you specifically filter down to where an end date is not null so wondering if that should have been done above in the filtering stage of this method as well?

tmy1313: If the capEndDate is null we use the current date.  
Here is the requirement:
_Numerator:  Total number of days from CAP_Approval_Date to (Was_Complete_Date or today)_

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -138,7 +138,7 @@
             <job-data-map>
                 <entry>
                     <key>generateCsvFile</key>
-                    <value>true</value>
+                    <value>false</value>
```
andlar: Is this for ease of debugging locally?

tmy1313: Yes - generation of the CSV with all of the historical data take time.  Just generating the current statistics only takes a few seconds.

tmy1313: Yes - generation of the CSV with all of the historical data take time. Just generating the current statistics only takes a few seconds.

PR #848, opened 2020-01-07
OCD-3174: Do not allow add, edit, delete of surveillance for 2014 listings

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/surveillance/Surveillance.java
```diff
@@ -116,7 +116,60 @@ public Surveillance() {
      * @param anotherSurveillance
      * @return whether the two surveillance objects are the same
      */
-    public boolean matches(final Surveillance anotherSurveillance) {
+    public boolean matches(Surveillance anotherSurveillance) {
```
tmy1313: I'm not going to hold up this PR, but these loops that are looking for existence, non-existence, etc., could probably be written in a more functional manner.

PR #851, opened 2020-01-14
Ocd 3187 - Update All API calls to show Transparency Attestation has been removed

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperEditTransparencyAttestationValidation.java
```diff
@@ -47,17 +48,24 @@ private boolean isTransparencyAttestationUpdated(final DeveloperDTO original, fi
...
                                 matchingMapping.get().getTransparencyAttestation())) {
                     return true;
                 }
             }
         }
         return false;
     }
+
+    private Boolean areTransparencyAttestationsEqual(TransparencyAttestationDTO a, TransparencyAttestationDTO b) {
+        if (a == null || b == null) {
```
andlar: Shouldn't this be `&&`? How can they be equal if one exists and the other doesn't?

PR #852, opened 2020-01-14
[#OCD-3235] feat!: Add Quartz Job to insert new Test Tool

andlar:
[#OCD-3235]


drbgfc:
Aprroved but have a question: This seems to be written and named rather generically to support adding any Test Tool. So why is this preset? If wanting to add a different test tool are we expected to update the string and rebuild? Or are we planning on making it more dynamic in the future, or am I missing something?
```
  private static final String TEST_TOOL = "Inferno";
```

drbgfc:
Accendently clicked 'close and comment' vs just 'comment' - re-opened...

PR #856, opened 2020-01-16
[#OCD-3234] Add Self-Developer element to Developer information

andlar:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/DeveloperController.java
```diff
@@ -293,6 +293,7 @@ public PermissionDeletedResponse deleteUserFromDeveloper(
             toCreate.setDeveloperCode(developerInfo.getDeveloper().getDeveloperCode());
             toCreate.setName(developerInfo.getDeveloper().getName());
             toCreate.setWebsite(developerInfo.getDeveloper().getWebsite());
+            toCreate.setSelfDeveloper(developerInfo.getDeveloper().getSelfDeveloper());
```
kekey1: I think there should be an additional change in this file... in the split method?

andlar:
I don't think the Self-developer field should go into the upload/confirm workflow yet. It's not in the upload template, which means it won't be in the pending_certified_product_details, and while I could add it to the developer associated with the pending listing, there's no way to edit that developer to set it to be anything other than "No" until after the Listing has been confirmed, and the new Developer has been created. When it has, it'll be created as "No", and can then be edited

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/DeveloperManager.java
```diff
@@ -389,11 +389,21 @@ public DeveloperDTO merge(List<Long> developerIdsToMerge, DeveloperDTO developer
     }
 
     /**
+<<<<<<< HEAD
```
kekey1: Conflict?

andlar: Ugh. Eclipse didn't show it to me, and why would I look for conflicts in documentation that I didn't touch? Sigh

PR #849, opened 2020-01-13
OCD-3226: Change name/number of (b)(3) test functionality

andlar:

PR #870, opened 2020-02-04
feat-flag!: Add updateMacraMeasuresJob to update g1g2 MACRA measures from ACI to PI

drbgfc:
[#OCD-3244]


drbgfc:
A few comments:
1. chpl/chpl-resources/src/main/resources-production/jobs.xml
    - Not sure if I needed/was supposed to add the job here because the file is fairly empty as is and I'm not exactly sure why/the process
2. chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/UpdateMacraMeasuresJob.java 
    - I added the flag as an extra safety precaution. But, if you'd rather I just rely on us waiting until the date to run it, that's fine too, and I can remove
3. I found a potential bug in the system jobs display while testing this. It seems that a job which is run and encounters an unhandled exception is not removed from the list by the backend. Therefore, the UI still displays the job and shows it as 'In Progress' since the date is null. I'm surprised the tests didn't catch this, I thought we did test it. Oh well.
    - Should I look into this as part of this ticket since somewhat related (and for obvious reasons I know that code well). Or, should it be a new ticket?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/MacraMeasureDAO.java
```diff
@@ -68,6 +68,14 @@ public void removeAllByValue(String value) {
         entityManager.flush();
     }
 
+    public void updateSubstringInAllValues(final String oldSubstring, final String newSubstring) {
```
kekey1: The two added DAO methods are pretty specific to this task which is pretty specific to Cures and the way we are using jobs to update data at certain times. We are trying to keep these type of methods out of the codebase, so the technique is to make a dao inner class in your job that has whatever methods you need.  When Cures is done we can delete these types of jobs. Check out ModifyTestFunctionalityJob for an example.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/UpdateMacraMeasuresJob.java
```diff
@@ -0,0 +1,68 @@
...
+    @Autowired
+    private JpaTransactionManager txnMgr;
+
+    @Autowired
+    private FF4j ff4j;
+
+    @Override
+    public void execute(final JobExecutionContext jobContext) throws JobExecutionException {
+        SpringBeanAutowiringSupport.processInjectionBasedOnCurrentContext(this);
+        if (!ff4j.check(FeatureList.EFFECTIVE_RULE_DATE)) {
```
kekey1: I think I did this in one of the jobs I wrote and then somebody said not to but I can't really remember. Most of the Cures jobs don't have this check so it might be better to leave it out for consistency. I don't have a strong feeling either way beyond the consistency argument.

andlar:
We'd also talked about extracting in progress quartz jobs as a separate file that could be imported using the quartz.properties file, and when the PR is ready to be merged into staging, only then adding the file to the regular jobs.xml files. I think that worked fairly well last time. The commit that I did to do that is shown in a closed PR here: https://github.com/chpladmin/chpl-api/pull/849/commits/366170e6d028376c4466eae452abac596e79d7b2

The "merged into staging" commit later shows how that works when it's time to push it for real. The idea was to try to reduce merge conflicts between new jobs

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/UpdateMacraMeasuresJob.java
```diff
@@ -0,0 +1,82 @@
...
+                .forEach(measure -> {
+                    measure.setValue(measure.getValue().replaceAll(oldSubstring, newSubstring));
+                    entityManager.merge(measure);
+                });
+        }
+
+        private List<MacraMeasureEntity> getAllMeasuresWhereValueIsNotNull() {
+            Query query = entityManager
+                    .createQuery("FROM MacraMeasureEntity mme "
+                            + "WHERE mme.value IS NOT NULL", MacraMeasureEntity.class);
```
andlar: Are there any deleted macra measures? Should this check that the deleted flag is set to "false"?

PR #872, opened 2020-02-07
OCD-3217: stop allowing transparency in surveillance

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/surveillance/AddDocumentActionPermissions.java
```diff
@@ -93,6 +98,15 @@ private boolean isNonconformityForRemovedCriteria(SurveillanceNonconformityEntit
                 && nonconformity.getCertificationCriterionEntity().getRemoved().booleanValue();
     }
 
+    private boolean isNonconformityForRemovedRequirement(SurveillanceNonconformityEntity nonconformity) {
+        if (!ff4j.check(FeatureList.EFFECTIVE_RULE_DATE_PLUS_ONE_WEEK)) {
+            return false;
+        }
+
+        return nonconformity != null
+                && nonconformity.getType().equalsIgnoreCase("170.523 (k)(2)");
```
kekey1: I am guilty of not always doing this but can we make this string a constant somewhere? I see it referenced in a number of the files throughout the PR.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/reviewer/NewSurveillanceRemovedCriteriaReviewer.java
```diff
@@ -64,6 +66,18 @@ private void checkRequirementForRemovedCriteria(Surveillance surv, SurveillanceR
         }
     }
 
+    private void checkRequirementForRemovedTransparency(Surveillance surv, SurveillanceRequirement req) {
+        if (req.getType() != null && !StringUtils.isEmpty(req.getType().getName())
+                && req.getType().getName().equalsIgnoreCase(SurveillanceRequirementType.TRANS_DISCLOSURE_REQ)) {
+            String requirement = req.getRequirement();
+            if (requirement != null && requirement.equalsIgnoreCase("170.523 (k)(2)")) {
+                surv.getErrorMessages().add(
+                        msgUtil.getMessage("surveillance.requirementNotAddedForRemovedRequirement",
```
kekey1: Is this error message only supposed to be added for ACBs? I could definitely be missing some larger context here but it looks like it might give an error for all users.

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/surveillance/AddDocumentActionPermissions.java
andlar: Where would you suggest it go? The string is referenced from both domain objects and reviewers, so I don't think it could be in a base class somewhere they both extend from. Do we have a "constants" file somewhere that would make sense to include it?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/reviewer/NewSurveillanceRemovedCriteriaReviewer.java
andlar: This method is called from the `review` method, but only after checks for both a flag and the user being ROLE_ACB. Yes, only ROLE_ACB will get this message; ROLE_ADMIN and ROLE_ONC never get into this method at all

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/surveillance/AddDocumentActionPermissions.java
kekey1: It looks like we already have it in RequirementTypeEnum

kekey1: and another enum NonconformityType

PR #879, opened 2020-02-13
OCD-3222: add BOM to allow for display of extended characters in CSV files

tmy1313:
OCD-3222
feat: Mark downloadable CSVs in a way that matches what Excel is expecting


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/presenter/SurveillanceCsvPresenter.java
```diff
@@ -57,6 +57,7 @@ public SurveillanceCsvPresenter(final Environment env) {
     public void presentAsFile(final File file, final List<CertifiedProductSearchDetails> cpList) {
         try (FileWriter writer = new FileWriter(file);
                 CSVPrinter csvPrinter = new CSVPrinter(writer, CSVFormat.EXCEL)) {
+            writer.write('\ufeff');
```
drbgfc: It might be nice to have a single external shared BOM constant for the '\ufeff' strings

PR #887, opened 2020-02-24
OCD-3230: update CMS ID logic

andlar:


kekey1:
Which commit has the change with "OR (g)(10)" logic? I looked at all the non-build ones and couldn't find it but may have easily missed it.

andlar:
None. I'd apparently already written that code when I updated the 2015 Validator to include code based on the ERD flag. It just couldn't be tested then because the g10 didn't exist

I didn't realize that until after doing all of the refactoring

PR #890, opened 2020-02-28
OCD-3220: Allow surveillance and nonconformities to work with criterion ID instead of number

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/NonconformityTypeStatisticsDTO.java
```diff
@@ -76,12 +80,20 @@ public Date getLastModifiedDate() {
...
 
     public void setLastModifiedUser(Long lastModifiedUser) {
         this.lastModifiedUser = lastModifiedUser;
     }
+
+    public CertificationCriterionDTO getCriterion() {
+        return criterion;
+    }
+
+    public void setCriterion(CertificationCriterionDTO criterion) {
```
andlar: Is this a candidate for refactoring to use the Lombok package/tool?

PR #893, opened 2020-03-03
OCD-3237 feat-flag!: Return errors when attesting to both old and new criteria

drbgfc:
OCD-3237


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/edition2015/InvalidCriteriaCombinationReviewer.java
```diff
@@ -1,68 +1,66 @@
...
-    private Optional<CertificationResult> findCerificationResult(CertifiedProductSearchDetails listing,
+    private Optional<CertificationResult> findCertificationResult(CertifiedProductSearchDetails listing,
             Integer criteriaId) {
         return listing.getCertificationResults().stream()
                 .filter(cr -> cr.getCriterion().getId().equals(Long.valueOf(criteriaId)) && cr.isSuccess())
                 .findFirst();
     }
+
+    private String getTitleForErrorMessage(CertificationCriterion criterion) {
+        return criterion.getTitle().contains(CURES_UPDATE_IN_TITLE)
```
kekey1: I had added a method to the CertificaitonCriterion class called formatCriteriaNumber() wanting to keep the "(Cures Update)" formatting logic in one place. Could that be used here?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/edition2015/InvalidCriteriaCombinationReviewer.java
```diff
@@ -1,68 +1,66 @@
...
     @Override
     public void review(CertifiedProductSearchDetails listing) {
         if (ff4j.check(FeatureList.EFFECTIVE_RULE_DATE)) {
             checkForInvalidCriteriaCombination(listing, criteriaB6Id, criteriaB10Id);
             checkForInvalidCriteriaCombination(listing, criteriaG8Id, criteriaG10Id);
+
+            initializeOldAndNewCriteria();
+            Iterator<Integer> oldCriteriaIdsIter = oldCriteriaIds.iterator();
+            Iterator<Integer> newCriteriaIdsIter = newCriteriaIds.iterator();
+            while (oldCriteriaIdsIter.hasNext() && newCriteriaIdsIter.hasNext()) {
```
andlar: I'm uncomfortable with this setup. This assumes each array has the correct order of paired criteria to check against, and while it looks like the "initialize" method sets that up, it's hard to parse that. What if, instead, you created an array of Tuples/Pairs, maybe using something like https://docs.spring.io/spring-data/data-commons/docs/current/api/org/springframework/data/util/Pair.html, which would actually bind the old & new criteria together in a single object, and then allow those to be iterated over and help together?

kekey1: I agree and was going to comment something similar but your suggestion is a good one and leaves most of the code how it's been written. I was going to suggest not using the IDs at all and just coding a more generic check that says the listing can't attest to 2 criterion of the same number.

andlar: Two of the same number would work, but that wouldn't let us roll the existing "can't have b6 and b10" or "g8 and g10" into the same philosophy

chpl/chpl-service/src/test/java/gov/healthit/chpl/util/UtilTest.java
```diff
@@ -0,0 +1,136 @@
...
+    public void sortCriteria_WithMatchingNumbers_SortCorrectly() {
+        criterionDto.setNumber("170.314 (a)(3)");
+        criterionDto.setTitle("This is a title");
+        criterionDto2.setNumber("170.314 (a)(3)");
+        criterionDto2.setTitle("This is a title (Cures Update)");
+        assertTrue("should sort by title if paragraphs match", Util.sortCriteria(criterionDto, criterionDto2) < 0);
+    }
+
+    @Test
+    public void isCures_CertificationCriterionWithCuresSuffixInTitle_IsNotCures() {
```
andlar:
```suggestion
    public void isCures_CertificationCriterionWithoutCuresSuffixInTitle_IsNotCures() {
```

chpl/chpl-service/src/test/java/gov/healthit/chpl/util/UtilTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Test
+    public void isCures_CertificationCriterionWithCuresSuffixInTitle_IsCures() {
+        criterion.setTitle("170.315(b)(7)" + Util.CURES_SUFFIX);
+        assertTrue("The criterion Should be identified as CURES but was instead identfied as NOT CURES", Util.isCures(criterion));
+    }
+
+    @Test
+    public void isCures_CertificationCriterionDtoWithCuresSuffixInTitle_IsNotCures() {
```
andlar:
```suggestion
    public void isCures_CertificationCriterionDtoWithoutCuresSuffixInTitle_IsNotCures() {
```

drbgfc: Good eye! Thanks.

chpl/chpl-resources/src/main/resources/lookup.properties
```diff
@@ -38,15 +38,24 @@ criterion.170_315_b_9_revised=170
 # 170.315(c)(3) Clinical Quality Measures - Report
 criterion.170_315_c_3=27
 criterion.170_315_c_3_revised=172
+# 170.315(d)(2) Auditable Events and Tamper-Resistance
+criterion.170_315_d_2=30
+criterion.170_315_d_2_revised=183
```
andlar: I feel like this one should be 179, d3_revised will be 180, and d1_revised will be 181

andlar: Could be completely wrong, but these revised ones will be inserted before the revised ones that start with any letter after d, so should have IDs lower than those

drbgfc: Ended up being this when cross-referenced with prod DB and running add job on it - updated as such:
# 170.315(d)(2) Auditable Events and Tamper-Resistance
criterion.170_315_d_2=30
criterion.170_315_d_2_revised=173
# 170.315(d)(3) Audit Report(s)
criterion.170_315_d_3=31
criterion.170_315_d_3_revised=174
# 170.315(d)(10) Auditing Actions on Health Information
criterion.170_315_d_10=38
criterion.170_315_d_10_revised=175

PR #902, opened 2020-03-16
OCD-3284: Add d2,d3, and d10 to v19 upload template

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/AddV19UploadTemplateJob.java
```diff
@@ -51,14 +52,23 @@ public void execute(final JobExecutionContext jobContext) throws JobExecutionExc
                 .findFirst();
 
             if (existingTemplate.isPresent()) {
-                LOGGER.info("Template " + UploadTemplateVersion.EDITION_2015_VERSION_5.getName() + " already exists.");
+                LOGGER.info("Template " + UploadTemplateVersion.EDITION_2015_VERSION_5.getName() + " exists and will be updated.");
+                UploadTemplateVersionDTO newUploadTemplate = new UploadTemplateVersionDTO();
```
andlar: Is it worth moving lines 56-60 to in front of line 54? These are the same lines as in the `else` clause of this if; the only difference is the log lines and the `update`/`insert` line...

PR #904, opened 2020-03-20
OCD-3279: Include "(Cures Update)" in all error/warning messages with cures criteria.

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/edition2015/duplicate/TestData2015DuplicateReviewer.java
```diff
@@ -37,7 +38,7 @@ public void review(final PendingCertifiedProductDTO listing, final PendingCertif
         if (testDataDuplicateResults.duplicatesExist()) {
             listing.getWarningMessages().addAll(
                     getWarnings(testDataDuplicateResults.getDuplicateList(),
-                            certificationResult.getCriterion().getNumber()));
+                            Util.formatCriteriaNumber(certificationResult.getCriterion())));
```
tmy1313: I know `Util.formatCriteriaNumber(...)` is all over the place and not necessarily part of this change, I do not like this method being in the `Util` class.  
http://ralin.io/blog/oop-anti-patterns-utility-or-helper-classes.html

Is there a way we can get this to be a part of the `CertificationResult`, `CertificationResultDTO` classes?  I see two different methods in `Util` that do the same thing to strings in different objects.

chpl/chpl-service/src/test/java/old/gov/healthit/chpl/validation/listing/reviewer/ChplNumberReviewerTest.java
```diff
@@ -60,6 +61,9 @@
     @Autowired
     private MessageSource messageSource;
 
+    @Spy
```
tmy1313: Are we trying to maintain the "old" tests.  I think this is a good reason for us to remove them from the project, but this might be a discussion for the Dev team.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/pendingListing/reviewer/edition2015/duplicate/TestData2015DuplicateReviewer.java
kekey1: I originally added it in CertificationCriterion domain object. Then Dan moved it into Util to have one method for each of those two types. I can refactor it back...

chpl/chpl-service/src/test/java/old/gov/healthit/chpl/validation/listing/reviewer/ChplNumberReviewerTest.java
kekey1: To me this test is close-ish to working as mock-only so maybe the thing that I should do is actually make it work like that and move it into the new tests.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/surveillance/reviewer/ComparisonReviewer.java
```diff
@@ -1,8 +1,19 @@
 package gov.healthit.chpl.validation.surveillance.reviewer;
 
 import gov.healthit.chpl.domain.surveillance.Surveillance;
+import gov.healthit.chpl.domain.surveillance.SurveillanceNonconformity;
+import gov.healthit.chpl.domain.surveillance.SurveillanceRequirement;
+import gov.healthit.chpl.util.SurveillanceUtil;
 
-public interface ComparisonReviewer {
+public abstract class ComparisonReviewer {
```
drbgfc: Maybe you could have used a default method instead of converting to an abstract class. Nothing wrong with this way, just a thought.

PR #905, opened 2020-03-20
OCD-3252: Support "Cures Update" per Listing in API

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
```diff
@@ -0,0 +1,129 @@
...
+        public static final String A_1 = "criterion.170_315_a_1";
+        public static final String A_2 = "criterion.170_315_a_2";
+        public static final String A_3 = "criterion.170_315_a_3";
+        public static final String A_4 = "criterion.170_315_a_4";
+        public static final String A_5 = "criterion.170_315_a_5";
+        public static final String A_6 = "criterion.170_315_a_6";
+        public static final String A_7 = "criterion.170_315_a_7";
+        public static final String A_8 = "criterion.170_315_a_8";
+        public static final String A_9 = "criterion.170_315_a_9";
+        public static final String A_10 = "criterion.170_315_a_1";
```
kekey1:
```suggestion
        public static final String A_10 = "criterion.170_315_a_10";
```

andlar:
> I'd like to see the exceptions thrown come from errors.properties eventually but am fine with doing it at some later point next time we touch this code.

All of these exceptions are placeholders that'll need to be replaced 24 months from now. Probably around the same time the `isPast24Months()` and `isPast36Months()` methods return anything other than `false`

PR #910, opened 2020-04-01
OCD-3290 - Change the order of display of revised criteria so "Cures Update" is first

tmy1313:
OCD-3290


chpl/chpl-service/src/main/java/gov/healthit/chpl/util/Util.java
```diff
@@ -112,64 +105,18 @@ public static Date getNewDate(final Date orig) {
...
-        if (s1 == null && s2 == null) {
-            return 0;
-        }
-        if (s1 == null) {
-            return -1;
+    private static Integer getCertificationResultSortIndex(String criteriaNumber) {
+        Integer index = certificationResultSortOrder().indexOf(criteriaNumber);
+        if (index.equals(-1)) {
+            // This is case when the criteria number is not in the array, just make it last...
+            index = 10000;
```
kekey1, line 117: This is a small thing and I realize the code will work as written but could you make it like Integer.MAX_VALUE instead of a randomly chosen large number?

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/Util.java
```diff
@@ -195,4 +142,191 @@ public static String formatCriteriaNumber(CertificationCriterionDTO criterion) {
         }
         return result;
     }
+
+    private static List<String> certificationResultSortOrder() {
```
kekey1, line 146: This isn't my favorite... now if we were to add a new criteria we have to mess with the lookup.properties file (which is basically a copy of our lookup tables in the database...) and have to also remember to change this list. I'm not sure what change to suggest but if I think of something I will post it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/CertifiedProductDownloadableResourceCreatorJob.java
```diff
@@ -106,8 +112,8 @@ private void initializeWritingToFiles(final CertifiedProductXmlPresenter xmlPres
         List<CertificationCriterionDTO> criteria = getCriteriaDao().findByCertificationEditionYear(edition)
                 .stream()
                 .filter(cr -> !cr.getRemoved())
-                .sorted((crA, crB) -> Util.sortCriteria(crA, crB))
-                .collect(Collectors.<CertificationCriterionDTO>toList());
+                .sorted((crA, crB) -> criterionService.sortCriteria(crA, crB))
+                .collect(Collectors.<CertificationCriterionDTO> toList());
```
andlar, line 116: I've see these show up before. Is this your auto-formatter doing it? I ask because when I set these lines like this I get a checkstyle error about `'>' is followed by whitespace.`

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
```diff
@@ -46,6 +56,88 @@ public CertificationCriterion get(String certificationCriterionDescriptor) {
         }
     }
 
+    public int sortCriteria(CertificationCriterionDTO c1, CertificationCriterionDTO c2) {
+        String valueA = formatCriteriaNumber(c1);
+        String valueB = formatCriteriaNumber(c2);
+        return getCertificationResultSortIndex(valueA) - getCertificationResultSortIndex(valueB);
+    }
+
+    public boolean isCures(CertificationCriterion criterion) {
```
andlar, line 65: Can we rename this function? It would not report d12, d13, b10, or g10 as "cures = true", even though those are also criteria considered Cures. Maybe "hasCuresUpdateTitle" or something

I worry someone trying to filter criteria to those that are actually cures would use this method and get incorrect results without realizing it

PR #913, opened 2020-04-03
OCD-3249 feat-flag!: Require d12 and d13 when editing certain 2015 listings in certain scenarios

drbgfc:
* Require d12 and d13 when editing a listing and all of the following are true
  * The ERD flag is on
  * The listing is 2015 Edition
  * The listing has an Active or Suspended Status
  * One or more critera were aded during the edit
  * The criteria or criterion added were/are from the 'required' list OR criteria from the 'required' list existed prior to the added criteria
  * The listing does not attest to d12 or d13 before saving

OCD-3249


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationCriterion.java
```diff
@@ -12,12 +12,14 @@
...
 import lombok.AllArgsConstructor;
 import lombok.Builder;
+import lombok.EqualsAndHashCode;
 
 @XmlType(namespace = "http://chpl.healthit.gov/listings")
 @XmlAccessorType(XmlAccessType.FIELD)
 @JsonIgnoreProperties(ignoreUnknown = true)
 @Builder
 @AllArgsConstructor
+@EqualsAndHashCode
```
andlar, line 22: According to https://projectlombok.org/features/Builder this annotation can't exist on a class that has `@Builder` as well.

First paragraph in "Overview" after the list

drbgfc: Updated with latest commit, thanks.

PR #915, opened 2020-04-07
OCD-3302 - Update Listings to be marked as Cures Update at ERD

tmy1313:
OCD-3302


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/AddCuresUpdateToListingsJob.java
```diff
@@ -0,0 +1,127 @@
...
+
+        LOGGER.info("********* Completed the Update Listing To Cures Updated  job. *********");
+
+    }
+
+    private void processListing(CertifiedProductDetailsDTO cp) {
+        try {
+            LOGGER.info("Retrieving listing information for  (" + cp.getId() + ") " + cp.getChplProductNumber());
+            CertifiedProductSearchDetails listing = certifiedProductDetailsManager.getCertifiedProductDetails(cp.getId());
+            if (curesUpdateService.isCuresUpdate(listing)) {
```
andlar, line 73: This job needs to add a line of "cures update = false" to the cures_update_event table for listings that don't meet the curesUpdate requirement, as well as the "cures update = true" that it's currently doing

It _shouldn't_ add activity for listings that are "cures update = false" though; only for ones that are getting set as true

chpl/chpl-resources/src/main/resources/ocd-3302.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<job-scheduling-data
+    xmlns="http://www.quartz-scheduler.org/xml/JobSchedulingData"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.quartz-scheduler.org/xml/JobSchedulingData http://www.quartz-scheduler.org/xml/job_scheduling_data_2_0.xsd"
+    version="2.0">
+    <schedule>
+        <!--  Update 2015 listings as Cures Updated, where appropriate -->
+        <job>
+            <name>addCuresUpdatedToListings</name>
```
kekey1, line 10: This is a small thing but the job is "addCuresUpdate**d**" and the logger is "addCuresUpdate" . Not sure which one you want to go with but I would like them to be the same!

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/AddCuresUpdateToListingsJob.java
```diff
@@ -0,0 +1,127 @@
...
+        txTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+        txTemplate.execute(new TransactionCallbackWithoutResult() {
+
+            @Override
+            // can't use the manager method because we don't want to record activity
+            // so wrapping the dao call in a transaction
+            protected void doInTransactionWithoutResult(TransactionStatus status) {
+                try {
+                    addCuresUpdateEvent(origListing.getId(), true, specialProperties.getEffectiveRuleDate());
+                    addCuresUpdateActivity(origListing,
```
kekey1, line 96: This is a little confusing because the comment above specifically says you are doing this so that you don't add activity but then you are calling a method that says it's adding activity.

PR #921, opened 2020-04-10
OCD-3288 - ONC-ACB should only be able to add or remove criteria if listing is in "Active" status

tmy1313:
OCD-3288


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/ListingStatusAndUserRoleReviewer.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Autowired
+    public ListingStatusAndUserRoleReviewer(ResourcePermissions resourcePermissions, ErrorMessageUtil messages) {
+        this.resourcePermissions = resourcePermissions;
+        this.messages = messages;
+    }
+
+    @Override
+    public void review(CertifiedProductSearchDetails existingListing, CertifiedProductSearchDetails updatedListing) {
+        if (!resourcePermissions.isUserRoleAdmin() && !resourcePermissions.isUserRoleOnc()) {
```
andlar, line 30: We want to try to get rid of nested if statements, in our pursuit of clean code, right? My first thought was to just collapse all of these ifs into a single if, with some && and || as necessary, but then ran across this: https://wpshout.com/unconditionally-refactoring-nested-statements-cleaner-code/

What if, instead, here the code was structured something like

```
public void...
if (isUserRoleAdmin or isUserRoleONC) {
    return
}

if (isCurrentListingActive && isNewListingActive) {
   return
}

if (haveCriteriaBeenAdded || haveCriteriaBeenRemoved) {
   add errormessage
}
```

No nesting, immediate exits from the method when we know a given check means we don't need to be here, simple to read

Thoughts?

chpl/chpl-service/src/test/java/gov/healthit/chpl/validation/listing/reviewer/ListingStatusAndUserRoleReviewerTest.java
```diff
@@ -0,0 +1,411 @@
...
+                .build();
+        updatedListing.setErrorMessages(new HashSet<String>());
+
+        reviewer.review(origListing, updatedListing);
+
+        assertEquals(1, updatedListing.getErrorMessages().size());
+    }
+
+    @Test
+    public void review_UserIsAcbAndListingUpdatedStatusFromRetiredToActiveAndCriteriaDeleted_ErrorMessageAdded()
```
andlar, line 369: Can we have one test of the form `review_UserIsAcbAndListingIsUpdatedWithAChangedCriteria_ErrorMessageAdded` to cover the state where a listing in a non-active state has a criteria added and removed in the same event? I had to read the reviewer to verify that case was being handled, but it'd be nice to see in a test if, for instance, a listing added a1 and removed b1 in the same activity

chpl/chpl-service/src/test/java/gov/healthit/chpl/validation/listing/reviewer/ListingStatusAndUserRoleReviewerTest.java
```diff
@@ -12,8 +12,8 @@
 
 import gov.healthit.chpl.domain.CertificationCriterion;
 import gov.healthit.chpl.domain.CertificationResult;
-import gov.healthit.chpl.domain.CertificationStatus;
 import gov.healthit.chpl.domain.CertificationStatusEvent;
+import gov.healthit.chpl.domain.CertificationStatusProvider;
```
andlar, line 16: We're effectively making a "mock" domain object, right? Maybe worth passing in the ID & effective date as well, and then the Builder can make one with all of the properties? As it stands, the builders below are still passing in "magic" variables in the form of the ID and effective date. I know they're not flagged as "magic", but we know they actually are. Thinking of the class as something that we could call out as a mock as well; wouldn't want it to get confused with "real" code; should it be a "CertificationStatusMock" as opposed to a "CertificationStatusProvider"?

tmy1313: But these are not "mock" objects.  They are real object, just not populated by the DB.  They have all of the functionality of a real `CertificationStatus` object.  If we were to pass in an `id` and `effectiveDate` this would be producing `CertificationStatusEvent` objects.  The `CertificationStatusProvider` was designed to provide access to "static" list of statuses - basic lists of data.  I don't think `CertificationStatusEvent(s)` are static, since they are specific and different for every certified product.  That is why I wrote the helper method `getCertificationStatusEvent`.

PR #924, opened 2020-04-15
OCD-3302: remove the ocd-3302.xml job definition, moved to jobs.xml

tmy1313:
OCD-3302


chpl/chpl-resources/src/main/resources/quartz.properties
```diff
@@ -7,7 +7,7 @@ org.quartz.jobStore.tablePrefix = quartz.QRTZ_
 org.quartz.jobStore.isClustered = true
 
 org.quartz.plugin.jobInitializer.class = org.quartz.plugins.xml.XMLSchedulingDataProcessorPlugin
-org.quartz.plugin.jobInitializer.fileNames = jobs.xml,ocd-3122.xml,ocd-3302.xml
+org.quartz.plugin.jobInitializer.fileNames = jobs.xml,ocd-3122.xml
```
andlar, line 10: Is ocd-3122.xml still around? If so, it should be combined into jobs.xml. If not it shouldn't be in this file either

PR #926, opened 2020-04-17
OCD-3307 - Add 2015 Cures Edition metrics to summary statistics report

tmy1313:
OCD-3307


andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertificationResultDAO.java
```diff
@@ -1021,12 +1002,27 @@ private CertificationResultTestProcedureEntity getCertificationResultTestProcedu
...
+    private List<CertificationResultTestProcedureEntity> getTestProcedureEntitiesForListing(Long listingId) {
+
+        String hql = "SELECT tp "
+                + "FROM CertificationResultTestProcedureEntity tp "
+                + "LEFT OUTER JOIN FETCH tp.testProcedure "
+                + "WHERE (NOT tp.deleted = true) "
+                + "AND tp.certificationResultId IN "
+                + "(SELECT cr.id "
+                + "FROM CertificationResultEntity cr "
+                + "WHERE cr.certifiedProductId = :listingId)";
```
kekey1: Do you need an "and cr.deleted = false" type of statement as part of the "IN" clause? Maybe not but I can't tell just from thinking about it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/AsynchronousSummaryStatistics.java
```diff
@@ -540,7 +546,182 @@
...
+    public Future<Long> getAllListingsCountWithCuresUpdatedWithAlternativeTestMethods(
+            CertifiedProductDAO certifiedProductDAO, CertificationResultDAO certificationResultDAO) {
+        return new AsyncResult<Long>(certifiedProductDAO.findCuresUpdatedListings().stream()
+                .filter(cp -> doesListingHaveAlternativeTestMethod(cp.getId(), certificationResultDAO))
+                .count());
+    }
+
+    private boolean doesListingHaveAlternativeTestMethod(Long listingId, CertificationResultDAO certificationResultDAO) {
+        return certificationResultDAO.getTestProceduresForListing(listingId).stream()
+                .filter(crtp -> !crtp.getTestProcedure().getName().equals("ONC Test Method"))
```
kekey1: For checkstyle purposes could the string "ONC Test Method" be a constant?

kekey1 (changes requested):
My two comments may not be required changes (not sure about the HQL one) so if you think it's all correct how it is I won't hold up the PR for that.. let me know.

andlar (dismissed):

kekey1 (dismissed):

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -142,7 +142,7 @@
             <job-data-map>
                 <entry>
                     <key>generateCsvFile</key>
-                    <value>false</value>
+                    <value>true</value>
```
andlar: Is this necessary? I think I'd rather not do it by default on local environments, which is the only place this job runs

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertifiedProductDAO.java
```diff
@@ -1,66 +1,672 @@
...
+import gov.healthit.chpl.entity.listing.CertifiedProductEntity;
+import gov.healthit.chpl.entity.listing.CertifiedProductSummaryEntity;
 import gov.healthit.chpl.exception.EntityCreationException;
 import gov.healthit.chpl.exception.EntityRetrievalException;
 import gov.healthit.chpl.scheduler.job.urlStatus.UrlType;
+import gov.healthit.chpl.util.AuthUtil;
+
+@Repository(value = "certifiedProductDAO")
+public class CertifiedProductDAO extends BaseDAOImpl {
+    private static final Logger LOGGER = LogManager.getLogger(CertifiedProductDAO.class);
```
andlar: `@Log4j2` ?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/SummaryStatisticsEmailJob.java
```diff
@@ -119,13 +103,18 @@ private Statistics getStatistics(SummaryStatisticsEntity summaryStatistics)
...
 
         emailMessage.append(createMessageHeader(endDate));
-        emailMessage.append(createUniqueDeveloperSection(stats));
-        emailMessage.append(createUniqueProductSection(stats));
-        emailMessage.append(createListingSection(stats));
-        emailMessage.append(createSurveillanceSection(stats));
-        emailMessage.append(createNonconformitySection(stats));
+        emailMessage.append(developerStatisticsSectionCreator.build(stats, activeAcbs, ff4j));
+        emailMessage.append(productStatisticsSectionCreator.build(stats, activeAcbs, ff4j));
+        //emailMessage.append(listingStatisticsSectionCreator.build(stats, activeAcbs, ff4j));
```
andlar: Should this be here?

ready-for-integration.sh
```diff
@@ -0,0 +1 @@
+../../resources/scripts/ready-for-integration.sh
```
andlar: This should be ignored in the .gitignore file. It is on the staging branch now

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/EditionCriteria.java
```diff
@@ -0,0 +1,5 @@
+package gov.healthit.chpl.scheduler.job.summarystatistics;
+
+public enum EditionCriteria {
+    ALL, EDITION_2011, EDITION_2014, EDITION_2015_CURES, EDITION_2015_NON_CURES, EDITION_2015_CURES_AND_NON_CURES
```
andlar: Worth putting these on multiple lines so when we (inevitably) need to change them it's easier to read the line diff?

andlar (commented):
There's a PR for OCD-3275 that has a "maybe going to be standard" email signature thing in it. Might be worth checking to see if that should be put in this email as well. I know it'll block this until that one gets merged, but that one is ready to merge now, so probably won't block for long

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/CsvStatistics.java
```diff
@@ -0,0 +1,57 @@
+package gov.healthit.chpl.scheduler.job.summarystatistics;
```
kekey1: EmailStatistic and EmailStatistics are in the domain package and CSVStatistics is in the job/summarystatistics package. Just based on the class names I would think they should probably all be in the job/summarystatistics package because "Email" is really specific to what the job does (as is CSV). Thoughts on the naming or location of these?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/NonConformityStatisticsSectionCreator.java
```diff
@@ -0,0 +1,51 @@
+package gov.healthit.chpl.scheduler.job.summarystatistics;
```
kekey1: Just another naming thing for this class and at least one other.. I think we usually write "Nonconformity" as a single word without camel case. At least it is that way in most of the surveillance code. Now I wonder if that is wrong since I sometimes see it written with a dash (Non-conformity)... could we figure out which way is right/which way we want to have it in the code and move towards that?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/StatisticsMassager.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private Boolean isAcbInStatistics(CertificationBodyDTO acb, List<EmailCertificationBodyStatistic> stats) {
+        return stats.stream()
+                .filter(stat -> stat.getAcbName().equals(acb.getName()))
+                .findAny()
+                .isPresent();
+    }
+
+    private EmailCertificationBodyStatistic getNewCertifiedBodyStatistic(String acbName) {
```
kekey1: Can you rename to getNewCertificationBodyStatistic (instead of CertifiedBody)? I think we use CertificationBody more consistently in the code although maybe not everywhere.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/SummaryStatisticsEmailJob.java
```diff
@@ -59,14 +46,11 @@
     @Autowired
     private Environment env;
 
+    @Autowired
+    private FF4j ff4j;
```
kekey1: I don't see this used anywhere

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertifiedProductDAO.java
```diff
@@ -1,66 +1,671 @@
 package gov.healthit.chpl.dao;
 
+import java.util.ArrayList;
```
kekey1: FYI I have a PR for OCD-3275 ("Reviewed" status) that also removes this Impl. It would be good to make sure we know what other changes were made to the file in each case so we don't lose them because this will be a conflict for sure (I would think if you pushed this to DEV it has already become an issue there?).

kekey1 (commented):
Most of my comments are actually just comments and don't require changes but I want to discuss a couple of them before settling on Approval.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/NonConformityStatisticsSectionCreator.java
andlar: I like this suggestion, but think we also need to look at the UI to see what the "business domain" usage has been. A quick check suggests that all of "Nonconformity", "Non-conformity", and "Non-Conformity" are in use (though the middle one only once) and with "Non-Conformity" the slight favorite. Suggest we bring it to the team and get an answer from the PO, then move towards whatever answer that is everywhere we can

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #931, opened 2020-04-21
OCD-3341 - ACBs receive error when editing a listing with removed MACRA measures

tmy1313:
OCD-3341


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/MacraMeasure.java
```diff
@@ -35,15 +39,15 @@
     private String abbreviation;
 
     /**
-     * The name of the measure that was successfully tested. For example,
-     * "Computerized Provider Order Entry - Medications: Eligible Hospital/Critical"
+     * The name of the measure that was successfully tested. For example, "Computerized Provider Order Entry -
```
kekey1, line 42: Not sure above the two changes in the comments here. I assume they were made by an auto-formatter. They get put as comments into the XSD file... I know it won't change it in any meaningful way but we don't really need/want it to change at all right?

PR #947, opened 2020-05-05
OCD-3350 - Update Quartz jobs to be non-blocking - BrokenSurveillanceRulesCreatorJob

tmy1313:
OCD-3350


tmy1313:
Sorry the diff looks so bad - might just want to review the updated file without the diff.

PR #948, opened 2020-05-08
OCD-3251: Allow developer split without considering retired listings

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/ActionPermissions.java
```diff
@@ -60,18 +61,20 @@ public boolean isDeveloperValidForCurrentUser(final Long developerId) {
...
 
     @Transactional(readOnly = true)
-    public boolean doesCurrentUserHaveAccessToAllOfDevelopersListings(final Long developerId) {
+    public boolean doesCurrentUserHaveAccessToAllOfDevelopersListings(Long developerId,
+            List<CertificationStatusType> listingStatuses) {
         List<CertifiedProductDetailsDTO> cpDtos = certifiedProductDAO.findByDeveloperId(developerId);
         for (CertifiedProductDetailsDTO cpDto : cpDtos) {
-            if (!isAcbValidForCurrentUser(cpDto.getCertificationBodyId())) {
+            if (isInStatuses(cpDto.getCertificationStatusName(), listingStatuses)
+                    && !isAcbValidForCurrentUser(cpDto.getCertificationBodyId())) {
```
andlar, line 69: I wonder if this entire construction could be switched to a stream style thing that would look a lot like the "private isInStatuses method below. Not necessary, but might simplify the logic

PR #957, opened 2020-05-26
OCD-3360 - feat!: Restrict use of old Criteria based on Certification Date & ICS

andlar:


chpl/chpl-resources/src/main/resources/environment.properties
```diff
@@ -22,7 +22,7 @@ downloadFolderPath=SECRET
 keyLocation=SECRET
 uploadErrorEmailRecipients=SECRET
 #format MM/DD/YYYY
-cures.ruleEffectiveDate=07/01/2020
+cures.ruleEffectiveDate=03/03/2020
```
tmy1313, line 25: I'm not sure we want to make this change.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/edition2015/OldCriteriaWithoutIcsReviewer.java
```diff
@@ -0,0 +1,110 @@
...
+        String dateAsString = env.getProperty(date);
+        SimpleDateFormat sdf = new SimpleDateFormat("MM/dd/yyyy");
+        return sdf.parse(dateAsString).getTime();
+    }
+
+    private boolean hasRelevantCriteria(CertifiedProductSearchDetails listing, String criteria) {
+        return listing.getCertificationResults().stream()
+                .anyMatch(result -> result.getCriterion().getNumber().equals(criteria)
+                        && result.isSuccess()
+                        && !CertificationCriterionService.hasCuresInTitle(result.getCriterion()));
```
kekey1, line 100: I guess since the ticket is in final code review you don't have to/shouldn't change this but... I don't like checking against the criterion number and title together. We got bit once by using criterion numbers everywhere and spent a lot of time changing it all to use the IDs. I think you could use the criterion IDs from whatever lookup file we have and check for ID equality rather than number and title.

PR #962, opened 2020-06-05
OCD-3385 - Adding TF to old c3 after ERD+7d shouldn't save the TF, as ONC-ACB

tmy1313:
fix!: Check that the current user has the correct roles for restricted TFs
OCD-3385


kekey1 (dismissed):

andlar (changes requested):
There are unit tests that cover this file. If the tests didn't find this issue can they be updated to have a test that would have failed without this change?

Also, there's a non-pending Listing Reviewer of the same type; does it have the same issue?

kekey1:
> There are unit tests that cover this file. If the tests didn't find this issue can they be updated to have a test that would have failed without this change?
> 
> Also, there's a non-pending Listing Reviewer of the same type; does it have the same issue?

I totally missed this but good points. The unit test compile actually fails so at least that needs an update!

kekey1 (approved):

andlar (approved):

PR #963, opened 2020-06-08
OCD-3275: Split developer action as quartz job

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductManager.java
```diff
@@ -216,11 +216,11 @@
     private CuresUpdateService curesUpdateService;
     private FF4j ff4j;
 
-    private static final int PROD_CODE_LOC = 4;
-    private static final int VER_CODE_LOC = 5;
-    private static final int ICS_CODE_LOC = 6;
-    private static final int SW_CODE_LOC = 7;
-    private static final int DATE_CODE_LOC = 8;
+    private static int PROD_CODE_LOC = 4;
```
andlar: I think these should still be `final`

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/SchedulerManager.java
```diff
@@ -52,24 +52,25 @@
 @Service
 public class SchedulerManager extends SecuredManager {
 
-    private static final String AUTHORITY_DELIMITER = ";";
-    private static final String DATA_DELIMITER = "\u263A";
-    public static final String CHPL_JOBS_KEY = "chplJobs";
-    public static final String SYSTEM_JOBS_KEY = "systemJobs";
+    private static String AUTHORITY_DELIMITER = ";";
```
andlar: These are probably also `final`

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/ProductManager.java
```diff
@@ -37,7 +37,7 @@
 
 @Service
 public class ProductManager extends SecuredManager {
-    private static final Logger LOGGER = LogManager.getLogger(ProductManager.class);
+    private static Logger LOGGER = LogManager.getLogger(ProductManager.class);
```
andlar: Should either be final or replaced with `@Log4j2`

andlar (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertifiedProductDAO.java
```diff
@@ -1,66 +1,660 @@
...
 import gov.healthit.chpl.exception.EntityCreationException;
 import gov.healthit.chpl.exception.EntityRetrievalException;
 import gov.healthit.chpl.scheduler.job.urlStatus.UrlType;
+import gov.healthit.chpl.util.AuthUtil;
+import gov.healthit.chpl.util.ErrorMessageUtil;
+import lombok.NoArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+@NoArgsConstructor
```
tmy1313: Should this have a no args constructor?  I think this is always instantiated by Spring and always uses the Autowired constructor.

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/rules/developer/DeveloperActiveStatusValidation.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.logging.log4j.Logger;
+
+import gov.healthit.chpl.dto.DeveloperDTO;
+import gov.healthit.chpl.dto.DeveloperStatusEventDTO;
+import gov.healthit.chpl.entity.developer.DeveloperStatusType;
+import gov.healthit.chpl.manager.rules.ValidationRule;
+import gov.healthit.chpl.util.ErrorMessageUtil;
+
+public class DeveloperActiveStatusValidation extends ValidationRule<DeveloperValidationContext> {
+    private static final Logger LOGGER = LogManager.getLogger(DeveloperActiveStatusValidation.class);
```
tmy1313: @Log4j2?

tmy1313 (changes requested):

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/SplitDeveloperJob.java
```diff
@@ -288,6 +297,19 @@ private String createHtmlEmailBodyFailure(DeveloperDTO oldDeveloper, DeveloperDT
                 oldDeveloper.getName(),
                 newDeveloper.getName(),
                 ex.getMessage());
+        htmlMessage += createHtmlEmailFooter();
         return htmlMessage;
     }
+
+    private String createHtmlEmailFooter() {
```
andlar: I think this footer is likely going to end up in emails coming from the system a lot going forward. Is there an email-util file somewhere that this could be moved to so we don't duplicate the effort?

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/EmailBuilder.java
```diff
@@ -108,6 +109,18 @@ public EmailBuilder htmlMessage(String val) {
...
 
+    public EmailBuilder htmlFooter() {
+        htmlFooter = String.format("<p>"
+                + "If there are any questions about this process, please visit the "
+                + "<a href=\"%s\">"
+                + "ONC-ACB & ONC-ATL Portal</a> to submit a ticket."
+                + "</p>"
+                + "<p>Thank you!</p>"
+                + "ONC CHPL Team",
+                env.getProperty("splitDeveloper.help.url"));
```
andlar: Should this property be moved out of the `splitDeveloper` namespace?

chpl/chpl-resources/src/main/resources/email.properties
```diff
@@ -5,6 +5,8 @@
 # * key/value pair in [Tomcat]/conf/email-override.properties file.
 #########################################################################################
 
+#URL to ONC website used in default HTML footer
+footer.url=https://inquiry.healthit.gov/support/servicedesk/customer/user/login?destination=site%2Fforum
```
andlar: I went down a rabbit hole on this one

```suggestion
footer.url=https://inquiry.healthit.gov/support/servicedesk/customer/user/login?destination=site/forum
```

In the RFC for URIs/URLs the query section allows `/` and `?` as valid characters and suggests that it's better to not URL encode those characters: https://tools.ietf.org/html/rfc3986#section-3.4 Also, this URL does actually work :)

tmy1313 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/SplitDeveloperJob.java
```diff
@@ -249,6 +249,14 @@ private void sendEmail(String recipientEmail, String subject, String htmlMessage
                 .sendEmail();
     }
 
+    private String getSubject(boolean success) {
+        if (success) {
+            return env.getProperty("splitDeveloper.success.emailSubject");
+        } else {
```
drbgfc, line 256: Technically the else block is not needed. Just my preference to have less code when more code doesn't tell one anything but you don't need to change on my accord. The bytecode will look the same so not a performance issue...
```
        if (success) {
            return env.getProperty("splitDeveloper.success.emailSubject");
        }
        return env.getProperty("splitDeveloper.failed.emailSubject");
```

You could actually go even further and make this ternary, but that's not objectively more clear, just shorter (one liner). And doesn't offer any (pre-compiled) logic change from the current state. So I prefer the elmination of the else block. To each their own.

If this entire comment is useless, please let me know :-)

andlar (approved):

drbgfc (approved):

PR #975, opened 2020-06-24
API Automation Test

palakpatel311:


Plow74 (dismissed):

e2e/acb-controller.postman_collection.json
```diff
@@ -0,0 +1,139 @@
+{
+	"info": {
```
kekey1: It looks like this JSON file and one other use mostly tabs but the package.json file uses spaces for whitespace. Throughout the rest of the code I think we've pretty much standardized on using spaces instead of tabs - can you make these all use spaces too? It is likely a setting you can turn on in your IDE to do it automatically. I also have something on in Eclipse that shows me the whitespace characters so I can easily see which one I've used - might be helpful to you as well.

kekey1 (changes requested):

kekey1:
I checked out this code to look at it and I noticed that as I've got everything imported into Eclipse none of the stuff in the "e2e" folder appears... it's all one level above the parent pom which is where we import from. Would it be possible to 1) teach me how to see it in eclipse along with the java code the way I have it currently, or 2) move the e2e folder and any other files i might need to edit if I were writing tests down into the chpl/ directory so that I can navigate to them within Eclipse the way I currently have the code imported? Not sure which is the right solution exactly but one way or the other I should be able to see this code along with all the other chpl-api code.

kekey1 (dismissed):

tmy1313 (dismissed):

ashwinimore (approved):

athomas-git (dismissed):

chpl/chpl-api/e2e/collections/acb-controller.postman_collection.json
```diff
@@ -0,0 +1,142 @@
...
+            "event": [
+                {
+                    "listen": "test",
+                    "script": {
+                        "id": "0127e346-3602-4ef0-9987-362dd5003565",
+                        "exec": [
+                            "pm.test(\"Status code is 200\", function () {",
+                            "    pm.response.to.have.status(200);",
+                            "});",
+                            "pm.test(\"Acb name as Drummond Group should return\", function () {",
```
ashwinimore, line 79: Make test description meaningful. The purpose of this test is to verify acb id 4 returns Drummond Group. When the test is run test description reads 'Acb name as drummond group should return' which doesn't convey test purpose.  
On other note, I think this test does not add as much value (acb code verification, other test verifies acbs exist) and is ok to remove unless the team sees value in it

ashwinimore (changes requested):

.gitignore
```diff
@@ -274,3 +274,5 @@ ready-for-test.sh
 reset-qa.sh
 reset-local.sh
 ready-for-integration.sh
+e2e/env/*.postman_environment.json
```
andlar: I don't think this .gitignore line works. I have a file in here that git shows as a new file but that I think needs to be ignored

andlar (dismissed):

chpl/chpl-api/e2e/collections/acb-controller.postman_collection.json
palakpatel311: This test will provide skeleton to add more test to check details of Drummond group. Since this is POC, this ACB controller tests will evolve in future once this POC is approved.

.gitignore
palakpatel311: Pushed a fix

ashwinimore (approved):

andlar (approved):

PR #992, opened 2020-07-22
OCD-3325: Add /activity/metadata/beta/users endpoint for paginated user activity

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/activity/GetUserMaintenanceMetadataActionPermissions.java
```diff
@@ -33,12 +33,11 @@ public GetUserMaintenanceMetadataActionPermissions(final UserCertificationBodyMa
...
-            return false;
-        } else {
-            return true;
-        }
-
+        return getResourcePermissions().isUserRoleAdmin()
+                || getResourcePermissions().isUserRoleOnc()
+                || getResourcePermissions().isUserRoleAcbAdmin()
+                || getResourcePermissions().isUserRoleAtlAdmin()
+                || getResourcePermissions().isUserRoleCmsStaff();
```
andlar, line 40: The ResourcePermissions.java file has stuff about Developer Users; does this Permissions need something about them too?

kekey1: I don't think so? The ResourcePermissions code is getting a list of all the users the logged-in user has access to. This is saying whether the logged-in user is allowed to see user activity. It looked like ROLE_DEVELOPER is not allowed to see any user activity (although they do have access to a subset of users).

andlar (approved):

Plow74 (approved):

PR #997, opened 2020-07-31
OCD-3395: Developer Access Report

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/ApiKeyDTO.java
```diff
@@ -4,131 +4,36 @@
...
-    public Boolean getWhitelisted() {
-        return whitelisted;
-    }
-
-    public void setWhitelisted(final Boolean whitelisted) {
-        this.whitelisted = whitelisted;
-    }
-
-    public Date getCreationDate() {
-        return Util.getNewDate(creationDate);
```
andlar, line 88: I think this line might need to be restored. Not sure exactly what the function does, or if it's actually needed, but using the Lombok thing would change functionality. Same thing with the other dates

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/ApiKeyEntity.java
```diff
@@ -59,93 +60,4 @@
...
-    public Boolean getWhitelisted() {
-        return whitelisted;
-    }
-
-    public void setWhitelisted(final Boolean whitelisted) {
-        this.whitelisted = whitelisted;
-    }
-
-    public Date getCreationDate() {
-        return Util.getNewDate(creationDate);
```
andlar: Similar thing to the DTO date thing

andlar (changes requested):

andlar (dismissed):

tmy1313 (dismissed):

andlar (approved):

tmy1313 (approved):

PR #998, opened 2020-08-03
OCD-3432: Add curesUpdate field to XML download; valid XSD/XML

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/Developer.java
```diff
@@ -58,7 +58,7 @@
     /**
      * Indication of whether a health IT developer is a "self-developer" or not.
      */
-    @XmlElement(required = true)
+    @XmlElement(required = true, nillable = true)
```
andlar: The data-model defines this as `self_developer bool NOT NULL DEFAULT false,`. Should it be nillable?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/TestProcedure.java
```diff
@@ -2,14 +2,36 @@
...
 
-    public TestProcedure() {
-        super();
-    }
+    private Long id;
+    private String name;
 
-    public TestProcedure(Long id, String name) {
-        super(id, name);
+    public Long getId() {
```
andlar, line 25: Could these be handled with the `@Data` Lombok annotation? I have a vague memory that our XML things can't use those annotations for some reason, but wanted to double check

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/DeveloperManager.java
```diff
@@ -70,6 +70,7 @@
 @Log4j2
 public class DeveloperManager extends SecuredManager {
     public static final String NEW_DEVELOPER_CODE = "XXXX";
+    private static final Integer FIVE_SECONDS = 5000;
```
andlar: Maybe `DELAY_BEFORE_JOB_START` so there's meaning behind the magic number?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/TestProcedure.java
kekey1: I also recall that we can't use @Data in classes that are involved in the XSD/XML stuff. That's why I didn't.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/Developer.java
kekey1: Fixed. This was leftover from an attempt to avoid refactoring ProductDTO.

andlar (dismissed):

tmy1313 (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1004, opened 2020-08-14
OCD-3371: Connect to Jira for direct reviews

kekey1:


Plow74 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/DirectReviewService.java
```diff
@@ -1,118 +1,73 @@
...
-import gov.healthit.chpl.domain.compliance.DirectReviewNonconformity;
+import gov.healthit.chpl.domain.compliance.DirectReviewNonConformity;
 import lombok.extern.log4j.Log4j2;
 
 @Component("directReviewService")
 @Log4j2
 public class DirectReviewService {
+    @SuppressWarnings("checkstyle:linelength")
+    private static final String JIRA_DIRECT_REVIEW_URL = "/search?jql=project=\"Review for Signals/Direct Review\" and type=\"Direct Review\" and \"CHPL Developer ID\"~\"%s\" and \"Make Visible to CHPL\"=\"Yes\"";
+    private static final String JIRA_NONCONFORMITY_URL = "/search/?jql=project=\"Review for Signals/Direct Review\" and type=\"Requirement/Non-Conformity\" and \"Make Visible to CHPL\"=\"Yes\" and parent=\"%s\"";
```
tmy1313: Should these be in a properties file?

tmy1313 (dismissed):

chpl/chpl-api/src/main/java/gov/healthit/chpl/ApiExceptionControllerAdvice.java
```diff
@@ -169,25 +170,52 @@
...
                         : "A reason is required to perform this action."),
                 HttpStatus.BAD_REQUEST);
     }
 
+    @ExceptionHandler(HttpClientErrorException.class)
+    public ResponseEntity<ErrorResponse> exceptionHandler(HttpClientErrorException e, HttpServletRequest request) {
+        LOGGER.error("HttpClientErrorException: " + e.getMessage(), e);
+        return new ResponseEntity<ErrorResponse>(
+                new ErrorResponse(e.getMessage() != null ? e.getMessage()
+                        : "The request could not be completed. Please check your credentials and the connection to the server."),
```
andlar, line 194: Is this the message that gets sent back if the credentials for logging into Jira are wrong? If so, I'm not sure it's working. There was a report of a message that's just {status: 401} or something like it when the credentials were wrong. Not sure how much it matters, or even if this is the right place for it, but wanted to ask

kekey1: It's working. The "e.getMessage()" value from Jira is literally "401". I preferred to send back the error message from Jira thinking it would be more useful. I can change up this logic to just always send back a string of our choosing if you would prefer.

andlar: Meh. Maybe take a look at it with the tickets in the backlog about "what to do when it doesn't respond" stuff. I think it's good now

andlar (approved):

tmy1313 (approved):

PR #1005, opened 2020-08-16
OCD-3246 Verifying return status for API Requests with invalid API key

athomas-git:
OCD-3246 Verifying return status for API Requests with invalid API key


chpl/chpl-api/e2e/collections/api-controller.postman_collection.json
```diff
@@ -0,0 +1,74 @@
...
+	"item": [
+		{
+			"name": "GET /api-docs",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"id": "655ac14f-d31c-4384-b1c4-4d82281be6b1",
+						"exec": [
+							"pm.test(\"Status code is 200\", function () {\r",
```
palakpatel311: Can we give more readable test names rather than status is 200. May be this test could be worded as 'api controller returns status 200 even if api key is not passed'? 

Please update all of the controller's test name to be more readable.

chpl/chpl-api/e2e/collections/api-controller.postman_collection.json
```diff
@@ -0,0 +1,74 @@
+{
+	"info": {
+		"_postman_id": "daa83386-aeb9-4596-a942-624fad9a7588",
+		"name": "api-controller",
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+	},
+	"item": [
+		{
+			"name": "GET /api-docs",
```
palakpatel311: Can we give this name of request a meaningful name such as 'GET/ api-docs - Without api key'? ( not sure what this test is)

palakpatel311: Same changes needed in all controllers

chpl/chpl-api/e2e/collections/api-controller.postman_collection.json
```diff
@@ -0,0 +1,74 @@
+{
+	"info": {
+		"_postman_id": "daa83386-aeb9-4596-a942-624fad9a7588",
+		"name": "api-controller",
```
palakpatel311: Which controller is this? I dont find it on our chpl-api page.

chpl/chpl-api/e2e/collections/developer-controller.postman_collection.json
```diff
@@ -0,0 +1,54 @@
...
+		{
+			"name": "GET /developers",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"id": "89af904b-9c4d-4c17-9fd1-83ff5d945beb",
+						"exec": [
+							"pm.test(\"Invalid api key header should return 401 status\", function () {\r",
+							"    pm.response.to.have.status(401);\r",
```
palakpatel311, line 17: Is there any reason some controllers have both valid and invalid api test and some dont. Example, this developer controller has only invalid api key test. Why not add valid api key test?
Status controller has both valid and invalid api key tests, which is good

chpl/chpl-api/e2e/collections/products-controller.postman_collection.json
```diff
@@ -0,0 +1,55 @@
...
+		{
+			"name": "GET /products",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"id": "e517dbed-0a00-403e-8fe6-429adda650cc",
+						"exec": [
+							"pm.test(\"Invalid api key parameter should return 401 status\", function () {\r",
+							"    pm.response.to.have.status(401);\r",
```
palakpatel311, line 17: Is there any reason some controllers have both valid and invalid api test and some dont. Why not add valid api key test?

chpl/chpl-api/e2e/collections/testing-lab-controller.postman_collection.json
```diff
@@ -0,0 +1,43 @@
...
+		{
+			"name": "{{url}}/rest/atls",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"id": "e73cd4a6-ff42-430e-8118-88ec9f3bfb84",
+						"exec": [
+							"pm.test(\"Status code is 400\", function () {\r",
+							"    pm.response.to.have.status(400);\r",
```
palakpatel311, line 17: Test name needs to be updated.

Is there any reason some controllers have both valid and invalid api test and some dont. Why not add valid api key test?

chpl/chpl-api/e2e/collections/api-controller.postman_collection.json
athomas-git: /api-docs end point is not part of api-controller , deleted api-controller collection and added the test to verify response without api-key into status controller

chpl/chpl-api/e2e/collections/api-controller.postman_collection.json
athomas-git: fixed

chpl/chpl-api/e2e/collections/developer-controller.postman_collection.json
athomas-git: fixed

chpl/chpl-api/e2e/collections/products-controller.postman_collection.json
athomas-git: fixed

chpl/chpl-api/e2e/collections/testing-lab-controller.postman_collection.json
athomas-git: fixed

palakpatel311 (dismissed):

ashwinimore (dismissed):

chpl/chpl-api/e2e/script/newman-run-dev.js
```diff
@@ -0,0 +1,76 @@
...
+
+const acbControllerTests={
+	...commonOptions,
+	collection: collection_path + '/acb-controller.postman_collection.json',
+	reporter : {
+		junit :{
+			export: reports_path + '/acb-controller-tests.xml'
+		}
+	}
+}
```
drbgfc, line 22: It would be good to decide on a consistent formatting and stick to it. This can be done automatically with a scripted formatter if desired. Above we have examples of:
type name= value
type name = {
type name={
type : {
type :{

I would suggest updating to the following options consistently to match our existing formatting:
type name = value
type : {

If you have a conviction for another formatting, or keeping the varying formatting, I won't hold back the PR. Just mentioning as, even though not all that important, this would not get through non-testing code PR review. The pros are that It helps readability, but, more importantly if all style is the same from all users it helps identify actual impactful changes vs style changes in new PRs.

Also, it is best practice to have a new line at the end of file. Only mentioning cause I have already made this comment...it's not a big deal enough to hold back a PR.

chpl/chpl-api/e2e/script/newman-run-dev.js
```diff
@@ -0,0 +1,76 @@
+const path= require('path'),
+async= require('async'),
+newman= require('newman');
+
+const collection_path= path.resolve('chpl/chpl-api/e2e/collections')
```
drbgfc, line 5: These 2 const declarations are missing semicolons. Technically JS doesn't need them, but, it is best practice to avoid bugs and other statments in the file have them (as does other CHPL JS) so it is best for consistency as well. I would request that they are added please.

chpl/chpl-api/e2e/collections/developer-controller.postman_collection.json
```diff
@@ -0,0 +1,99 @@
...
+						"rest",
+						"developers"
+					]
+				}
+			},
+			"response": []
+		}
+	],
+	"protocolProfileBehavior": {}
+}
```
drbgfc, line 99: Mulitple files are missing an extra newline. I would never hold back a PR for it, just mentioning since other changes might be made.

chpl/chpl-api/e2e/script/newman-run-dev.js
andlar: Agree with Dan's comments, except for `type : {`. In the UI code those are always `type: {`

I'd actually suggest adding eslint with the same configs as the UI project to this one, targeted at the js files, so we can be consistent across the projects while in the same language. I don't know if we have a json linter anywhere, but that would be good to investigate as well

drbgfc: Good point Andrew - I guess I should have linted my comment :-)

chpl/chpl-api/e2e/collections/products-controller.postman_collection.json
```diff
@@ -0,0 +1,95 @@
...
+						"products\n"
+					],
+					"query": [
+						{
+							"key": "api_key",
+							"value": "atest123cg"
+						}
+					]
+				},
+				"description": "Request with Invalid api_key parameter returns 401 status and empty response"
```
andlar: I think this description doesn't match what the test says it's expecting. It looks like the test is only expecting the call to have a 401, but not anything about the response

drbgfc (dismissed):

drbgfc (approved):

chpl/chpl-api/e2e/collections/products-controller.postman_collection.json
athomas-git: Fixed: updated test description

andlar (approved):

ashwinimore (approved):

PR #1007, opened 2020-08-19
test: Adding data driven testing example of acb controller

palakpatel311:
[#OCD-3505]


andlar (changes requested):
Can you update whatever editor you're using to use spaces instead of tabs and re-indent these files? In all of our other code on this project we use spaces, and with the change to tabs I can't see what any actual changes are

palakpatel311:
Checkstyle is not formatting .js files. so I am not sure how to format those in eclipse.
Json collection is formatted.

andlar (commented):
Is this code on DEV? Is it getting tested overnight?

palakpatel311:
Yes it is running overnight in dev

PR #1011, opened 2020-08-25
Ocd 3393 - Add ability to validate and save Real World Testing data

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/Edition2014ListingValidator.java
```diff
@@ -147,6 +152,7 @@
             comparisonReviewers.add(chplNumberComparisonReviewer);
             comparisonReviewers.add(devBanComparisonReviewer);
             comparisonReviewers.add(listingStatusAndUserRoleReviewer);
+            comparisonReviewers.add(realWorldTestingReviewer);
```
andlar, line 155: RWT isn't valid for 2014 Edition listings, right? Why review them here? Or, if they need to be reviewed, shouldn't it be a different reviewer? One that rejects anything with that data?

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/listing/CertifiedProductEntity.java
```diff
@@ -152,36 +151,36 @@
     @Column(name = "last_modified_user", nullable = false)
     private Long lastModifiedUser;
 
-    public CertifiedProductEntity(Long id) {
-        this.id = id;
-    }
+    @Basic(optional = true)
+    @Column(name = "rwt_plan_url", nullable = true)
```
andlar: Does this need a data-model PR?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProduct.java
```diff
@@ -69,6 +70,36 @@
...
+     * The current Real World Testing Plan URL
+     */
+    @XmlElement(required = false, nillable = true)
+    private String rwtPlanUrl;
+
+    /**
+     * The date the Real World Testing Plan submission was confirmed
+     */
+    @XmlElement(required = false, nillable = true)
+    private Date rwtPlanSubmissionDate;
```
kekey1: I'm curious how you chose Date vs the Long millis-since-epoch type value.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
```diff
@@ -809,7 +833,39 @@ public LegacyCertificationStatus getCertificationStatus() {
...
+    public void setRwtPlanUrl(String rwtPlanUrl) {
+        this.rwtPlanUrl = rwtPlanUrl;
+    }
+
+    public Date getRwtPlanSubmissionDate() {
+        return rwtPlanSubmissionDate;
+    }
+
+    public void setRwtPlanSubmissionDate(Date rwtPlanSubmissionDate) {
+        this.rwtPlanSubmissionDate = rwtPlanSubmissionDate;
```
kekey1: Sometimes we use this Util.getNewDate method in date getters/setters and I'm not really sure why. Since you are using a Date field here, do you need to use that method?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertifiedProductDTO.java
```diff
@@ -48,7 +49,11 @@
     private String productAdditionalSoftware;
     private Long pendingCertifiedProductId;
     private Boolean transparencyAttestation = null;
-    private Integer rwtEligiblityYear;
+    private String rwtPlanUrl;
+    private Date rwtPlanSubmissionDate;
```
kekey1: Do the new date fields needs getters/setters using Util.getNewDate?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertifiedProductDetailsDTO.java
```diff
@@ -240,6 +250,46 @@ public void setDecertificationDate(Date decertificationDate) {
...
+    public void setRwtPlanUrl(String rwtPlanUrl) {
+        this.rwtPlanUrl = rwtPlanUrl;
+    }
+
+    public Date getRwtPlanSubmissionDate() {
+        return rwtPlanSubmissionDate;
+    }
+
+    public void setRwtPlanSubmissionDate(Date rwtPlanSubmissionDate) {
+        this.rwtPlanSubmissionDate = rwtPlanSubmissionDate;
```
kekey1: Same questions in this class about the Date thing.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertifiedProductDetailsDTO.java
```diff
@@ -240,6 +250,46 @@ public void setDecertificationDate(Date decertificationDate) {
         this.decertificationDate = Util.getNewDate(decertificationDate);
     }
 
+    public String getRwtPlanUrl() {
+        return rwtPlanUrl;
+    }
+
+    public void setRwtPlanUrl(String rwtPlanUrl) {
+        this.rwtPlanUrl = rwtPlanUrl;
```
kekey1: Do you need getters/setters for the String fields? This class is annotated with @Data I think.

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/listing/CertifiedProductEntity.java
```diff
@@ -152,36 +151,36 @@
...
+     * @param id
+     *            to set
+     */
+    public CertifiedProductEntity(Long id) {
+        this.id = id;
     }
 
-    public void setSedTestingEnd(final Date sedTestingEnd) {
-        this.sedTestingEnd = Util.getNewDate(sedTestingEnd);
+    @Transient
```
kekey1: I don't know what this method is for... why do we need it?

kekey1 (changes requested):
Not all are changes needed... some are just comments/questions.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProduct.java
```diff
@@ -69,6 +70,36 @@
     @XmlElement(required = false, nillable = true)
     private Boolean curesUpdate;
 
+    /**
+     * The first year that the listing is eligible Real World Testing
+     */
+    @XmlElement(required = false, nillable = true)
```
kekey1: Why do we need these elements here? I thought they might only be in the CertifiedProductSearchDetails class.

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/Edition2014ListingValidator.java
tmy1313: We need to make sure that the user has not passed in any RWT information.  Based on the reviewer checking for a rwtEligibility year, which can only exist for a 2015 listing,by adding this reviewer to 2014, it will always reject the data if it is passed in.

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
tmy1313: If someone can tell me the purpose of creating a new date in this instance, I'd be more than happy to use it.  But like any other setter where the parameter is an object, we don't create a copy of the object before setting the private property.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertifiedProductDTO.java
tmy1313: Same as above.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertifiedProductDetailsDTO.java
tmy1313: Same as above

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProduct.java
tmy1313: Probably should be longs to be more consistent across the download files.  I have actually removed these elements from this class, but I think the concept applies to `CertifiedProductSearchDetails`.

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
```diff
@@ -349,7 +349,7 @@
      * Date the listing's Real World Testing Plan was submitted
      */
     @XmlElement(name = "rwt_plan_submission_date", nillable = true, required = false)
-    private Date rwtPlanSubmissionDate;
+    private Long rwtPlanSubmissionDate;
```
andlar: Is this documented in the "date" page on the wiki? I'm not sure that the RWT submission Date shouldn't be a "Date" type

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/RealWorldTestingReviewer.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Autowired
+    public RealWorldTestingReviewer(ValidationUtils validationUtils, ErrorMessageUtil errorMessageUtil) {
+        this.validationUtils = validationUtils;
+        this.errorMessageUtil = errorMessageUtil;
+    }
+
+    @Override
+    public void review(CertifiedProductSearchDetails existingListing, CertifiedProductSearchDetails updatedListing) {
+        //Always use the rwt elig year fvrom the existing listing
```
andlar:
```suggestion
        //Always use the rwt elig year from the existing listing
```

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/Edition2014ListingValidator.java
andlar: I think the RWT reviewer should check against the Listing's edition. We can't know, for sure, that only 2015 Edition Listings get that rwt eligibility date. I know the code that adds the date as a value filters to 2015, but with defensive coding practices we ought to check from this end as well

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
kekey1: A Java Date represents a timestamp and we usually send that back as a long. If this is supposed to be a date-without-time type of date then I think it should be a different type entirely. I believe LocalDate? That would match the data model type.

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/presenter/CertifiedProductCsvPresenter.java
```diff
@@ -96,7 +97,11 @@ public Logger getLogger() {
         result.add("Product Name");
         result.add("Product Database ID");
         result.add("Version");
-        result.add("Version Database ID");
```
andlar, line 99: I don't think this line should be removed

andlar (changes requested):

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertifiedProductDTO.java
```diff
@@ -158,4 +172,28 @@ public CertifiedProductDTO(CertifiedProductSearchDetails from) throws InvalidArg
...
+
+    public void setCreationDate(final Date creationDate) {
+        this.creationDate = Util.getNewDate(creationDate);
+    }
+
+    public Date getLastModifiedDate() {
+        return Util.getNewDate(lastModifiedDate);
+    }
+
+    public void setLastModifiedDate(final Date lastModifiedDate) {
```
andlar: Probably don't need to be final

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductDetailsManager.java
```diff
@@ -719,6 +720,14 @@ private CertifiedProductSearchDetails getCertifiedProductSearchDetails(Certified
         searchDetails.setQmsStandards(getCertifiedProductQmsStandards(dto.getId()));
         searchDetails.setTargetedUsers(getCertifiedProductTargetedUsers(dto.getId()));
         searchDetails.setAccessibilityStandards(getCertifiedProductAccessibilityStandards(dto.getId()));
+        searchDetails.setRwtPlansUrl(dto.getRwtPlansUrl());
+        if (Objects.nonNull(dto.getRwtPlansCheckDate())) {
```
andlar: Why does this need to check against null where the URL doesn't, and it doesn't look like any of the other fields do either?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/QuestionableActivityManager.java
```diff
@@ -206,6 +209,16 @@ public void checkListingQuestionableActivityOnEdit(CertifiedProductSearchDetails
                 createListingActivity(activity, origListing.getId(), activityDate, activityUser,
                         QuestionableActivityTriggerConcept.CRITERIA_B3_ADDED_TO_EXISTING_LISTING_WITH_ICS, activityReason);
             }
+            activity = listingQuestionableActivityProvider.checkRealWorldTestingPlanRemoved(origListing, newListing);
+            if (Objects.nonNull(activity)) {
```
andlar: I don't like the inconsistency of having the other null checks for activity be `activity != null` but these two using the `Objects.nonNull`. It'd be one thing if they were in separate files, but here it just looks like we don't have a standard

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

andlar (approved):

kekey1 (approved):

PR #1015, opened 2020-09-03
OCD-3392: Send user an email when their account is locked.

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/auth/UserManager.java
```diff
@@ -189,9 +194,14 @@ public void updateFailedLoginCount(UserDTO userToUpdate) throws UserRetrievalExc
         String maxLoginsStr = env.getProperty("authMaximumLoginAttempts");
         int maxLogins = Integer.parseInt(maxLoginsStr);
 
-        if (userToUpdate.getFailedLoginCount() >= maxLogins) {
+        if (userToUpdate.getFailedLoginCount() == maxLogins) {
```
andlar: I know it's unlikely, but if something happened where the failed login count jumped over the maxLogins value, and somehow didn't actually lock the account, then we never would, right? Can we keep this comparison as `>=` and add a separate `if` inside this block to send the email only when the `==` comparison is true?

andlar (changes requested):

andlar (dismissed):

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/AuthenticationController.java
```diff
@@ -253,17 +253,20 @@ public String resetPassword(@RequestBody final UserResetPasswordRequest userInfo
...
         UserResetTokenDTO userResetTokenDTO = userManager.createResetUserPasswordToken(userInfo.getUserName(),
                 userInfo.getEmail());
-        String htmlMessage = "<p>Hi, <br/>" + "Please follow this link to reset your password </p>" + "<pre>"
-                + env.getProperty("chplUrlBegin") + "/#/admin/authorizePasswordReset?token="
-                + userResetTokenDTO.getUserResetToken() + "</pre>" + "<br/>" + "</p>" + "<p>Take care,<br/> "
-                + "The Open Data CHPL Team</p>";
+        String htmlMessage = "<p>Please follow this link to reset your password </p>"
+                + "<pre>" + env.getProperty("chplUrlBegin")
+                    + "/#/admin/authorizePasswordReset?token=" + userResetTokenDTO.getUserResetToken()
+                + "</pre>";
```
tmy1313: This should be moved to the email.properties file.

tmy1313 (dismissed):

andlar (dismissed):

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1017, opened 2020-09-10
Ocd 3435 - Clean up duplicate ONC and ONC-ACB scheduled jobs report

tmy1313:
OCD-3435


chpl/chpl-resources/src/main/resources-production/user-triggers.xml
```diff
@@ -4,12 +4,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://www.quartz-scheduler.org/xml/JobSchedulingData http://www.quartz-scheduler.org/xml/job_scheduling_data_2_0.xsd"
     version="2.0">
-    <pre-processing-commands>
```
andlar, line 7: These pre-processing commands were here for a reason. I don't want to remove them unless we're sure they can be

chpl/chpl-resources/src/main/resources/email.properties
```diff
@@ -22,7 +22,6 @@ questionableActivityNoDataEmailBody=<p>No questionable activity was found betwee
 #pending change request email properties
 pendingChangeRequestReportFilename=pending-change-request
 pendingChangeRequestEmailSubject=Pending Change Request Report
-pendingChangeRequestEmailAcbSubject=ONC-ACB Pending Change Request Report
```
andlar, line 25: There were four duplicate reports; did only one of them have values in this property file?

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -186,7 +126,7 @@
                 </entry>
                 <entry>
                     <key>email</key>
-                    <value></value>
+                    <value>amore@ainq.com☺aiyer@ainq.com</value>
```
andlar: Will this overwrite whatever value exists in PROD? We might need to treat this job separately if so. Even if not, probably don't need that second email address in here

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -597,12 +321,48 @@
...
+            <durability>true</durability>
+            <recover>false</recover>
+            <job-data-map>
+                <entry>
+                    <key>authorities</key>
+                    <value>ROLE_ADMIN;ROLE_ONC</value>
+                </entry>
+            </job-data-map>
+        </job>
+
```
andlar, line 364: Can we add a comment above this job that marks the transition to ONC-ACB specific jobs? Something like the ` <!-- Real World Testing Eligibility Job -->` things above?

andlar (changes requested):

chpl/chpl-resources/src/main/resources-production/user-triggers.xml
tmy1313: The pre-processing commands are still in `resources/jobs.xml` which is used by all environments.

chpl/chpl-resources/src/main/resources/email.properties
tmy1313: This was the only report that used a different subject for the two emails.

chpl/chpl-resources/src/main/resources/jobs.xml
tmy1313: This has been addressed with `<ignore-duplicates>` & `<overwrite-existing-data>` changes in jobs.xml

andlar (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1021, opened 2020-09-18
OCD-2968 -  feat!: Reschedule 2014 Download file generation to be monthly

andlar:


chpl/chpl-resources/src/main/resources/system-triggers.xml
```diff
@@ -23,7 +23,8 @@
                 <job-name>downloadFileJob2014</job-name>
                 <job-group>systemJobs</job-group>
                 <misfire-instruction>MISFIRE_INSTRUCTION_DO_NOTHING</misfire-instruction>
-                <cron-expression>0 0 6 * * ?</cron-expression> <!-- At 0600 UTC every day -->
+                <cron-expression>0 45 10 ? * 1#1</cron-expression> <!-- At 1045 UTC on the first Sunday of the month -->
+                <!--<cron-expression>0 45 10 ? JAN,APR,JUL,OCT 1#1</cron-expression> At 1045 UTC on the first Saturday of the quarter -->
```
kekey1, line 27: Is this comment needed?

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -468,7 +468,7 @@
 
         <!-- Pending Change Request Job - by ONC-ACB -->
         <job>
-            <name>Pending Change Request Report</name>
+            <name>Pending "Change Request" Report</name>
```
kekey1, line 471: Just curious - where was the request to rename this job? I didn't see a note about it in the ticket description nor did I see a test case to check it.

kekey1 (changes requested):

chpl/chpl-resources/src/main/resources/system-triggers.xml
andlar: That's going to be the schedule when it goes to quarterly in a month or two. Seemed reasonable to put it here for then, as it's a planned change, but I could remove it if necessary

kekey1: Ok, I'm fine with that then. Will resolve this one.

chpl/chpl-resources/src/main/resources/jobs.xml
andlar: This fixes [#OCD-3525], and the third commit tags that ticket. The ticket is in the backlog, and is the issue that we have with the code going to PROD today. I could remove this, but I was investigating it while in this code and saw this as an easy fix to an issue. When OCD-3525 comes up we'll be able to mark it as "cannot reproduce"

andlar: I consider it part of the "leave it better than you found it" principle

kekey1 (approved):

tmy1313 (approved):

PR #1027, opened 2020-10-01
OCD-3508- Edit SVAP information for specific criteria in listing details

tmy1313:
OCD-3508


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SearchViewController.java
```diff
@@ -1370,4 +1375,12 @@ private FuzzyChoices updateFuzzyChoices(final FuzzyChoices fuzzyChoices)
         result.setData(data);
         return result;
     }
+
+    @ApiOperation(value = "Get all possible SVAP and associated criteri in the CHPL")
```
andlar:
```suggestion
    @ApiOperation(value = "Get all possible SVAP and associated criteria in the CHPL")
```

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
```diff
@@ -213,6 +219,14 @@
     @Singular
     private List<MacraMeasure> g2MacraMeasures = new ArrayList<MacraMeasure>();
 
+    /**
+     * NEED TO GET THIS TEXT
```
andlar: Still need to

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertificationResultManager.java
```diff
@@ -336,15 +348,12 @@ public int update(final CertifiedProductSearchDetails existingListing,
     private Boolean haveMacraMeasuresChanged(List<MacraMeasure> orig, List<MacraMeasure> updated) {
         if (orig != null && updated != null) {
             return !orig.equals(updated);
-        } else if (orig != null && updated == null) {
-            return true;
-        } else if (orig == null && updated != null) {
```
andlar: Why are these other branches removed?

chpl/chpl-service/src/main/java/gov/healthit/chpl/svap/domain/CertificationResultSvap.java
```diff
@@ -0,0 +1,94 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import gov.healthit.chpl.svap.entity.CertificationResultSvapEntity;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+
+/**
+ * NEED THIS TEXT
```
andlar: Reminder

chpl/chpl-service/src/main/java/gov/healthit/chpl/svap/domain/SvapCriteriaMap.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class SvapCriteriaMap {
+    private static final String INVALID_EDITION_ERROR_KEY = "listing.criteria.f3CannotHaveGap";
```
andlar: Don't think these are needed

andlar (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertificationResultDAO.java
```diff
@@ -1270,4 +1272,79 @@ public void deleteTestParticipantMapping(Long testTaskId, Long testParticipantId
...
+        entityManager.persist(entity);
+        entityManager.flush();
+        entityManager.refresh(entity);
+    }
+
+    private CertificationResultSvapEntity getCertificationResultSvapEntityById(Long certificationResultSvapId) {
+        Query query = entityManager.createQuery(
+                "SELECT s "
+                        + "FROM CertificationResultSvapEntity s "
+                        + "LEFT OUTER JOIN FETCH s.svap "
```
kekey1: Does this need to be a left join or would JOIN FETCH be sufficient? I don't think there would be an entry in this table if the association wasn't there.

chpl/chpl-service/src/main/java/gov/healthit/chpl/svap/domain/CertificationResultSvap.java
```diff
@@ -0,0 +1,94 @@
...
+@XmlAccessorType(XmlAccessType.FIELD)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+@ToString
+public class CertificationResultSvap implements Serializable {
+    private static final long serialVersionUID = -1935940788953178006L;
+
+    /**
```
kekey1, line 33: If this stuff is supposed to be in the XSD, the classes need to get added into the schemagen operation in pom.xml

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertificationResultManager.java
tmy1313: It was suggested by Checkstyle that they should be removed.

andlar: Then checkstyle is wrong. If the original thing is null, and the updated thing is not null, then we want this method to tell us that the measures have changed, so it needs to return true. In that case, with the new code, we do not return from line 350, because in 349 at least one of the things is null, so we pass that check. That takes us to the else, which, in the first clause, checks that `orig != null`, but `orig == null`, so it short-circuits the conditional and returns `false`, which is incorrect

tmy1313: I'll revert the code....
Just to show I'm not crazy!  
image
Then when I click on the "Simplify return statement" link, it automatically refactored to:
image

andlar: Never said you were crazy! Just walking through the logic...

andlar (dismissed):

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
andlar: Does this text get inserted anywhere? If it's displayed to an end user we need it to be filled in before it goes to PROD

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
kekey1: We should get this text. It gets put in the XSD as a comment that describes the field.

kekey1 (changes requested):

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
```diff
@@ -220,7 +220,10 @@
     private List<MacraMeasure> g2MacraMeasures = new ArrayList<MacraMeasure>();
 
     /**
-     * NEED TO GET THIS TEXT
+     * ONC has established the Standards Version Advancement Process (SVAP) to enable health IT developers’
+     * ability to incorporate newer versions of Secretary-adopted standards and implementation specifications,
+     * as part of the "Real World Testing" Condition and Maintenance of Certification requirement (§170.405)
```
andlar, line 201: Does this `§` character show up correctly in the XSD?

tmy1313: Yep - 
image

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/svap/domain/CertificationResultSvap.java
```diff
@@ -54,6 +54,12 @@
     @XmlElement(required = true)
     private String approvedStandardVersion;
 
+    /**
+     *  Indicates if the SVAP has been deprecated
```
andlar:
```suggestion
     *  Indicates if the SVAP has been replaced
```

andlar (changes requested):
Still needs update to XMLElement description

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/svap/domain/CertificationResultSvap.java
```diff
@@ -22,6 +24,7 @@
  *
  */
 @XmlType(namespace = "http://chpl.healthit.gov/listings")
+@XmlAccessorOrder(XmlAccessOrder.ALPHABETICAL)
```
andlar: What is this for? Should this be standard?

tmy1313: This is often put in package-info.java in our code, but this particular package does not package-info.java.  Often, package-info.java is used for JavaDoc documentation about a package.  You can also specify annotations that apply to all classes in the package.  I'm not sure if we really have a standard for this.

andlar: I guess I'm asking why we need it now. We've not used it in any of our other XML domain definition documents, so what's the purpose here? If this is a good thing to add, then should we add it to all of our things? If it's superfluous or otherwise "extra", then why do we have it?

andlar: And if it is usually done in the package-info file, and is needed, then why isn't it in that file here?

kekey1: Everywhere else we use this annotation it is used in the package-info file so that it applies to all the classes in the package. I'm not sure if there are other classes in the package that it would apply to, but I'd prefer it be in a package-info file in case we add more classes to the package later (at which time we would surely forget to add this annotation to them). Wherever it ends up going, it IS needed so that the XSD generates elements in a defined and expected order.

andlar: If we have a standard for how we get these things to be alphabetized, and that standard is the package-info.java file, then we should stick to it.

andlar (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/svap/domain/CertificationResultSvap.java
tmy1313: Updated...  In the future, do we want to move these to `package-info`? I think it is a very similar situation. 
image

kekey1 (dismissed):

andlar (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertificationResult.java
```diff
@@ -163,6 +168,7 @@
      * This variable indicates if any additional software is relied upon by the Health IT Module to demonstrate its
      * compliance with a certification criterion or criteria. It is applicable for 2014 and 2015 Edition.
      */
+    @SuppressWarnings("checkstyle:linelength")
```
kekey1: If this PR removes the line length rule - do you need this?

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #1028, opened 2020-10-06
OCD- 3055 Correct the wrong URLs in few emails

palakpatel311:
[#OCD-3055]


kekey1 (changes requested):
- There is a "/product" URL that should be "/listing" put into the Broken Surveillance Rules Report.
- There is a "/product" URL that should be "/listing" put into the ICS Errors Report
- There is a "/product" URL that should be "/listing" put into both Surveillance CSV Downloads
- Consider moving the URL parts out of the code and into a properties file. It's not done consistently but I think we prefer it as a property.

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/TriggerDeveloperBanJob.java
```diff
@@ -91,15 +91,9 @@ private String createHtmlEmailBody(JobExecutionContext jobContext) {
...
-                + "and certified by \"%s\" has been set on \"%s\" by \"%s\" to a Certification Status of \"%s\" with "
-                + "an effective date of \"%s\".</p>"
-                + "<p>%s</p>"
-                + "<p>%s</p>"
-                + "<p>There %s %d Open Nonconformit%s and %d Closed Nonconformit%s.</p>"
-                + "<p>ONC should review the activity and all details of the listing to determine if "
-                + "this action warrants a ban on the Developer.</p>",
+        String htmlMessage = String.format(env.getProperty("developerBanEmailBody"),
                 env.getProperty("chplUrlBegin"), // root of URL
+                env.getProperty("developerBanEmailUrl"),
```
andlar: Why isn't this property just in the `developerBanEmailBody` string?

palakpatel311: yes. But Katy had comment to take URL as much as I can so I have it as separate. I can do either way. let me know which way to proceed

kekey1: Here's what I think: If the URL is something that's only in one place then it could be just put right in the HtmlMessage property in the properties file. This URL, however, is used in several places (including the three additional places in my first comment) so it could be kept as a separate property in the properties file but given a more generic name - maybe "listingDetailsUrl" and then used in this place plus the other three places that were changed.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/presenter/SurveillanceCsvPresenter.java
```diff
@@ -152,7 +152,7 @@ public void presentAsFile(final File file, final List<CertifiedProductSearchDeta
             final Surveillance surv) {
         List<String> result = new ArrayList<String>();
         result.add(listing.getChplProductNumber());
-        result.add(env.getProperty("chplUrlBegin") + "#/product/" + listing.getId());
+        result.add(env.getProperty("chplUrlBegin") + "#/listing/" + listing.getId());
```
kekey1: this is the same as the new listingDetailsUrl property right? can we use that?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/InheritanceErrorsReportCreatorJob.java
```diff
@@ -168,7 +168,7 @@ private InheritanceErrorsReportDTO check(CertifiedProductSearchDetails listing)
                 if (!productDetailsUrl.endsWith("/")) {
                     productDetailsUrl += "/";
                 }
-                productDetailsUrl += "#/product/" + listing.getId();
+                productDetailsUrl += env.getProperty("listingDetailsUrl").substring(1) + listing.getId();
```
kekey1: This just feels a little weird. In the above "if" (which is existing code), it is making sure the URL ends in a "/" and then the line you changed to use the property is ignoring the first "/" in the listingDetailsUrl. I know we don't want to have two "//"s but we do have complete control over the values in the properties file so it seems unnecessary to have to do these checks. Maybe the values aren't standardized in the way that they should be.. I don't know! I won't hold up the PR for this comment if changing it would require too many other changes but if you could take a look at it to see if this is all really needed I would appreciate it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/presenter/SurveillanceReportCsvPresenter.java
```diff
@@ -105,7 +105,7 @@ public SurveillanceReportCsvPresenter(Environment env, CertificationCriterionSer
         if (!productDetailsUrl.endsWith("/")) {
             productDetailsUrl += "/";
         }
-        productDetailsUrl += "#/product/" + data.getId();
+        productDetailsUrl += "#/listing/" + data.getId();
```
kekey1: can we use the new listingDetailsUrl property?

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/presenter/SurveillanceReportCsvPresenter.java
palakpatel311: yes only thing while doing it was it was asking to change env property to be protected. Are we okay doing it?

kekey1: You probably don't need to make it protected... the compiler is asking for that because "env" is a private field in the superclass. You should be able to access it with a getter. That is done elsewhere in this file. `getEnv().getProperty("listingDetailsUrl")`

kekey1 (dismissed):

andlar (dismissed):

tmy1313 (dismissed):

andlar (approved):

kekey1 (approved):

tmy1313 (approved):

PR #1041, opened 2020-11-06
OCD-3402: Create direct review cache and download file

kekey1:


chpl/chpl-resources/src/main/resources/quartz.properties
```diff
@@ -7,7 +7,7 @@ org.quartz.jobStore.tablePrefix = quartz.QRTZ_
 org.quartz.jobStore.isClustered = true
 
 org.quartz.plugin.jobInitializer.class = org.quartz.plugins.xml.XMLSchedulingDataProcessorPlugin
-org.quartz.plugin.jobInitializer.fileNames = jobs.xml,system-triggers.xml,user-triggers.xml
+org.quartz.plugin.jobInitializer.fileNames = jobs.xml,ocd-3402.xml,system-triggers.xml,user-triggers.xml
```
andlar: I think we moved towards just putting these jobs & triggers in the jobs.xml and system-triggers.xml files, and we don't need specific ocd-nnnn.xml files. I see the download file generation trigger, but not the cache refresh one

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/AsynchronousCacheInitialization.java
```diff
@@ -12,27 +12,40 @@
...
+    }
 
     @Async
     @Transactional
     public Future<Boolean> initializeSearchOptions() throws EntityRetrievalException {
         LOGGER.info("Starting cache initialization for SearchViewController.getPopulateSearchData()");
         dimensionalDataManager.getSearchableDimensionalData(true);
         dimensionalDataManager.getSearchableDimensionalData(false);
+        dimensionalDataManager.getDimensionalData(true);
+        dimensionalDataManager.getDimensionalData(false);
```
andlar, line 48: Was trying to figure out why we call the same method twice with two different booleans. Don't see anything in the PR about it, so will ask here. This just looks odd...

andlar (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/AsynchronousCacheInitialization.java
kekey1: That's how ehCache works. It will cache the response to the getDimensionalData call with the key "true" and another entry in the appropriate cache with the key "false". Then when we call the method whether we call it with an argument of true or false the data is cached. I don't think this code change was related to the ticket but the above lines (45-46) are deprecated methods and I think we missed caching the non-deprecated methods intended to replace them. That's why I added them here.

andlar (dismissed):

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1043, opened 2020-11-10
[#OCD-3410] feat!: Make "username" not required for new users

andlar:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/AuthenticationController.java
```diff
@@ -271,23 +269,36 @@ public String resetPassword(@RequestBody final UserResetPasswordRequest userInfo
         return "{\"passwordResetEmailSent\" : true }";
     }
 
+    @Deprecated
     @ApiOperation(value = "Impersonate another user.", notes = "")
```
kekey1: Can you add "DEPRECATED" to the text here? Done in a few other controllers with deprecated methods

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -376,14 +373,26 @@ public String deleteUser(@PathVariable("userId") final Long userId)
         return response;
     }
 
+    @Deprecated
     @ApiOperation(value = "View a specific user's details.",
```
kekey1: Add "DEPRECATED" to text

chpl/chpl-service/src/main/java/gov/healthit/chpl/auth/jwt/JWTConsumer.java
```diff
@@ -7,7 +7,7 @@
 public interface JWTConsumer {
 
     /**
-     * Returns map of JWT claims, or returns null if token is invalid 
+     * Returns map of JWT claims, or returns null if token is invalid
      */
-    public Map<String, Object> consume(String jwt) throws InvalidJwtException;
+    Map<String, Object> consume(String jwt) throws InvalidJwtException;
```
kekey1: I feel like this interface and the JWTAuthor interfaces are instances in which we can get rid of the interfaces (like we have been doing with the DAO and DAOImpl, Manager and ManagerImpl, etc. code). If there is no other class that implements these interfaces I *think* we don't need them; the interfaces could be deleted and the implementing class could be renamed to JWTConsumer

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/auth/UserDAO.java
```diff
@@ -251,6 +246,27 @@ private UserEntity getEntityByName(String uname) throws UserRetrievalException {
         return result.get(0);
     }
 
+    private UserEntity getEntityByNameOrEmail(String email) throws MultipleUserAccountsException, UserRetrievalException {
```
kekey1, line 250: I think the addition of this method means there are now two methods in this class with essentially the same HQL. Can you make the one that returns the DTO call this method so the HQL is only in one place?

kekey1 (changes requested):
I had a few small things to be changed in the code. In doing a search of the code for where getSubjectName() is referenced, I would like to ask if you have tested/there are test cases for the following: 
- Performing questionable activity as a user with email only (NULL username) and scheduling the job to send the email. Receive the email and look in it to make sure the data is filled in right. (existing code puts in the username to the report)
- Doing some kind of edit of a user without a username and viewing activity reports
- Viewing any activity report when the activity was performed by a user without a username
- Uploading a surveillance file with more than <threshold for background job> records when the user doesn't have permissions on the ACB that owns the listings in the file (obscure I know but I see subject name referenced in an error message there)

kekey1 (dismissed):

tmy1313 (dismissed):

kekey1 (approved):

tmy1313 (approved):

PR #1044, opened 2020-11-11
OCD-2969 - Update statistics history to use certification status history table

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/SummaryStatisticsCreatorJob.java
```diff
@@ -61,51 +74,71 @@ public SummaryStatisticsCreatorJob() throws Exception {
...
-            Integer numDaysInPeriod = Integer.valueOf(env.getProperty("summaryEmailPeriodInDays").toString());
+            LOGGER.info("Getting all listings.");
+            List<CertifiedProductDetailsDTO> allListings = certifiedProductDAO.findAll();
+            LOGGER.info("Completing getting all listings.");
 
-            EmailStatistics emailBodyStats = emailStatisticsCreator.getStatistics();
+            EmailStatistics emailBodyStats = emailStatisticsCreator.getStatistics(allListings);
+            saveSummaryStatistics(emailBodyStats);
 
+            Boolean generateCsv = Boolean.valueOf(jobContext.getMergedJobDataMap().getString("generateCsvFile"));
```
andlar: Inside the `createSummaryStatisticsFile` method you use the private `isGenerateStatisticsFlagOn` method to immediately exit. Seems like this check only needs to happen in one place, and if it does happen in two they should both use this private method

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/SummaryStatisticsCreatorJob.java
```diff
@@ -61,51 +74,71 @@ public SummaryStatisticsCreatorJob() throws Exception {
...
             DateRange csvRange = new DateRange(startDateCal.getTime(), new Date(endDateCal.getTimeInMillis()));
             CsvStatistics historyStat = new CsvStatistics();
+
             historyStat.setDateRange(csvRange);
-            historyStat = historicalStatisticsCreator.getStatistics(csvRange);
+            historyStat = historicalStatisticsCreator.getStatistics(allListings, statusesForAllListings, csvRange);
+
             csvStats.add(historyStat);
+
             LOGGER.info("Finished getting csvRecord for start date " + startDateCal.getTime().toString() + " end date "
```
andlar: As far as I can tell, the only place the `startDate` part of the `dateRange` variable is used is in this logging statement. Even in the initial "set up the dates" thing, it's not using the dateRange variable yet, just the response from the `getStartDate()` method. Is it worth keeping the dateRange around for just this statement? The log could be written instead as something like `Finished getting records for <numDaysInPeriod> ending at <endDate>` and it would have the same result

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/data/HistoricalStatisticsCreator.java
```diff
@@ -14,33 +17,46 @@
...
 @Component()
 public class HistoricalStatisticsCreator {
     private static final Logger LOGGER = LogManager.getLogger("summaryStatisticsCreatorJobLogger");
+    private static final Long EDITION_2015_ID = 3L;
+    private static final Long EDITION_2014_ID = 2L;
 
     private ListingStatisticsDAO listingStatisticsDAO;
     private DeveloperStatisticsDAO developerStatisticsDAO;
     private SurveillanceStatisticsDAO surveillanceStatisticsDAO;
+    private CertificationStatusEventDAO certificationStatusEventDAO;
```
andlar: Eclipse tells me this field is never used; is that true?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/summarystatistics/data/HistoricalStatisticsCreator.java
```diff
@@ -14,33 +17,46 @@
...
 import gov.healthit.chpl.domain.DateRange;
+import gov.healthit.chpl.dto.CertificationStatusEventDTO;
+import gov.healthit.chpl.dto.CertifiedProductDetailsDTO;
+import gov.healthit.chpl.dto.ProductDTO;
 import gov.healthit.chpl.entity.CertificationStatusType;
+import one.util.streamex.StreamEx;
 
 @Component()
 public class HistoricalStatisticsCreator {
     private static final Logger LOGGER = LogManager.getLogger("summaryStatisticsCreatorJobLogger");
```
andlar: I see Eclipse mark this as not used either. I'm fine keeping it around because I wouldn't want to later want to log something, and forget to use the `summaryStatisticsCreatorJobLogger` thing, but I was wondering if the `@Log4j2` annotation has a way of specifying the logger we should use? I know there's an optional `topic`, but I don't know if that would map to our `getLogger` name. Maybe worth check if `@Log4j2(topic=summaryStatisticsCreatorJobLogger)` would work?

andlar (changes requested):

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #1049, opened 2020-12-07
OCD-3531: Execute developer merge as background quartz job

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/MergeDeveloperJob.java
```diff
@@ -0,0 +1,330 @@
...
+import gov.healthit.chpl.util.EmailBuilder;
+import net.sf.ehcache.CacheManager;
+
+public class MergeDeveloperJob implements Job {
+    public static final String JOB_NAME = "mergeDeveloperJob";
+    public static final String OLD_DEVELOPERS_KEY = "oldDevelopers";
+    public static final String NEW_DEVELOPER_KEY = "newDeveloper";
+    public static final String PRODUCT_IDS_TO_MOVE_KEY = "productIdsToMove";
+    public static final String USER_KEY = "user";
+    private static final Logger LOGGER = LogManager.getLogger("mergeDeveloperJobLogger");
```
tmy1313: I think Andrew showed me that you can use `@Log4j2(topic="mergeDeveloperJobLogger")` for this.

tmy1313 (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/MergeDeveloperJob.java
```diff
@@ -156,6 +158,13 @@ private void setSecurityContext(UserDTO user) {
         SecurityContextHolder.setStrategyName(SecurityContextHolder.MODE_INHERITABLETHREADLOCAL);
     }
 
+    private void checkPreMergeDevelopers() throws EntityRetrievalException {
```
andlar: This method looks like all it does is read Developers. I'm guessing the point of the method is to throw an exception if there's something odd that happens. If that's true, it feels like the method should be named differently, or called differently, or something, so that someone looking at this code in 6 months will recognize that the point is that we're checking to see if an "EntityRetrievalException" is thrown or not

andlar (approved):

tmy1313 (approved):

PR #1053, opened 2020-12-17
OCD-3538: Archive audit data

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/logging/Loggable.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.springframework.stereotype.Component;
+
+@Component
+@Documented
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Loggable {
+
+    //String value() default "INFO";
```
kekey1: Remove commented out line. What is @Documented?

chpl/chpl-service/src/main/java/gov/healthit/chpl/logging/LoggingFilter.java
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.logging.log4j.ThreadContext;
+import org.springframework.web.filter.GenericFilterBean;
+
+public class LoggingFilter extends GenericFilterBean {
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+
+        ThreadContext.put("tracker", UUID.randomUUID().toString());
```
andlar: Is this part of this ticket? I had thought we were going towards using Datadog for this logging stuff. I'm not sure if this stuff should still be here. Not sure about the other `@Loggable` annotations though; if those are Datadog then that's maybe fine, but not sure

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/auditdata/ApiKeyActivityAuditDataRetentionService.java
```diff
@@ -0,0 +1,82 @@
...
+
+        query.executeUpdate();
+    }
+
+    @Override
+    public void archiveDataToFile(Integer month, Integer year, String fileName, boolean includeHeaders) throws SQLException {
+        String copyCmd = "COPY "
+                + "(SELECT * "
+                + "FROM openchpl.api_key_activity "
+                + "WHERE EXTRACT(MONTH FROM creation_date) = " + month.toString() + " "
```
kekey1, line 58: Is this safe? Or should it be/can it be done more like a prepared statement with the ":param" sort of notation we use elsewhere in DAOs? Or even written as a function that gets added to the database and just called from here with the parameters?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/auditdata/LoggedActionsAuditDataRetentionService.java
```diff
@@ -0,0 +1,82 @@
...
+
+        query.executeUpdate();
+    }
+
+    @Override
+    public void archiveDataToFile(Integer month, Integer year, String fileName, boolean includeHeaders) throws SQLException {
+        String copyCmd = "COPY "
+                + "(SELECT * "
+                + "FROM audit.logged_actions "
+                + "WHERE EXTRACT(MONTH FROM action_tstamp) = " + month.toString() + " "
```
kekey1, line 58: Same question I guess about if this is a safe way to create the SQL. If you have an entity I think you can declare a native sql query in that. Might be another idea.. I've done it in some of the Activity entities.

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/logging/LoggingFilter.java
kekey1: I had similar questions about the Loggable changes. If we are proceeding with it I would like a technical overview so I know which classes/methods to add it to, what we are expecting to do with it, are there test cases for the Loggable changes, etc.

tmy1313: This particular change should be removed.  I think the `@Loggable` should remain, though.  The `@Loggable` annotation on a class will write out for each public method called 1) that the method was called, 2) that the method completed, 3) how long the method took.  I think this can be expanded to also include some error logging. I would actually like to expand this to nearly all of our classes.  As far as a unique token to track which calls where made by which request - this will be injected by some DataDog via some java startup params.  Thoughts?

chpl/chpl-service/src/main/java/gov/healthit/chpl/logging/Loggable.java
tmy1313: @Documented was a leftover from a copy/paste...  It has been removed.

kekey1 (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/logging/LoggingAspect.java
```diff
@@ -19,16 +19,17 @@ public Object around(ProceedingJoinPoint proceedingJoinPoint) throws Throwable {
 
         MethodSignature signature = (MethodSignature) proceedingJoinPoint.getSignature();
         Method method = signature.getMethod();
-        Loggable loggableMethod = method.getAnnotation(Loggable.class);
+        //Loggable loggableMethod = method.getAnnotation(Loggable.class);
```
andlar, line 22: Do we need these commented out things?

kekey1 (approved):

andlar (approved):

PR #1055, opened 2020-12-23
OCD-3180 - Add reports on missing / soon-to-be-missing RWT Plans & Results

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/realworldtesting/domain/RealWorldTestingReport.java
```diff
@@ -0,0 +1,64 @@
...
+        results.add(chplProductNumber);
+        results.add(productName);
+        results.add(productId == null ? null : productId.toString());
+        results.add(developerName);
+        results.add(developerId == null ? null : developerId.toString());
+        results.add(rwtEligibilityYear == null ? null : rwtEligibilityYear.toString());
+        results.add(rwtPlansUrl);
+        results.add(rwtPlansCheckDate == null ? null : rwtPlansCheckDate.toString());
+        results.add(rwtResultsUrl);
+        results.add(rwtResultsCheckDate == null ? null : rwtResultsCheckDate.toString());
```
kekey1, line 41: Just curious - does the generated CSV look right when any of the "could be null" values are actually null? I think in the past I've usually put in empty strings instead of null for the CSVs but maybe null works out just fine with the apache CSV library. Approving but I wanted to ask the question!

kekey1 (dismissed):

chpl/chpl-resources/src/main/resources/email.properties
```diff
@@ -170,3 +170,8 @@ rwt_email_styles=table.blueTable {\
         padding: 2px 8px;\
         border-radius: 5px;\
     }
+
+# Real World Testing Report Email properties
+rwt.report.filename=real-world-testing-report-
+rwt.report.subject=ONC Real World Testing Report
```
andlar: Should this be:

```suggestion
rwt.report.subject=ONC-ACB Real World Testing Report
```

The ticket describes the report as an ONC-ACB report, so I'm not sure why this has "ONC" alone in the subject or body

andlar (dismissed):
I have a question about text, but no concerns about the logic, so am leaving the text question to be resolved later

chpl/chpl-service/src/main/java/gov/healthit/chpl/realworldtesting/domain/RealWorldTestingReport.java
tmy1313: Double checked the CSV and NULLs are output as empty strings.

chpl/chpl-resources/src/main/resources/email.properties
tmy1313: Update to "CHPL Real World Testing Report" as per discussion in stand-up.

andlar (approved):

kekey1 (approved):

PR #1058, opened 2020-12-28
[#OCD-3568]feat--flag!: Only put RWT data in downloadable CSV files when flag is enabled

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
```diff
@@ -6,4 +6,5 @@ private FeatureList() {
 
     public static final String CHANGE_REQUEST = "change-request";
     public static final String ROLE_DEVELOPER = "role-developer";
+    public static final String RWT_ENABLED = "rwt-enabled";
```
kekey1, line 9: I'm a little confused about this. Is it weird that we've had a number of RWT tickets on the board but this is the first one that's behind a flag?

kekey1 (changes requested):
I couldn't leave a comment on the package-info file... it shows that it's empty in the diff so I'm not sure what/why it was changed. I think we had agreed that unless it was being used for something we wouldn't require them with checkstyle and could delete them. I've seen one place in our codebase where we actually use the package-info file to apply a package-level annotation but that wasn't in this one.

andlar:
> I couldn't leave a comment on the package-info file... it shows that it's empty in the diff so I'm not sure what/why it was changed. I think we had agreed that unless it was being used for something we wouldn't require them with checkstyle and could delete them. I've seen one place in our codebase where we actually use the package-info file to apply a package-level annotation but that wasn't in this one.

I'd accidentally deleted it, then restored it using "ctrl-z", because I wasn't planning on making any changes in the controllers. Have since deleted it

chpl/chpl-service/src/main/java/gov/healthit/chpl/FeatureList.java
andlar: Yep, it's weird. The difference is that almost all of the previous tickets had no visible effect for anonymous users. Without running the job to make Listings RWT eligible, there was no way to add RWT data to them, and we weren't going to run that job until next year some time, so there was no need to put anything behind a flag

Unfortunately, there was _one_ place where an anonymous user could see indications that RWT even existed, and that was in the columns in the 2015 downloadable CSV. That ended up not being okay for the customer, so this feature removes those columns until that flag is turned on. My guess is we'll set that flag to be a scheduled flag for some time next year, and then forget about it until then

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/CertifiedProductDownloadableResourceCreatorJob.java
```diff
@@ -126,6 +130,7 @@ private void initializeWritingToFiles(CertifiedProductXmlPresenter xmlPresenter,
         xmlPresenter.open(tempXmlFile);
 
         csvPresenter.setLogger(LOGGER);
+        csvPresenter.setFf4j(ff4j);
```
tmy1313, line 133: This might be a bad a pattern I introduced when this was originally written, but we should fix it.  By having these 'setters' it is possible to call `csvPresenter.open( ... )` without having passed in the `Ff4j` or the `LOGGER`.  
Do you think we should make these parameters in the `.open( ... )` method?  That way we are sure that all dependencies exist when making the call?

tmy1313 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/CertifiedProductDownloadableResourceCreatorJob.java
andlar: The only `csvPresenter.open()` call we have is looking for a File input: 
image

I'm not sure if it was a bad pattern when you introduced it, but it was you that did it, though it's also now been this way since Oct of 2018: 
image

I just opened up the code to see if I could switch to the requested change, but I'm not sure how I'd go about doing that. If you'd like to pull this branch to your fork, make whatever changes would be necessary, and PR it back to this branch I'd love to see what you're talking about. Would you also want to make these changes in the XML presenter?

kekey1 (approved):

tmy1313 (approved):

PR #1061, opened 2021-01-11
OCD-3545: Add direct review data to listing CSV and XML files

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductDetailsManager.java
```diff
@@ -343,6 +352,22 @@ private CertifiedProductSearchDetails populateTestingLab(CertifiedProductDetails
         return searchDetails;
     }
 
+    private CertifiedProductSearchDetails populateDirectReviews(CertifiedProductSearchDetails listing) {
+
+        List<DirectReview> drs = new ArrayList<DirectReview>();
+        drs.addAll(drService.getListingDirectReviewsFromCache(listing.getId()));
+        if (listing.getDeveloper() != null && listing.getDeveloper().getDeveloperId() != null) {
+            drs.addAll(drService.getDeveloperDirectReviewsFromCache(listing.getDeveloper().getDeveloperId()));
```
andlar: Should this be `drService.getListingDirectReviewsFromCache` ?

kekey1: I don't think so. Line 358 is getting all the direct reviews that are tied to the listing ID (via developer-associated listings on the non-conformities). Line 360 is getting all the direct reviews associated with the developer ID (so regardless of what listings it's associated with, DRs that have 0 non-conformities, etc). I think that in reality most or all of these will be duplicates of each other so lines 363-365 below do a distinct on the Jira key (the "OCD-1234") so that I don't associate the same DR to a listing multiple times. That is the logic in my head anyway - thoughts? Ideas how I could make that more clear in the code?

andlar: Oh good grief. Yep. Thanks for pointing out line 358... Sign

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1063, opened 2021-01-13
OCD-3582: Associate both cures and original criteria to measures

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
```diff
@@ -39,14 +40,16 @@ public CertificationCriterionService(CertificationCriterionDAO certificationCrit
 
     @PostConstruct
     public void postConstruct() {
-        criteriaMap = certificationCriterionDAO.findAll().stream()
+        criteriaByIdMap = certificationCriterionDAO.findAll().stream()
                 .map(criterion -> new CertificationCriterion(criterion))
                 .collect(Collectors.toMap(CertificationCriterion::getId, cc -> cc));
+        criteriaByIdMap.values().stream()
+            .forEach(criterion -> insertCriterionIntoNumberMap(criterion));
```
andlar: Could this line be inserted in-between the `map` and `collect` lines above?

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
```diff
@@ -57,6 +60,11 @@ public CertificationCriterion get(String certificationCriterionDescriptor) {
         }
     }
 
+    public List<CertificationCriterion> getByNumber(String certificationCriterionNumber) {
+        return criteriaByNumberMap.containsKey(certificationCriterionNumber)
+                ? criteriaByNumberMap.get(certificationCriterionNumber) : null;
```
andlar: I think this can just be a `get`

From the Java docs:

> get(Object key)
> Returns the value to which the specified key is mapped, or null if this map contains no mapping for the key.

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2015Version1.java
```diff
@@ -751,7 +751,7 @@ protected void parseMeasures(PendingCertifiedProductEntity listing, PendingCerti
                             }
                             listing.getMeasures().add(measureEntity);
                         }
-                        addAssociatedCriterionToMeasure(measureEntity, cert);
+                        addAssociatedCriteriaToMeasure(measureEntity, cert);
```
andlar: Is the `addAssociatedCriterionToMeasure` method still used? If not, it ought to be removed

Also, both of these methods have very similar names, and have the same parameters that get passed to them, but using `criteria` in the new one suggests that we're passing in more than one criterion, which I don't see happening. Is there a way to clarify what the difference is in these two methods?

andlar (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
kekey1: This second section is filling in a totally different map so I think it would be weird to move it into the above section that's assigning the value to the criteriaIdMap. I rearranged it somewhat in a way that I think is more clear.

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
kekey1: Agree

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/certifiedProduct/CertifiedProductHandler2015Version1.java
kekey1: I tried renaming them... maybe it's better? One method just adds the criteria associated with the certificationResult that's passed in. I want to do this (and continue to do this) for any measures that aren't found in the database so we can produce an error message. The other method adds the criteria associated with the certificationResult that's passed in AND any other criteria that have the same number. I do this for any valid measure.

Also - I hope this code is removed someday soon in favor of the normalizers in OCD-3549.

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
andlar: I missed that it was a different map

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/CertificationCriterionService.java
```diff
@@ -39,14 +40,17 @@ public CertificationCriterionService(CertificationCriterionDAO certificationCrit
...
                 .collect(Collectors.toMap(CertificationCriterion::getId, cc -> cc));
+        criteriaByNumberMap = criteriaByIdMap.values().stream()
+                .collect(Collectors.groupingBy(CertificationCriterion::getNumber));
+
         referenceSortingCriteriaList = getReferenceSortingCriteriaList();
     }
 
     public CertificationCriterion get(Long certificationCriterionId) {
-        return criteriaMap.containsKey(certificationCriterionId) ? criteriaMap.get(certificationCriterionId) : null;
+        return criteriaByIdMap.containsKey(certificationCriterionId) ? criteriaByIdMap.get(certificationCriterionId) : null;
```
andlar: Can this be just the `criteriaByIdMap.get(certificationCriterionId)` call? As that call returns null if the key doesn't exist?

andlar (dismissed):

tmy1313 (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1069, opened 2021-01-26
OCD-3498 - Questionable Activity Report - activity details missing for old G1-G2 measure activity

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -506,24 +505,33 @@ private void putCertResultActivityInRow(final QuestionableActivityCertificationR
...
+        currRow.set(LINK_COL, env.getProperty("chplUrlBegin") + env.getProperty("listingReportsUrlPart") + "/" + activity.getListing().getId());
+        currRow.set(ACTIVITY_USER_COL, activity.getUser().getSubjectName());
 
         String currActivityRowValue = currRow.get(ACTIVITY_DESCRIPTION_COL);
         if (!StringUtils.isEmpty(currActivityRowValue)) {
             currActivityRowValue += "; ";
         }
 
         if (activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.G1_SUCCESS_EDITED.getName())
+                || activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.G2_SUCCESS_EDITED.getName())) {
```
andlar: Looking at the resulting file, I think I see three places where the output of all of these if/else statements is the same `+=`. In those places it also looks like there's some duplication of `if` clauses. For instance, the check against GAP_EDITED could be true in both `if`s on lines 525 and 527.

There are two places where I think we should consider combining things; the `G1/G2 MEASURE ADDED` and `REPLACED SVAP ADDED` results in almost the same thing, and given that SVAP hasn't been put in the system yet, I suggest we make the output of the SVAP addition QA mimic the format of the addition of the G1/G2 QA stuff

Also, can we replace this whole thing with a `switch` statement? It's really hard to read as is

andlar (changes requested):
Has this been built on DEV? I see the code in the dev branch, but don't see the plan as having been run. I believe we said we were going to build on DEV, and at least smoke test the functionality, before marking the feature as ready for review

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -506,24 +505,38 @@ private void putCertResultActivityInRow(final QuestionableActivityCertificationR
...
 
         String currActivityRowValue = currRow.get(ACTIVITY_DESCRIPTION_COL);
         if (!StringUtils.isEmpty(currActivityRowValue)) {
             currActivityRowValue += "; ";
         }
 
         if (activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.G1_SUCCESS_EDITED.getName())
+                || activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.G2_SUCCESS_EDITED.getName())
+                || activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.GAP_EDITED.getName())
+                || activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.G1_SUCCESS_EDITED.getName())
```
andlar: Am I missing something, or does this set of `or`s have each option twice?

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/QuestionableActivityEmailJob.java
```diff
@@ -506,24 +505,38 @@ private void putCertResultActivityInRow(final QuestionableActivityCertificationR
...
+            currActivityRowValue += formatCriteriaNumber(activity.getCertResult()) + ": " + activity.getAfter();
+
+        } else if (activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.G1_MEASURE_REMOVED.getName())
+                || activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.G2_MEASURE_REMOVED.getName())) {
+
+            currActivityRowValue += formatCriteriaNumber(activity.getCertResult()) + ": " + activity.getBefore();
+
         } else if (activity.getTrigger().getName().equals(QuestionableActivityTriggerConcept.REPLACED_SVAP_ADDED.getName())) {
 
             currActivityRowValue += formatCriteriaNumber(activity.getCertResult()) + ": added " + activity.getAfter();
```
andlar: Is there a reason the SVAP-ADDED one isn't in the if block with the other MEASURE_ADDED options? The only difference I can see is the ` added` as part of the message, but I don't know if that's required, or if there's some other way we can put that in there, in order to simplify the logic of this set of statements

tmy1313: I left it separate because the message was different.  If we want to make the message the same, I'm fine with that.

andlar: The message is different now, but it's not a message anyone has ever seen. So unless there's a firm requirement from the PO that this particular message should be different from all of the ones that we've already had in the system for years, then I don't think we should assume that using a different message is a good idea

andlar (approved):

kekey1 (approved):

PR #1070, opened 2021-01-29
OCD-3476: Add API AQA tests for surveillance upload

ashwinimore:


chpl/chpl-api/e2e/script/newman-run-dev.js
```diff
@@ -104,6 +113,7 @@ const jobs = [
     cb => newman.run(statusControllerTests,cb),
     cb => newman.run(testinglabControllerTests,cb),
     cb => newman.run(searchviewControllerTests,cb),
+	cb => newman.run(surveillanceControllerTests,cb),
```
palakpatel311: Looks like has 2 extra spaces in this line. Same with other newman files.

chpl/chpl-api/e2e/collections/surveillance-controller.postman_collection.json
```diff
@@ -0,0 +1,382 @@
...
+					    "rest",
+						"surveillance",
+						"upload"
+					]
+				}
+			},
+			"response": []
+		},
+		{
+			"name": "POST /surveillance/upload - ROLE_ACB has access to upload surveillance activity on own listings",
```
palakpatel311, line 65: Can we also add a test mentioned in TC document as "ACB A attempts to add surveillance activity on listing owned ACB B via upload"?
This test is not in UI test so good to add it here.

ashwinimore: I'd like to add that test too but I put that one in excluded list due to lack of aqa user account that is attached to one ACB organization only. The one we have now is attached to all active ACB organizations, it won't work for this scenario.

chpl/chpl-api/e2e/script/newman-run-dev.js
ashwinimore: pushed a commit to fix that

chpl/chpl-api/e2e/collections/surveillance-controller.postman_collection.json
palakpatel311: I see. We will need to request new ACB credentials from Andrew. so we can start having that user for these types of test. Not holding PR for this.

palakpatel311 (approved):

athomas-git (approved):

PR #1073, opened 2021-02-03
[#OCD-3575] feat!: Put CHPL Product Number in CMS ID results data

andlar:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CertificationIdController.java
```diff
@@ -330,7 +330,7 @@ private CertificationIdResults findCertificationByProductIds(
         List<CQMMetDTO> cqmDtos = certificationIdManager.getCqmsMetByCertifiedProductIds(productIdList);
 
         boolean isValid = validator.validate(criteriaDtos, cqmDtos, new ArrayList<Integer>(yearSet));
-        results.setIsValid(isValid);
+        results.setValid(isValid);
```
kekey1, line 335: Just confirming - this doesn't change the way the field is represented in the JSON result object right? i.e. not backwards-compatibility breaking?

andlar: No, this object isn't used anywhere except as the response that gets sent back when doing stuff in the API with the CMS ID validation stuff

tmy1313 (approved):

kekey1 (approved):

PR #1075, opened 2021-02-03
OCD-3517 - Enable SVAP Administration

tmy1313:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SvapController.java
```diff
@@ -0,0 +1,73 @@
...
+import gov.healthit.chpl.exception.EntityRetrievalException;
+import gov.healthit.chpl.exception.ValidationException;
+import gov.healthit.chpl.svap.domain.Svap;
+import gov.healthit.chpl.svap.manager.SvapManager;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+
+@Api(value = "svaps")
+@RestController
+@RequestMapping("/svaps")
```
kekey1, line 24: Does it need @Loggable which was added to (all?) other controllers in another PR?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SvapController.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private SvapManager svapManager;
+
+    @Autowired
+    public SvapController(SvapManager svapManager) {
+        this.svapManager = svapManager;
+    }
+
+    @ApiOperation(value = "Update an SVAP.",
+            notes = "NEED TO ADD DESCRIPTION"
```
kekey1: A few of these "notes" fields have this value - can you add it?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductDetailsManager.java
```diff
@@ -277,9 +269,9 @@ public CertifiedProductSearchDetails getCertifiedProductDetailsBasic(Long certif
                 dto.getId(), true);
 
         CertifiedProductSearchDetails searchDetails = getCertifiedProductSearchDetails(dto);
+        List<SvapCriteriaMap> svapCriteriaMap = svapDao.getAllSvapCriteriaMap();
```
kekey1, line 272: Does this need to be called every time we get details for a listing? Does this data change often? I'm curious why you chose to move it out of the init method here. If it does change only occasionally, I wonder if we could do something with Ehcache instead?

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/domains/svap/CreateActionPermissions.java
```diff
@@ -0,0 +1,22 @@
...
+import gov.healthit.chpl.permissions.domains.ActionPermissions;
+
+@Component("svapCreateActionPermissions")
+public class CreateActionPermissions extends ActionPermissions {
+
+    @Override
+    public boolean hasAccess() {
+        return  getResourcePermissions().isUserRoleAdmin()
+            || getResourcePermissions().isUserRoleOnc()
+            || getResourcePermissions().isUserRoleOncStaff();
```
kekey1: Are we letting ONC_STAFF do this? I saw admin and onc in the ticket and I thought we were only letting ONC_STAFF deal with scheduled jobs for now.

chpl/chpl-service/src/main/java/gov/healthit/chpl/svap/manager/SvapManager.java
```diff
@@ -1,24 +1,207 @@
...
+            messages.add(errorMessageUtil.getMessage("svap.edit.noCriteria"));
+        }
+
+        //If there are removed criteria, make sure there are no listings attesting to SVAP/criteria
+        getCriteriaRemovedFromSvap(updatedSvap, originalSvap).stream()
+                .forEach(crit -> {
+                    List<CertifiedProductDetailsDTO> listings = svapDao.getCertifiedProductsBySvapAndCriteria(originalSvap, crit);
+                    if (listings.size() > 0) {
+                        messages.add(errorMessageUtil.getMessage("svap.edit.deletedCriteria.listingsExist",
+                                crit.getNumber() + (crit.getTitle().indexOf("Cures Update") == -1 ? "" : " (Cures Update)"),
```
kekey1: I think there is a criteria number formatter somewhere that does this sort of thing and we use in lots of the validators where we are printing out criterion numbers. Can you dig that up and use it here?

kekey1 (changes requested):
I had a few comments. I think I was also expecting to see somewhere that the new certification_criterion_attribute table was used to indicate whether a criterion could have svap or not. I see it used in the controller but I'm not sure if I missed it in the place I was expecting to see it or I am misunderstanding what was to be done or what. Maybe needs a demo/discussion?

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductDetailsManager.java
tmy1313: The data changes when an SVAP is editted/added/removed.  Not very often.  But if a user makes a change, this does not get updated.  We cannot call the `init` method when we change the SVAP because of  the multiple servers.  If we make the change on Server1, we cannot call the `init` method on Server2.
This pained me, too!

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SvapController.java
```diff
@@ -0,0 +1,75 @@
...
+                    + "Security Restrictions: To update: ROLE_ADMIN or ROLE_ONC.")
+    @RequestMapping(value = "", method = RequestMethod.PUT, consumes = MediaType.APPLICATION_JSON_VALUE,
+        produces = "application/json; charset=utf-8")
+    public @ResponseBody Svap updateSvap(@RequestBody(required = true) Svap svap) throws EntityRetrievalException, ValidationException {
+        return svapManager.update(svap);
+    }
+
+    @ApiOperation(value = "Create an SVAP.",
+            notes = "Provides functionality to add a new SVAP and the Criteria associated with it."
+                    + "Security Restrictions: To update: ROLE_ADMIN or ROLE_ONC.")
```
kekey1: I think there is maybe a copy/paste error in the notes although I could just be reading it incorrectly.  The first three methods here all say "To update:" but only the first method is doing an update.

kekey1 (dismissed):

andlar (dismissed):

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -689,3 +689,11 @@ realWorldTesting.report.missingPlansWarning:WARNING: Listings eligible for Real
...
 realWorldTesting.report.missingResultsWarning:WARNING: Listings eligible for Real World Testing in %s must have a Results URL by %s.
 realWorldTesting.report.missingResultsError:ERROR: Listings eligible for Real World Testing in %s must have a Results URL by %s.
+
+svap.delete.listingsExist=This SVAP is attested to by %s Listings: %s
+svap.edit.deletedCriteria.listingsExist=This SVAP and Criteria %s is attested to by %s Listings: %s
+svap.edit.noCriteria=An SVAP requires at least one Criterion associated with it.
+svap.edit.emptyApprovedStandardVersion=Approved Standard Version is required.
+svap.edit.emtptyRegulatoryTextCitation=Regulatory Text Citation is required.
+svap.edit.duplicate=Regulatory Text Citation and Approved Standard Version matches an existing SVAP.
+
```
andlar: We have a tool that's finding these things. Are we going to address them?

tmy1313: Sorry...  I messed up.

andlar (approved):

chpl/chpl-resources/src/main/resources/errors.properties
andlar: Not worried about messing up, was just wondering if we had been planning on addressing the auto-generated suggestions or not. I'm not going to hold things up based on those auto-generated ones, unless we decide we want to

kekey1 (approved):

PR #1079, opened 2021-02-17
OCD-3328: Remove API Key Usage report

andlar:


ashwinimore (dismissed):
Thank you for the consolidated newman script!

athomas-git (dismissed):

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/ApiKeyActivityDAO.java
```diff
@@ -1,33 +1,113 @@
...
+            }
+            if (dto.getCreationDate() != null) {
+                entity.setCreationDate(dto.getCreationDate());
+            } else {
+                entity.setCreationDate(new Date());
+            }
+            entity.setDeleted(dto.getDeleted());
+
+            if (AuthUtil.getCurrentUser() == null) {
+                entity.setLastModifiedUser(-3L);
```
tmy1313: I think we can User.SYSTEM_USER_ID.

tmy1313 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/ApiKeyActivityDAO.java
andlar: Good catch. If you're okay with it, I'll make that change before this goes to STG, but not hold up the progress across the board based on that change. Suggest leaving this conversation unresolved until that change is made though

ashwinimore (approved):

athomas-git (approved):

tmy1313 (approved):

kekey1 (approved):

PR #1085, opened 2021-03-01
OCD-3555: Allow questionable URL filtering by status code and ACB

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/urlStatus/QuestionableUrlReportGenerator.java
```diff
@@ -5,423 +5,258 @@
...
-    @Autowired
-    private JpaTransactionManager txnMgr;
+    private FailedUrlCsvFormatter csvFormatter = new FailedUrlCsvFormatter();
 
     @Override
     public void execute(JobExecutionContext jobContext) throws JobExecutionException {
         SpringBeanAutowiringSupport.processInjectionBasedOnCurrentContext(this);
         LOGGER.info("********* Starting the Questionable URL Report Generator job. *********");
 
-        TransactionTemplate txnTemplate = new TransactionTemplate(txnMgr);
```
andlar, line 85: Is this transaction no longer needed? I thought we had to wrap something in these execute methods in a transaction for some reason, but I'm not sure what that was. Maybe only for ones that are saving data?

kekey1: This code was all wrapped in a transaction because it called a bunch of DAO methods in other classes that were not marked Transactional (not, as far as I could tell, because it all has to be completed within a single transaction for data integrity). I refactored all of those calls into a separate DAO - QuestionableUrlLookupDao that is used by this job and does have Transactional annotations.

andlar (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/urlStatus/data/UrlResult.java
```diff
@@ -45,47 +51,12 @@ public boolean equals(Object obj) {
...
         return false;
     }
+
     @Override
     public int hashCode() {
         if (StringUtils.isEmpty(this.getUrl()) || this.getLastChecked() == null) {
             return -1;
         }
         return this.getUrl().hashCode() + this.getLastChecked().hashCode();
     }
```
tmy1313, line 61: The `@Data` lombok annotation generates a `hashCode()` method.  Could we just use that instead of this one, or is there something special about this method?

tmy1313 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/urlStatus/data/UrlResult.java
kekey1: I don't know.. I didn't change any functionality in this class other than moving it into a different package to better match the structure we follow and bring it up-to-date with our lombok Data, NoArgsConstructor, etc usage. This class is used in the "data collection" job which fills in the db with data that the Emailer job uses and the ticket was only to change the behavior of the Emailer job. I tried to bring all these related classes more in line with our coding standards without changing their behavior and I would rather not remove these equals and hashCode implementations since they do something other than the default behavior of comparing every field that Lombok would have them do. I think if we need/want to get rid of these methods then we'd need to regression test the data collection job as well.

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/urlStatus/QuestionableUrlReportGenerator.java
```diff
@@ -131,11 +136,12 @@ public int compare(final FailedUrlResult o1, final FailedUrlResult o2) {
...
 
     private List<FailedUrlResult> filterUrls(List<FailedUrlResult> badUrls, JobExecutionContext jobContext) {
-        String acbSpecific = jobContext.getMergedJobDataMap().getString("acbSpecific");
-        if (!StringUtils.isEmpty(acbSpecific) && BooleanUtils.toBoolean(acbSpecific)) {
+        if (isAcbSpecific(jobContext)) {
             List<Long> acbIds = getSelectedAcbIds(jobContext);
             return badUrls.stream()
                 .filter(badUrl -> isUrlRelatedToAcbs(badUrl, acbIds))
+                .filter(badUrl -> isNotListingUrl(badUrl) || isUrlRelatedTo2015Edition(badUrl))
+                .filter(badUrl -> isNotListingUrl(badUrl) || isUrlRelatedToActiveListing(badUrl))
```
andlar: Should these be combined into one?
```suggestion
                .filter(badUrl -> isNotListingUrl(badUrl) || isUrlRelatedTo2015Edition(badUrl) || isUrlRelatedToActiveListing(badUrl))
```

The line above is filtering based on ACBs based on the job definition, and the one below is about the business logic around the status codes. These two are both about what types of URLs should be in the report, which feels like a different check to the ACB & code checks, but they both feel like the same kind of check as each other, which is why I wonder if they can be combined into one line

andlar (approved):

tmy1313 (approved):

PR #1086, opened 2021-03-03
OCD-3580:  G1/G2 Cures Criteria Data Fix

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/measurefix/MeasureFixJob.java
```diff
@@ -0,0 +1,132 @@
...
+
+            CompletableFuture<Void> combinedFutures = CompletableFuture
+                    .allOf(futures.toArray(new CompletableFuture[futures.size()]));
+
+            // This is not blocking - presumably because the job executes using it's own ExecutorService
+            // This is necessary so that the system can indicate that the job and it's threads are still running
+            combinedFutures.get();
+            LOGGER.info("********* Completed the G1/G2 Measure Data Fix job. *********");
+        } catch (Exception e) {
+            LOGGER.catching(e);
```
kekey1, line 72: I did not know about this method! Just looked it up - I like it!

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/measurefix/MeasureFixJob.java
```diff
@@ -0,0 +1,132 @@
...
+        // The object's proxy is not called when the method is called from within this class. The object's proxy
+        // is called when the method is public and is called from a different object.
+        // https://stackoverflow.com/questions/3037006/starting-new-transaction-in-spring-bean
+        TransactionTemplate txTemplate = new TransactionTemplate(txManager);
+        txTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+        return txTemplate.execute(new TransactionCallback<Boolean>() {
+            @Override
+            public Boolean doInTransaction(TransactionStatus status) {
+                try {
+                    CertifiedProductSearchDetails listing = certifiedProductDetailsManager.getCertifiedProductDetails(cp.getId());
```
kekey1: It's not clear to me if/why this code needs to be wrapped in a transaction. It looks like it's calling a couple of manager methods which I expect are already marked as Transactional themselves. I don't think it would hurt anything to have this transaction here either, I'm just wondering if it's needed for another reason.

kekey1 (approved):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/measurefix/MeasureFixJob.java
```diff
@@ -0,0 +1,132 @@
...
+                    ListingUpdateRequest request = ListingUpdateRequest.builder()
+                            .listing(listing)
+                            .acknowledgeWarnings(true)
+                            .reason(null)
+                            .build();
+
+                    certifiedProductManager.update(request, false);
```
kekey1: I know we agreed not to log activity so we didn't have all those activities as a side-effect of this job. I'm also now thinking that this will change the lastUpdated date for each of these listings (even ones that are valid and won't have their g1/g2 measures changed at all) and I just want to make sure that we don't mind having that happen. Something to ask at the feature review maybe but I thought of it now and wanted to write it down.

kekey1 (dismissed):

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -648,5 +648,21 @@
                 </entry>
             </job-data-map>
         </job>
+
```
andlar: Our plan is to address these alerts, right? I don't see a reason to hold up the PR now, but maybe before it's merged?

andlar (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1098, opened 2021-03-23
OCD-3550: Basic listing validation for flexible uploads

kekey1:


chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -692,5 +692,21 @@
...
+            <name>listingUploadValidationJob</name>
+            <group>chplBackgroundJobs</group>
+            <description>Complete the work of building, validating, and saving error/warning counts of an uploaded listing.</description>
+            <job-class>gov.healthit.chpl.scheduler.job.ListingUploadValidationJob</job-class>
+            <durability>true</durability>
+            <recover>true</recover>
+            <job-data-map>
+                <entry>
+                    <key>authorities</key>
+                    <value>ROLE_ADMIN;ROLE_ONC;ROLE_ACB</value>
```
andlar: I don't think ROLE_ONC can upload listings

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertificationBodyDTO.java
```diff
@@ -49,70 +46,14 @@ public CertificationBodyDTO(final CertificationBody domain) {
...
-    public Date getRetirementDate() {
-        return retirementDate;
-    }
-
-    public void setRetirementDate(final Date retirementDate) {
-        this.retirementDate = retirementDate;
-    }
-
     @Override
     public String toString() {
```
andlar: Does one of the lombok annotations create a toString, and maybe an equals method?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/TestingLabDTO.java
```diff
@@ -43,75 +40,10 @@ public TestingLabDTO(final TestingLabEntity entity) {
...
-
-    public Date getRetirementDate() {
-        return retirementDate;
-    }
-
-    public void setRetirementDate(final Date retirementDate) {
-        this.retirementDate = retirementDate;
-    }
-
     @Override
```
andlar: Same toString question

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/handler/ListingDetailsUploadHandler.java
```diff
@@ -87,6 +90,7 @@ public CertifiedProductSearchDetails parseAsListing(CSVRecord headingRecord, Lis
                 .version(parseVersion(headingRecord, listingRecords))
                 .certificationEdition(editionHandler.handle(headingRecord, listingRecords))
                 .transparencyAttestationUrl(parseTransparencyAttestationUrl(headingRecord, listingRecords))
+                .transparencyAttestation(parseTransparencyAttestation(headingRecord, listingRecords))
```
andlar: Are we still allowing uploads of Transparency Attestations? I thought this feature was basically retired / obsolete

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/validation/ListingUploadValidator.java
```diff
@@ -0,0 +1,115 @@
...
+    public ListingUploadValidator(CSVHeaderReviewer csvHeaderReviewer,
+            ChplNumberFormatReviewer chplNumberFormatReviewer,
+            EditionCodeReviewer editionCodeReviewer,
+            TestingLabCodeReviewer atlCodeReviewer,
+            CertificationBodyCodeReviewer acbCodeReviewer,
+            DeveloperCodeReviewer developerCodeReviewer,
+            IcsCodeReviewer icsCodeReviewer,
+            AdditionalSoftwareCodeReviewer additionalSoftwareCodeReviewer,
+            EditionReviewer editionReviewer,
+            @Qualifier("testingLabReviewer") TestingLabReviewer atlReviewer,
```
andlar, line 65: Why is this the only one that has the `@Qualifier` annotation? I'm more curious than anything else...

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/validation/reviewer/DeveloperReviewer.java
```diff
@@ -0,0 +1,201 @@
...
+            listing.getWarningMessages().add(msgUtil.getMessage("listing.developer.userAndSystemMismatch", "website", developer.getUserEnteredWebsite(), developer.getWebsite()));
+        }
+        if (!StringUtils.isEmpty(developer.getUserEnteredSelfDeveloper())) {
+            Boolean userEnteredSelfDeveloper = null;
+            try {
+                userEnteredSelfDeveloper = uploadHandlerUtil.parseBoolean(developer.getUserEnteredSelfDeveloper());
+            } catch (Exception ex) {
+            }
+            if (userEnteredSelfDeveloper != null && !userEnteredSelfDeveloper.equals(developer.getSelfDeveloper())) {
+                listing.getWarningMessages().add(msgUtil.getMessage("listing.developer.userAndSystemMismatch", "self-developer", developer.getUserEnteredSelfDeveloper(), developer.getSelfDeveloper()+""));
```
andlar: Adding a comment here because I don't see the checkstyle alerts in the main discussion section of the PR

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/CertificationDateReviewer.java
```diff
@@ -18,7 +17,14 @@ public CertificationDateReviewer(ErrorMessageUtil msgUtil) {
...
     public void review(CertifiedProductSearchDetails listing) {
-        if (listing.getCertificationDate() > new Date().getTime()) {
+        if (listing.getCertificationDate() == null && StringUtils.isEmpty(listing.getCertificationDateStr())) {
+            listing.getErrorMessages().add(msgUtil.getMessage("listing.certificationDateMissing"));
+        } else if (listing.getCertificationDate() == null && !StringUtils.isEmpty(listing.getCertificationDateStr())) {
+            listing.getErrorMessages().add(msgUtil.getMessage("listing.badCertificationDate", listing.getCertificationDateStr()));
+        }
+
+        if (listing.getCertificationDate() != null && listing.getCertificationDate() > System.currentTimeMillis()) {
+            //TODO: print out the certification date
```
andlar: This is a newly added TODO, right? Should we do something with it?

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/handler/ListingDetailsUploadHandler.java
kekey1: Right. What happened is that I tested with all the currently active templates - v18,19,20. When I tested with v18 I found that there was a column I didn't handle (the transparency attestation one) so I added code for it not realizing that the "remove v18 template" ticket was about to show up on the board. Will undo this change.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertificationBodyDTO.java
kekey1: I'm going to add Lombok toStrings but I'm hesitant to do that with equals/hashcode because we aren't using all the fields to compare against.

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/validation/ListingUploadValidator.java
kekey1: There is another class with the same name (different package) and without the Qualifier annotation Spring doesn't know which one I intend for it to use.

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/reviewer/CertificationDateReviewer.java
kekey1: I don't remember. I think I wanted to add the parsed certification date to the error message so I added the TODO but later thought this is an existing error message that users have been used to seeing (assuming anyone has ever gotten this error) so I decided not to change it. I'm going to remove the TODO.

tmy1313 (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/normalizer/DeveloperDetailsNormalizer.java
```diff
@@ -4,121 +4,91 @@
...
+    private void copyUserEnteredDeveloperValues(Developer userEnteredDev) {
+        if (userEnteredDev == null) {
+            return;
+        }
+        userEnteredDev.setName(userEnteredDev.getUserEnteredName());
+        Boolean selfDeveloper = null;
+        try {
+            selfDeveloper = uploadHandlerUtil.parseBoolean(userEnteredDev.getUserEnteredSelfDeveloper());
+        } catch (Exception ex) {
+        }
```
andlar, line 91: Should we do something with this Exception? Maybe default the `selfDeveloper` to false and log the error? Or tell them there was some bad value in this?

kekey1: I added a line to log when this happens. The normalizers are not a place where I want to add errors to the listing, hoping to only do that in the validators. I save the string value entered for this field in a transient property in the DeveloperDetailsHandler (place we do parsing). In the normalizer, if it's not something we can turn into a boolean value, the way it's written we will leave the field as null. When it gets into the validator there are checks about if it matches what's in the system for the existing developer (or if it's a new developer the field is required and we check then, too).

andlar (approved):

tmy1313 (approved):

PR #1099, opened 2021-03-24
[#OCD-3628] - feat!: Display "Informed ONC per 170.523" in Complaints report

andlar:


kekey1 (dismissed):

tmy1313 (dismissed):

tmy1313 (approved):

kekey1 (approved):
I approve - I'm assuming Jim requested the reorganization of columns and is happy with the change. But I just want to note that this will change how their past quarterly/annual reports look and we went to pretty great lengths to make sure the ones they would have already downloaded (2019) won't change in the future by making different versions of the reports. Should this have only been rearranged in 2020+ reports?

PR #1111, opened 2021-04-19
OCD-3654 - Improve listing detail system performance

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertificationResultDetailsDAO.java
```diff
@@ -21,6 +22,28 @@
...
+            + "LEFT OUTER JOIN FETCH crd.certificationResultTestFunctionalities crtf "
+            + "LEFT OUTER JOIN FETCH crtf.testFunctionality tf "
+            + "LEFT OUTER JOIN FETCH crd.certificationResultTestProcedures crtp "
+            + "LEFT OUTER JOIN FETCH crtp.testProcedure tp "
+            + "LEFT OUTER JOIN FETCH crd.certificationResultTestTools crtt "
+            + "LEFT OUTER JOIN FETCH crtt.testTool tt "
+            + "LEFT OUTER JOIN FETCH crd.certificationResultTestStandards crts "
+            + "LEFT OUTER JOIN FETCH crd.certificationResultAdditionalSoftware cras "
+            + "WHERE crd.deleted = false ";
+//            + "AND (crtd.deleted = false OR crtd.deleted IS NULL) "
```
kekey1: Remove the commented-out code if we don't need it, although maybe I am missing it but I'm surprised this is working. Although in testing it does seem to be. I commented to you previously that instead of having all the deleted checks you might use the @Where annotation on the entities. That would mean putting something like this 
    @Where(clause = "deleted <> 'true'")
on each of the joined entities (CertificationResultTestData, CertificationResultTestProcedure, etc) but I don't see that in CertificationResultDetailsEntity. See DeveloperEntity for a couple of examples of using that annotation. It basically makes the generated SQL look more like one would write it with multiple conditions on each join (JOIN ... ON table1.id = table2.fk AND table1.deleted= false).

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/listing/CertificationResultDetailsEntity.java
```diff
@@ -233,4 +264,52 @@ public CertificationCriterionEntity getCertificationCriterion() {
     public void setCertificationCriterion(final CertificationCriterionEntity certificationCriterion) {
         this.certificationCriterion = certificationCriterion;
     }
+
+    public Set<CertificationResultTestDataEntity> getCertificationResultTestData() {
```
kekey1: Could we use @Data instead?

chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/questionableActivity/QuestionableActivityCertificationResultEntity.java
```diff
@@ -70,98 +70,122 @@
     @Column(name = "last_modified_date", insertable = false, updatable = false)
     private Date lastModifiedDate;
 
+    @Override
```
kekey1: I'm not following why these Overrides are needed. Can we just use @Data?

chpl/chpl-service/src/main/java/gov/healthit/chpl/certifiedproduct/service/ListingService.java
```diff
@@ -0,0 +1,301 @@
...
+                .measures(listingMeasureService.getCertifiedProductMeasures(dto.getId(), false))
+                .targetedUsers(getCertifiedProductTargetedUsers(dto.getId()))
+                .accessibilityStandards(getCertifiedProductAccessibilityStandards(dto.getId()))
+                .rwtPlansUrl(dto.getRwtPlansUrl())
+                .rwtPlansCheckDate(dto.getRwtPlansCheckDate())
+                .rwtResultsUrl(dto.getRwtResultsUrl())
+                .rwtResultsCheckDate(dto.getRwtResultsCheckDate())
+                .rwtEligibilityYear(dto.getRwtEligibilityYear())
+                .svapNoticeUrl(dto.getSvapNoticeUrl())
+                //.ics(getInheritedCertificationStatus(dto.getIcs()))
```
kekey1: Is this needed or just remove it?

chpl/chpl-service/src/main/java/gov/healthit/chpl/certifiedproduct/service/ListingService.java
```diff
@@ -0,0 +1,301 @@
...
+        if (!StringUtils.isEmpty(dto.getChplProductNumber())) {
+            return dto.getChplProductNumber();
+        } else {
+            return dto.getYearCode() + "." + dto.getTestingLabCode() + "." + dto.getCertificationBodyCode() + "."
+                    + dto.getDeveloper().getDeveloperCode() + "." + dto.getProductCode() + "." + dto.getVersionCode()
+                    + "." + dto.getIcsCode() + "." + dto.getAdditionalSoftwareCode() + "." + dto.getCertifiedDateCode();
+        }
+    }
+
+    private Map<String, Object> getCertifificationEdition(CertifiedProductDetailsDTO dto) {
```
kekey1:
```suggestion
    private Map<String, Object> getCertificationEdition(CertifiedProductDetailsDTO dto) {
```

chpl/chpl-service/src/main/java/gov/healthit/chpl/certifiedproduct/service/ListingService.java
```diff
@@ -0,0 +1,301 @@
...
+        return certEdition.orElse(null);
+    }
+
+    //private InheritedCertificationStatus getInheritedCertificationStatus(boolean ics) {
+    //    return InheritedCertificationStatus.builder()
+    //            .inherits(ics)
+    //            .build();
+    //}
+
+ // This should probably be refactored to use ChplProductNumberUtil
```
kekey1: I would be supportive of doing this refactor!

kekey1 (changes requested):

chpl/chpl-resources/src/main/resources/environment.properties
```diff
@@ -216,3 +216,6 @@ realWorldTestingCriteriaKeys=criterion.170_315_b_1_old,\
 auditDataFilePath=SECRET
 auditDataRetentionPolicyInMonths=2
 ####################################
+
+
+useUpdatedCertifiedProductDetailsManager=false
```
kekey1: I don't think this is used anymore, is it? Can you also remove from the DEV bamboo plan?

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertificationResultDetailsDAO.java
tmy1313: I'll take a look at this.  This code was commented out as part of the fix for the activity problem I was looking at this weekend.  When all the child data was deleted from the cert results, the cert result would no longer appear in the query results...

tmy1313: This gets a little weird due to the left joins.  Since we are getting all of the cert results and associated collections in a single query, sometimes the all of the values from a table are null and we cannot evaluate the deleted flag.  I /feel/ like the commented code should work for the deleted and not found data, but it just doesn't seem to work.  This leads to having to check for null values when populating the domain collections in the CertificationResultService.  It's not the best solution, and now that I think about about it I'm not sure what we should do - we are using this same query "base" for all queries in that DAO, but I'm not sure we are properly handling deleted/null values.  I'm continuing to research.

chpl/chpl-service/src/main/java/gov/healthit/chpl/certifiedproduct/CertifiedProductDetailsManager.java
```diff
@@ -0,0 +1,128 @@
...
+@Component("certifiedProductDetailsManager")
+public class CertifiedProductDetailsManager {
+    private CertifiedProductSearchResultDAO certifiedProductSearchResultDAO;
+    private ListingService listingService;
+    private CqmResultsService cqmResultsService;
+    private CertificationResultService certificationResultService;
+    private ListingMeasuresService listingMeasuresService;
+    private CertificationStatusEventsService certificationStatusEventsService;
+
+    @SuppressWarnings({"checkstyle:parameternumber"})
```
andlar: Is this annotation still needed?

chpl/chpl-service/src/main/java/gov/healthit/chpl/certifiedproduct/service/CertificationResultService.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    public List<CertificationResultDetailsDTO> getCertificationResultDetailsDTOs(Long id) {
+        List<CertificationResultDetailsDTO> certificationResultDetailsDTOs = null;
+        certificationResultDetailsDTOs = certificationResultDetailsDAO.getAllCertResultsForListing(id);
+        return certificationResultDetailsDTOs;
+    }
+
+
+    @SuppressWarnings({"checkstyle:methodlength"})
```
andlar: What about this one?

chpl/chpl-service/src/main/java/gov/healthit/chpl/surveillance/report/builder2019/ActivitiesAndOutcomesWorksheetBuilder2019.java
```diff
@@ -19,32 +19,37 @@ public ActivitiesAndOutcomesWorksheetBuilder2019(SurveillanceReportManager repor
         super(reportManager, detailsManager, survManager, privilegedSurvDao);
     }
 
+    @Override
```
andlar, line 22: What changed to make these `@Override` annotations required?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertificationResultDetailsDAO.java
kekey1: Understood that it is challenging. Maybe a better example is to look at the SurveillanceDAO. That behaves very similarly in that it is a big HQL statement that left joins all the possible pieces of surveillance and sometimes some of those things exist but not all of them. Those entities that get left-join-fetched in the surveillance HQL all have the `@Where(clause = "deleted <> 'true'")` annotation where they are referenced in the entity files. That code has been around for so long unchanged that I'm thinking it works and may be a good model to follow.

chpl/chpl-service/src/main/java/gov/healthit/chpl/surveillance/report/builder2019/ActivitiesAndOutcomesWorksheetBuilder2019.java
tmy1313: `@Override` is never required but is considered a good programming practice.  I didn't intentionally make this change, but think that Eclipse did it as part of refactoring when the references were updated.

kekey1 (approved):

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertificationResultDetailsDTO.java
```diff
@@ -48,141 +57,47 @@ public CertificationResultDetailsDTO(final CertificationResultDetailsEntity enti
...
-    public void setUseCases(String useCases) {
-        this.useCases = useCases;
-    }
-
-    public String getApiDocumentation() {
-        return apiDocumentation;
-    }
+        if (entity.getCertificationResultTestData() != null) {
+            this.testData = entity.getCertificationResultTestData().stream()
+                    //.filter(e -> e != null && e.getDeleted() != null && !e.getDeleted())
```
andlar, line 63: Should this be here?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dto/CertificationResultDetailsDTO.java
```diff
@@ -48,141 +57,47 @@ public CertificationResultDetailsDTO(final CertificationResultDetailsEntity enti
...
+                    .map(e -> new CertificationResultTestDataDTO(e))
+                    .collect(Collectors.toList());
+        }
 
-    public void setApiDocumentation(final String apiDocumentation) {
-        this.apiDocumentation = apiDocumentation;
-    }
+        if (entity.getCertificationResultTestFunctionalities() != null) {
+            this.testFunctionality = entity.getCertificationResultTestFunctionalities().stream()
+                    //.filter(e -> e != null && e.getDeleted() != null && !e.getDeleted())
```
andlar, line 70: These too?

andlar (approved):

PR #1117, opened 2021-05-07
OCD-3505: Add AQA tests for ACB and ATL controllers

ashwinimore:


palakpatel311:
I feel there are many tests which are repetitive. for example: get ACBs by onc, admin, onc staff- I don't see a value to run this test by 3 users. Also, this test is covered by UI. We should only add API tests which are not covered by UI otherwise those will be duplicates.

palakpatel311 (commented):
I feel there are many repetitive tests for example- get ACBs by admin, onc, onc staff. Running same tests by these 3 users doesn't add any value. Also, most of this tests are covered by UI, we should only add tests in API if they are not covered by UI otherwise it will be all duplicate values.

palakpatel311 (approved):

athomas-git (approved):

PR #1130, opened 2021-06-09
OCD-3680: Upgrade to Spring 5

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/AsynchronousCacheInitialization.java
```diff
@@ -37,19 +37,22 @@ public AsynchronousCacheInitialization(CertificationIdManager certificationIdMan
...
     @Async
     @Transactional
-    public Future<Boolean> initializeSearchOptions() throws EntityRetrievalException {
+    public void initializeSearchOptions() throws EntityRetrievalException {
         LOGGER.info("Starting cache initialization for SearchViewController.getPopulateSearchData()");
         dimensionalDataManager.getSearchableDimensionalData(true);
         dimensionalDataManager.getSearchableDimensionalData(false);
+        CacheManager manager = CacheManager.getInstance();
+        Cache productNamesCache = manager.getCache(CacheNames.PRODUCT_NAMES);
+        System.out.println("PRODUCT NAMES Cache KEy size: " + productNamesCache.getKeys().size());
```
andlar:
```suggestion
        System.out.println("PRODUCT NAMES Cache Key size: " + productNamesCache.getKeys().size());
```

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ChangeRequestController.java
```diff
@@ -49,7 +52,7 @@ public ChangeRequestController(final ChangeRequestManager changeRequestManager,
     @RequestMapping(value = "/{changeRequestId}", method = RequestMethod.GET, produces = "application/json; charset=utf-8")
     public @ResponseBody ChangeRequest getChangeRequest(@PathVariable final Long changeRequestId) throws EntityRetrievalException {
         if (!ff4j.check(FeatureList.CHANGE_REQUEST)) {
-            throw new NotImplementedException();
+            throw new NotImplementedException(msgUtil.getMessage("notImplemented"));
```
andlar, line 55: I'm curious about this change. I see the message string in a properties file, but am wondering why it needs to be added here and why it couldn't be in the NotImplementedException class as part of the constructor

andlar (dismissed):

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ChangeRequestController.java
kekey1: Hmmm NotImplementedException is not our code, it's from an Apache Commons library. Spring 4 brought with it an older version of the apache commons library and we had a bunch of references to it (including here). Spring 5 no longer brings that old commons library in so everywhere we referenced commons.lang I had to change it to commons.lang3. FYI - something we need to be conscious of until this code gets on STG is to NOT reference that library in new code otherwise we'll have integration problems. Anyway, that library is where NotImplementedException comes from and in the lang3 - new version - of the library it no longer supports a default no-arg constructor. It is required to pass in a string to the constructor now. Given that information, I am not sure if I understand what change you're suggesting.

andlar: I was asking why we couldn't use the default constructor, more for my own edification than anything else, and given your explanation it's now clear. Thanks!

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1138, opened 2021-06-16
OCD-3431: feat!: Disable Merge ability for ROLE_ACB

andlar:


kekey1 (changes requested):
I think you also need to make a change to CreateBackgroundJobTriggerActionPermissions.... and the jobs.xml to not allow ACBs for the developer merge. Also, jobs.xml indicates ONC STAFF should be able to do it but I'm not sure the code in the permissions class matches that.

kekey1 (dismissed):

kekey1 (approved):

tmy1313 (approved):

PR #1156, opened 2021-07-27
OCD-3707: mark confirmed listings as RWT eligible, based on rules

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/entity/search/CertifiedProductBasicSearchResultEntity.java
```diff
@@ -107,11 +99,12 @@
     @Column(name = "meaningful_use_users")
     private Long meaningfulUseUserCount;
 
-    @Setter(AccessLevel.NONE)
-    @Getter(AccessLevel.NONE)
     @Column(name = "meaningful_use_users_date")
     private Date meaningfulUseUserDate;
 
+    @Column(name = "rwt_eligibility_year")
+    private Integer rwtEligibilityYear;
```
kekey1: I'm uncertain about this field being added here and in the data model. The search view that it was added to plus this class are used pretty much only to fill out the collections endpoint when you get all listings and also send back in the new /search/beta endpoint. If you add the field to the entity, I wonder if it needs to also be added to those domain objects and return it in those other places? Or does it not really need to be here?

kekey1 (dismissed):

tmy1313 (dismissed):

kekey1 (approved):

tmy1313 (approved):

PR #1164, opened 2021-08-05
OCD-3709:  Generate SVAP Summary CSV and allow download of that file and schema

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/DownloadableResourceController.java
```diff
@@ -236,4 +286,11 @@ public void downloadSurveillance(@RequestParam(value = "type", required = false,
         LOGGER.info("Downloading " + downloadFile.getName());
         fileUtils.streamFileAsResponse(downloadFile, "text/csv", response);
     }
+
+    @RequestMapping(value = "/listings-with-svap",
+    method = RequestMethod.GET,
+    produces = "application/json; charset=utf-8")
+    public @ResponseBody List<Long> getListingIdsWithSvap() {
```
andlar: Is this method used? Is the endpoint used? This seems like an odd method to have here

chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/history/ListingActivityHistoryHelper.java
```diff
@@ -0,0 +1,111 @@
...
+            return true;
+        } else if (orig == null && updated != null && !StringUtils.isBlank(updated.getSvapNoticeUrl())
+                && updated.getSvapNoticeUrl().equals(currentSvapNoticeUrl)) {
+            return true;
+        }
+        return false;
+    }
+
+    @Transactional
+    public CertifiedProductSearchDetails getListingOnDate(Long listingId, Date date) {
```
andlar: Would it make sense to move the SVAP helper functions into a different class? This method looks like a "ListingActivityHistoryHelper" method, but the other ones look very specific for a particular type of history

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/activity/history/ListingSvapActivityHistoryHelper.java
```diff
@@ -0,0 +1,84 @@
...
+
+import gov.healthit.chpl.dao.ActivityDAO;
+import gov.healthit.chpl.domain.CertifiedProductSearchDetails;
+import gov.healthit.chpl.domain.activity.ActivityConcept;
+import gov.healthit.chpl.dto.ActivityDTO;
+import lombok.extern.log4j.Log4j2;
+
+@Service
+@Log4j2
+public class ListingSvapActivityHistoryHelper {
```
tmy1313: I'm not a big fan of classes ending with "Helper".  I know I have added several classes to CHPL with this suffix, so you can take this with a grain of salt.  It is often considered a code smell.  This class looks like it get Activity based on some type of criteria (not a CHPL "Criteria"), and will/can be expanded in the future to support other criteria.  Maybe something like `ListingActivityHistoryEvent` - I'm not sure...  Naming is hard!!

Or maybe make a `ListingActivityHistory` interface.  Then make an implementation of 'LastSvapNoticeUrlListingActivityHistory` .  In this model, each "query would be it's own class, but we might have more control in the long run, by being able to run several of these "queries" in a generic way.

tmy1313 (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1166, opened 2021-08-09
[#OCD-3713] feat!: Restrict Listings to have only valid Test Procedures on edit

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/Edition2015ListingValidator.java
```diff
@@ -107,6 +108,10 @@
     @Qualifier("testStandardReviewer")
     private TestStandardReviewer tsReviewer;
 
+    @Autowired
+    @Qualifier("testProcedureReviewer")
+    private TestProcedureReviewer tpReviewer;
```
kekey1, line 113: Lines 315-342 in gov.healthit.chpl.validation.listing.reviewer.edition2015.RequiredData2015Reviewer are doing (I think) more or less what the reviewer you added here is doing. I'm curious if/why that existing code is not working? If we are replacing it with this reviewer, can that chunk of code be removed from the RequiredData2015Reviewer?

kekey1 (dismissed):

tmy1313 (dismissed):

tmy1313 (approved):

PR #1184, opened 2021-08-26
OCD-3657 - 2 different error message when splitting product which is owned by different ACB

tmy1313:


kekey1 (dismissed):

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -17,7 +17,6 @@ notImplemented=Not yet implemented.
 #acb errors
 acb.notFound=Certification body not found.
 acb.accessDenied=User %s does not have access to activity for certification body with ID %s.
-acb.accessDenied.listingUpdate=Access is denied to update listing %s because it is owned by %s.
```
andlar, line 20: Did this entire error go away? Is there still an error that shows up if Drummond edits an ICSA listing?

tmy1313: It did entirely go away - if the ACB does not have access to the all the developer's listings the ticket indicated that we should use the message used in the "ActionPermission" class.
From the ticket - 
image

andlar (dismissed):

kekey1 (approved):

andlar (approved):

PR #1187, opened 2021-09-08
OCD-3742 - Update overnight job and maint plan schedule

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/certifiedproduct/CertifiedProductDetailsManager.java
```diff
@@ -59,6 +61,12 @@ public CertifiedProductSearchDetails getCertifiedProductDetails(Long certifiedPr
         return listingService.createCertifiedSearchDetails(certifiedProductId);
     }
 
+    @Transactional(readOnly = true)
+    @Cacheable(value = CacheNames.LISTING_DETAILS)
```
kekey1, line 65: I really like the use of this cache and would like to use it in lots of other places when returning listing details to the user! Should you add some places where an entry in the cache will expire when that listing is updated? Not 100% sure how that works because we tend to just clear out entire caches vs doing it for only the one key/value pair that would have changed but I'm sure that is standard ehcache stuff. I know people probably aren't updating listings in the middle of the night when the jobs run but if it did happen I think we'd want to be using the correct data.

kekey1 (dismissed):
Also remember to add this new cache to the STG and PROD Bamboo plans! I think  new cache can be added at any time (i.e. now) without hurting anything.

kekey1 (dismissed):

chpl/chpl-resources/src/main/resources-dev/quartz.properties
```diff
@@ -5,7 +5,7 @@ org.quartz.jobStore.driverDelegateClass = org.quartz.impl.jdbcjobstore.PostgreSQ
 org.quartz.jobStore.class = org.quartz.impl.jdbcjobstore.JobStoreTX
 org.quartz.jobStore.dataSource = openchpl
 org.quartz.jobStore.tablePrefix = quartz.QRTZ_
-org.quartz.jobStore.isClustered = true
+org.quartz.jobStore.isClustered = false
```
andlar: Do we need to change these? I ask because if it doesn't hurt to leave it clustered I'd prefer to do that. Two reasons: one is just in case we deploy another appserver. The other is that it matches what's on STG/PROD, so we _might_ have an opportunity to catch something weird earlier in the process

tmy1313: Change made.  Not sure I 100% agree - I tend to think the config should match the environment in which it is running.  You are correct that it doesn't seem to have an adverse effect.  Not a hill worth dying on, though :)

kekey1 (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/Edition2015ListingValidator.java
```diff
@@ -221,9 +223,151 @@
     private List<ComparisonReviewer> comparisonReviewers;
 
     @Override
-    public List<Reviewer> getReviewers() {
+
+    public synchronized List<Reviewer> getReviewers() {
         if (reviewers == null) {
             reviewers = new ArrayList<Reviewer>();
+//            if (chplNumberReviewer == null) {
```
kekey1: Debug code?

chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/Validator.java
```diff
@@ -33,9 +30,19 @@
      * errors and warnings as appropriate.
      * @param listing the listing to validate
      */
-    public void validate(final CertifiedProductSearchDetails listing) {
+    public synchronized void validate(final CertifiedProductSearchDetails listing) {
         for (Reviewer reviewer : getReviewers()) {
-            reviewer.review(listing);
+            try {
+                if (reviewer != null) {
+                    //LOGGER.info("Running Reviewer: " + reviewer.getClass().getName());
```
kekey1: Debug code

kekey1 (changes requested):

kekey1 (approved):

andlar (approved):

PR #1193, opened 2021-09-17
[#OCD-3773] feat!: Remove "frequency" data from Scheduled Jobs

andlar:


tmy1313 (approved):

kekey1 (approved):
Do you need any pre-processing commands in the jobs.xml file? I see frequency is removed in the job definitions but I suppose it will be hanging around in the quartz tables. Maybe it doesn't matter and can be left there?

PR #1194, opened 2021-09-23
OCD-3142: Make certification date field derived and read-only.

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
```diff
@@ -165,10 +168,7 @@
     @Singular
     private List<CertifiedProductTestingLab> testingLabs = new ArrayList<CertifiedProductTestingLab>();
 
-    /**
-     * Certification date represented in milliseconds since epoch
-     */
-    @XmlElement(required = true)
+    @XmlTransient
```
andlar, line 171: Does this mean the certification date _won't_ be in the XML? Should this have the XSD-Impact component back on?

andlar (approved):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/CertifiedProductSearchDetails.java
kekey1: It is in the XML (I tested locally at least) but it comes from the method getCertificationDate() rather than from this member of the class. I couldn't remove the class member due to the million places it is used and that removal would be done in whatever ticket we actually make this a LocalDate (maybe). I think the main thing is it's used/parsed in the upload in a subclass of this and I can't not set it here without changing the way that part of the upload works.

tmy1313 (approved):

PR #1203, opened 2021-10-11
OCD-3734: Track and report of usage of API endpoints with deprecated response fields

kekey1:


chpl/chpl-api/src/test/java/gov/healthit/chpl/util/DeprecatedFieldExplorerTest.java
```diff
@@ -0,0 +1,869 @@
...
+        assertTrue(deprecatedFieldNames.contains("apiKey"));
+    }
+
+    @Test
+    public void findDeprecatedFields_CertificationIdResults() {
+        Set<String> deprecatedFieldNames = deprecatedFieldExplorer.getDeprecatedFieldsForClass(CertificationIdResults.class);
+
+        assertNotNull(deprecatedFieldNames);
+        assertEquals(1, deprecatedFieldNames.size());
+        assertTrue(deprecatedFieldNames.contains("isValid"));
```
andlar, line 256: I like having both the count and the list of expected ones. I feel like these tests will help find any unexpectedly deprecated elements

And I just looked to see if there were any that were expecting 0, and found some, so now the only thing that we would miss should be an entirely new response Class that isn't in this test

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

andlar (dismissed):

tmy1313 (dismissed):

PR #1205, opened 2021-10-15
OCD-3757 - Convert Test Standards to Approved Standards

tmy1313:


andlar (dismissed):

kekey1 (approved):

kekey1 (dismissed):
I'm curious if the xlsx mapping file could be co-located with the job in that "onetime" package you made? Or does it have to go under src/main/resources to be picked up? I was thinking from the perspective of deleting the job later it would be nice to look in just the one place.

andlar (approved):

kekey1 (approved):

PR #1206, opened 2021-10-18
OCD-3761 - Alert users when using a Username to login instead of email address

tmy1313:
OCD-3761

### DIFF
--- a/chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/auth/AuthenticationManager.java
+++ b/chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/auth/AuthenticationManager.java
@@ -60,6 +60,7 @@ public class AuthenticationManager {
         if (user != null && user.isPasswordResetRequired()) {
             throw new UserRetrievalException(msgUtil.getMessage("auth.changePasswordRequired"));
         }
+        logWhenUsername(credentials, user);
         return jwt;
     }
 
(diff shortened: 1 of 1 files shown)


chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/auth/AuthenticationManager.java
```diff
@@ -223,4 +224,10 @@ public String unimpersonateUser(User user) throws JWTCreationException, UserRetr
     MultipleUserAccountsException {
         return getJWT(getUserByNameOrEmail(user.getSubjectName()));
     }
+
+    private void logWhenUsername(LoginCredentials creds, UserDTO user) {
+        if (!creds.getUserName().equals(user.getEmail())) {
+            LOGGER.warn(String.format("The following user logged in with username: %s / %s", creds.getUserName(), user.getEmail()));
```
andlar: I think the variables are backwards. It looks like the error message would look like

> The following user logged in with username: username / sample@email.com

That suggests that the user is identified by `username` and has the username of `sample@email.com`

Could we use something like 

```suggestion
            LOGGER.warn(String.format("The user with email %s logged in using the username: %s ", creds.getEmail(), user.getUserName()));
```

to give us the phrase

> The user with email sample@email.com logged in using the username: username

kekey1 (approved):

andlar (approved):

PR #1210, opened 2021-10-28
OCD-3748 - Modify values in "Requirement" dropdown and "Non-Conformity Type" dropdown under surveillance activity

tmy1313:


andlar (dismissed):
This feature looks fine as is, but is there any way to add some unit tests? It seems like the specific Listing Activity "checkers" are small enough that we ought to be able to build at least stub listings that can verify that each of the activity checkers report on correct Questionable Activities. I know there are a lot of them, so maybe we don't need tests on all right now, but we could at least have tests for the new QA functionality that was added related to removed surveillance data elements, right?

kekey1 (dismissed):

andlar (commented):
Any thoughts on unit tests? We're now at the point where we either need to do them or not. I'd like to get at least one in the system

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #1212, opened 2021-11-05
OCD-3771: Update /search/beta endpoint to support RWT fields

kekey1:


chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -727,6 +728,8 @@ search.complianceFilter.unavailable=Compliance and non-conformity filtering is u
 search.hasHadSurveillance.invalid=Parameter value '%s' for hasHadSurveillance is invalid. It must be either 'true' or 'false'.
 search.hasHadComplianceActivity.invalid=Parameter value '%s' for hasHadComplianceActivity is invalid. It must be either 'true' or 'false'.
 search.nonconformitySearchOption.invalid=No non-conformity search option matches '%s'. Values must be one of %s.
+search.rwt.missingSearchOperator=Multiple RWT search options were found without a search operator (AND/OR). A search operator is required.
```
andlar, line 732: Is this error still used? I thought we defaulted to `OR` if this operator isn't found

chpl/chpl-service/src/main/java/gov/healthit/chpl/search/SearchRequestValidator.java
```diff
@@ -42,11 +46,18 @@ public SearchRequestValidator(DimensionalDataManager dimensionalDataManager,
         this.drService = drService;
         this.msgUtil = msgUtil;
         dateFormatter = DateTimeFormatter.ofPattern(SearchRequest.CERTIFICATION_DATE_SEARCH_FORMAT);
+        allowedDerivedCertificationEditions = Stream.of(CertificationEditionConcept.CERTIFICATION_EDITION_2011.getYear(),
+                CertificationEditionConcept.CERTIFICATION_EDITION_2014.getYear(),
+                CertificationEditionConcept.CERTIFICATION_EDITION_2015.getYear(),
+                CertificationEditionConcept.CERTIFICATION_EDITION_2015.getYear() + " Cures Update")
```
andlar: Should this use the `CURES_SUFFIX` constant?

chpl/chpl-resources/src/main/resources/errors.properties
kekey1: It's still used in the POST version of the endpoint. That one doesn't make use of the same default parameter values nor does it have the same notes about that in Swagger. If we need to refine that endpoint's behavior and documentation could/should that be a different ticket or do you think it needs done in this one?

chpl/chpl-service/src/main/java/gov/healthit/chpl/search/SearchRequestValidator.java
kekey1: pushed a fix for this, thanks

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductManager.java
```diff
@@ -1202,6 +1209,17 @@ private void updateListingsChildData(CertifiedProductSearchDetails existingListi
...
+        rwtCachingService.calculateRwtEligibility(listingId);
+
+        List<CertifiedProductDTO> listingChildren = listingGraphDao.getChildren(listingId);
+        if (!CollectionUtils.isEmpty(listingChildren)) {
+            for (CertifiedProductDTO child : listingChildren) {
+                updateRwtEligibilityForListingAndChildren(child.getId());
+            }
+        }
+    }
+
```
tmy1313: Could we maybe move this method or logic to the `RealWorldTestingEligiblityCachingService`.  That way if we need to perform this task outside this Manager, it would be available.

tmy1313 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/manager/CertifiedProductManager.java
kekey1: I am not sure how to move this and have it work... the method "calculateRwtEligibility" in the RealWorldTestingEligiblityCachingService has a CachePut annotation. This method is called recursively from itself so if it were all in the RealWorldTestingEligiblityCachingService class I think it would end up ignoring the CachePut annotation at because of Spring proxies. Am I thinking about it wrong?

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1213, opened 2021-11-05
OCD-3451 - ACBs should be able to edit Developer users

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/ResourcePermissions.java
```diff
@@ -358,6 +363,11 @@ public boolean hasPermissionOnUser(UserDTO user) {
         } else if (isUserRoleOncStaff()) {
             return getRoleByUserId(user.getId()).getAuthority().equalsIgnoreCase(Authority.ROLE_ONC_STAFF);
         } else if (isUserRoleAcbAdmin()) {
+            if (ff4j.check(FeatureList.ROLE_DEVELOPER)) {
```
kekey1: I see that the ticket has the flag for ROLE_DEVELOPER listed on it, but I wonder if this code really needs to be behind a flag? There shouldn't be any developer users in the system to edit until the flag is turned on anyway (maybe not totally true with our preloaded developer and AQA user accounts)

chpl/chpl-service/src/main/java/gov/healthit/chpl/permissions/ResourcePermissions.java
```diff
@@ -358,6 +363,11 @@ public boolean hasPermissionOnUser(UserDTO user) {
...
+                if (getRoleByUserId(user.getId()).getAuthority().equalsIgnoreCase(Authority.ROLE_DEVELOPER)) {
+                    return true;
```
kekey1: I think the comment at the top of this method needs an update based on this change or possibly just removed. Can we add any unit tests?

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1215, opened 2021-11-15
OCD-3812: Create two "snapshot" jobs to present a listing and/or criteria as it was at a certain time

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/presenter/CertifiedProductCsvPresenter.java
```diff
@@ -147,18 +148,21 @@ public void setFf4j(FF4j ff4j) {
                 : listing.getDeveloper().getWebsite());
         result.add(formatSelfDeveloper(listing));
         result.addAll(getContactCells(listing));
-        result.add(listing.getProduct().getName());
-        result.add(listing.getProduct().getProductId().toString());
-        result.add(listing.getVersion().getVersion());
-        result.add(listing.getVersion().getVersionId().toString());
+        result.add(listing.getProduct() != null ? listing.getProduct().getName() : "?");
```
andlar: What about using a constant for the `?` at the end of this ternary? That way, if we _do_ need to change it for some reason it'll be easy to do it in one place. Especially if this ends up becoming something that could be seen by the public

andlar (approved):

tmy1313 (approved):

PR #1231, opened 2021-12-07
OCD-3823: Use new notification_date field of deprecated api usage entities

kekey1:
[#OCD-3823]


chpl/chpl-service/src/main/java/gov/healthit/chpl/api/deprecatedUsage/DeprecatedResponseFieldApiUsageDao.java
```diff
@@ -66,9 +66,10 @@ public void createOrUpdateUsage(DeprecatedResponseFieldApiUsage apiUsage) {
     }
 
     @Transactional
-    public void delete(Long id) {
+    public void markAsUserNotified(Long id) {
         DeprecatedResponseFieldApiUsageEntity entity = entityManager.find(DeprecatedResponseFieldApiUsageEntity.class, id);
         if (entity != null) {
+            entity.setNotificationSent(new Date());
             entity.setDeleted(true);
```
andlar: Should these be marked deleted? If we're marking them as sent I'm not sure why we mark them as deleted too

kekey1: "deleted" is there because it's there on every table... when we sent the notification we mark it deleted. But the notification_sent date can be used in a unique constraint whereas "deleted" cannot. Open to other suggestions.

andlar: Keeping the column there makes sense, but I'm not sure why we're marking the row as deleted when a message is sent. We usually use "deleted" to flag that a thing in the database isn't really valid data any more, but I feel like these rows are more of an indication of what messages had been sent in the past, and marking them as deleted is sort of like saying they were invalid, or didn't get sent, or were otherwise bad data

andlar (approved):

tmy1313 (approved):

PR #1244, opened 2021-12-21
OCD-3808 - Re-send confirmation email on login if user hasn't confirmed

tmy1313:


andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/ChplHttpStatus.java
```diff
@@ -0,0 +1,48 @@
+package gov.healthit.chpl;
+
+import org.springframework.lang.Nullable;
+
+public enum ChplHttpStatus {
+    RESENT_USER_CONFIRMATION_EMAIL(461, "Account's Email Not Confirmed");
```
kekey1, line 6: All looks good to me. My only comment is that some flavor of this message is in a few different places and I wonder if it could/should be consolidated. It's in the UI, it's here, it's in the errors.properties file, and it's in the ApiExceptionControllerAdvice file. They all appear to be just slight variations of each other.

kekey1 (dismissed):

kekey1 (dismissed):

andlar (dismissed):

andlar (dismissed):

andlar (approved):

kekey1 (approved):

PR #1252, opened 2022-01-13
OCD-3831 : API aqa tests for search controller using removed isEligible parameter…

athomas-git:
OCD-3831 : API aqa tests for search controller using removed isEligible parameter

### DIFF
--- a/chpl/chpl-api/e2e/collections/search-controller.postman_collection.json
+++ b/chpl/chpl-api/e2e/collections/search-controller.postman_collection.json
@@ -322,6 +322,118 @@
 				"description": "GET /search/beta request using IS_ELIGIBLE filter should return 400 status and error"
 			},
 			"response": []
+		},
+		{
+			"name": "POST /search/beta - 200 status using valid rwt  parameters",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"POST /search/beta end point should return 200 status and correct response when valid rwt filters are used\", function () {\r",
+							"    var jsonData = pm.response.json();\r",
+							"        pm.response.to.have.status(200);\r",
+							"        pm.expect(jsonData.results[0]).to.have.property('rwtPlansUrl');\r",
+							"        pm.expect(jsonData.results[0]).to.have.property('rwtResultsUrl');\r",
+							"});"
+						],
+						"type": "text/javascript"
+					}
+				}
+			],
+			"request": {
+				"method": "POST",
+				"header": [
+					{
+						"key": "API-KEY",
+						"value": "{{apiKey}}",
+						"type": "text"
+					},
+					{
+						"key": "Content-Type",
+						"value": "application/json",
+						"type": "text"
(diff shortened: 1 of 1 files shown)


kekey1 (dismissed):

chpl/chpl-api/e2e/collections/search-controller.postman_collection.json
```diff
@@ -237,6 +257,130 @@
...
+						],
+						"type": "text/javascript"
+					}
+				},
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"POST /search/beta request using IS_ELIGIBLE filter should return 500 status and error\", function () {\r",
+							"    pm.response.to.have.status(500);\r",
```
ashwinimore: IS_ELIGIBLE parameter is being removed and I think the aqa test to check what it does on next line (#279) is good but running a test every day to test that the request with removed parameter returns 500 does not add value. Same comment for test at line #320

ashwinimore (changes requested):

kekey1 (dismissed):

chpl/chpl-api/e2e/collections/search-controller.postman_collection.json
```diff
@@ -237,6 +257,71 @@
 				"description": "POST /search/beta - filters other than compliance filters - should return 200 status"
 			},
 			"response": []
+		},
+		{
+			"name": "GET /search/beta -using IS_ELIGIBLE should return 400 status",
```
ashwinimore, line 262: How is this test different than the one that was removed expect that was POST and this is GET but both were written originally to make a request with removed parameter and expect bad response and this one still does that? I was suggesting to keep the removed check to expect valid parameters in response - 
 pm.expect(pm.response.text()).to.include(\"Values must be one of HAS_PLANS_URL,HAS_RESULTS_URL,NO_PLANS_URL,NO_RESULTS_URL\");\r", 
 remove check for 'expecting bad response' and change the test to say something like GET /search/beta endpoint returns relevant/expected parameters in response...

ashwinimore (changes requested):

chpl/chpl-api/e2e/collections/search-controller.postman_collection.json
ashwinimore: I would recommend keeping the test for POST but for positive scenario

chpl/chpl-api/e2e/collections/search-controller.postman_collection.json
athomas-git: Yes, this test is different from the other because for GET request passing removed parameters gives 400 status and for POST with removed parameter , the expected status is 500 which is the one I removed . I would like to leave the GET request with removed parameter test as is and I can add another test for POST with valid parameters

palakpatel311 (approved):

kekey1 (approved):

ashwinimore (approved):

PR #1253, opened 2022-01-20
OCD-3759: Upgrade to Java 17

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/ChplFF4JProvider.java
```diff
@@ -18,7 +16,9 @@ public ChplFF4JProvider() {
 
     @Override
     public FF4j getFF4j() {
+//        if (ff4j == null) {
```
andlar: Should this be removed? Or does it need to be restored?

chpl/pom.xml
```diff
@@ -42,55 +61,128 @@
...
+              <artifactId>maven-enforcer-plugin</artifactId>
+              <executions>
+                <execution>
+                  <id>enforce-versions</id>
+                  <goals>
+                    <goal>enforce</goal>
+                  </goals>
+                  <configuration>
+                    <rules>
+                      <!-- TODO: put this back in -->
```
andlar, line 162: Is this TODO for now or later?

andlar (dismissed):
Approving to go to QA; comments for things that are not critical

chpl/chpl-api/src/main/java/gov/healthit/chpl/ChplFF4JProvider.java
kekey1: Removed it.

chpl/pom.xml
kekey1: This one I don't know. This rule was in "chpl-service" and I moved it up to the parent because it seems like it could/should apply across the projects. I couldn't get it to pass though so I added a note to avoid forgetting about and promptly forgot about it! We could discuss among the dev team if it's important enough to resolve before moving to STG.

andlar (approved):

tmy1313 (approved):

PR #1255, opened 2022-01-27
OCD-3717 - Add new Attestation submission process

tmy1313:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/DeveloperController.java
```diff
@@ -313,6 +318,21 @@ public PermissionDeletedResponse deleteUserFromDeveloper(
...
+    @Operation(summary = "List attestations for a developer.",
+            description = "Security Restrictions: ROLE_ADMIN, ROLE_ONC, ROLE_ACB, or have administrative "
+                    + "authority on the specified developer.",
+            security = {
+                    @SecurityRequirement(name = SwaggerSecurityRequirement.API_KEY),
+                    @SecurityRequirement(name = SwaggerSecurityRequirement.BEARER)
+            })
+    @RequestMapping(value = "/{developerId}/attestations", method = RequestMethod.GET, produces = "application/json; charset=utf-8")
+    public @ResponseBody DeveloperAttestationSubmissionResults getAttestations(@PathVariable("developerId") Long developerId) throws InvalidArgumentsException, EntityRetrievalException {
+        if (!ff4j.check(FeatureList.ROLE_DEVELOPER)) {
```
andlar: ROLE_DEVELOPER is turned on. Should this check against ATTESTATION?

chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/domain/AttestationValidResponse.java
```diff
@@ -0,0 +1,29 @@
...
+import gov.healthit.chpl.attestation.entity.AttestationValidResponseEntity;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Getter
+@Setter
```
andlar: Is there a reason to use these instead of the `@Data` annotation?

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/validation/ChangeRequestValidationContext.java
```diff
@@ -1,16 +1,161 @@
...
+        this.resourcePermissions = resourcePermissions;
+        this.validationUtils = validationUtils;
+        this.validationDAOs = new ValidationDAOs(developerDAO, changeRequestDAO, changeRequestStatusTypeDAO, changeRequestTypeDAO);
+        this.domainManagers = new DomainManagers(attestationManager);
+        this.changeRequestStatusIds = new ChangeRequestStatusIds(cancelledStatus, acceptedStatus, rejectedStatus, pendingAcbActionStatus, pendingDeveloperActionStatus);
+        this.changeRequestTypeIds = new ChangeRequestTypeIds(websiteChangeRequestType, developerDetailsChangeRequestTypeId);
+    }
+
+
+    public static class ValidationDAOs {
```
andlar, line 62: Can these classes have `@Data` annotations, or something like them? Or does that not work? Not a required change, but I'm curious

andlar (changes requested):
The ticket has XSD-Impact on it. I think the object that includes "period - data submitted" is supposed to be included in the Developer domain objects as part of this ticket

The "attestation" enum changed to be a "transparency_attestation" enum. Has that been updated here?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/DeveloperController.java
kekey1: Agree with not checking against ROLE_DEVELOPER - that will not compile with the ticket coming up where we remove the ROLE_DEVELOPER flag.

chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/domain/AttestationPeriod.java
```diff
@@ -0,0 +1,49 @@
...
+    @JsonSerialize(using = LocalDateSerializer.class)
+    private LocalDate submissionStart;
+
+    @JsonDeserialize(using = LocalDateDeserializer.class)
+    @JsonSerialize(using = LocalDateSerializer.class)
+    private LocalDate submissionEnd;
+
+    private String description;
+
+    public AttestationPeriod(AttestationPeriodEntity entity) {
```
kekey1, line 41: Should we have a discussion about where we are putting the domain object creation code? I have been liking thinking of the domain objects as things that could be in their own jar and handed off to somebody (they aren't right now in reality due to dependencies on DTOs and entities). So I have been putting methods in the entity classes like "toDomain()" that uses the entity instance to create the domain object. And we clearly are thinking opposite ways about that since the domain objects here are referencing the entity. I don't care for now but would like to agree on what to do moving forward.

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/validation/ChangeRequestModificationValidation.java
```diff
@@ -32,9 +28,9 @@ private Boolean changeRequestDetailsEquals(ChangeRequestValidationContext contex
         } else if (context.getNewChangeRequest().getDetails() instanceof ChangeRequestDeveloperDetails) {
             return ChangeRequestDeveloperDetails.cast(context.getNewChangeRequest().getDetails()).matches(
                     ChangeRequestDeveloperDetails.cast(context.getOrigChangeRequest().getDetails()));
-        } else if (context.getNewChangeRequest().getDetails() instanceof ChangeRequestAttestation) {
-            return ChangeRequestAttestation.cast(context.getNewChangeRequest().getDetails()).matches(
-                    ChangeRequestAttestation.cast(context.getOrigChangeRequest().getDetails()));
+//        } else if (context.getNewChangeRequest().getDetails() instanceof ChangeRequestAttestation) {
```
kekey1, line 31: can this be removed?

chpl/chpl-service/src/main/java/gov/healthit/chpl/changerequest/validation/ChangeRequestValidationContext.java
tmy1313: At one point these inner classes were `private` and the Lombok annotations don't work.  At some point I switched them to be `public` and just never removed the `getters` and added `@Data`.  I have made this change.

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

andlar (approved):

kekey1 (approved):

PR #1257, opened 2022-01-29
feat: Remove developer, rwt, and cures+18month flags

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -254,11 +249,6 @@ public String authorizeUser(@RequestBody AuthorizeCredentials credentials)
     public UserInvitation inviteUser(@RequestBody UserInvitation invitation)
             throws InvalidArgumentsException, UserCreationException, UserRetrievalException,
             UserPermissionRetrievalException, AddressException, EmailNotSentException {
-
-        if (!ff4j.check(FeatureList.ROLE_DEVELOPER) && invitation.getRole().equals(Authority.ROLE_DEVELOPER)) {
-            throw new NotImplementedException(msgUtil.getMessage("notImplemented"));
```
andlar, line 259: Are there still any calls to this `notImplemented` message and/or NotImplementedException? Wondering because if there aren't maybe we should remove the message and/or maybe the NotImplementedException

andlar (approved):

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
kekey1: Yes and yes! They are still called in a number of places.

tmy1313 (approved):

PR #1268, opened 2022-02-23
OCD-3790 - Add controls to grant and limit access to Attestation submission process

tmy1313:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/DeveloperController.java
```diff
@@ -338,6 +343,21 @@ public PermissionDeletedResponse deleteUserFromDeveloper(
         return new PublicDeveloperAttestestationResults(developerManager.getDeveloperPublicAttestations(developerId));
     }
 
+    @Operation(summary = "Create a new attestation submission end date exception for a developer.",
+            description = "Security Restrictions: ROLE_ADMIN or ROLE_ONC",
```
andlar: Pretty sure ROLE_ONC_ACB should be able to hit this endpoint

chpl/chpl-resources/src/main/resources/errors.properties
```diff
@@ -908,3 +909,5 @@ svap.edit.emtptyRegulatoryTextCitation=Regulatory Text Citation is required.
 svap.edit.duplicate=Regulatory Text Citation and Approved Standard Version matches an existing SVAP.
 
 apiKeyRequest.notFound=API Key Request not found.
+
+attestation.submissionPeriodException.cannotCreate=Cannot create Submission Period Exception for developer.
```
andlar: Can we add a newline to the end of this line?

chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/entity/AttestationPeriodDeveloperExceptionEntity.java
```diff
@@ -0,0 +1,61 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Getter
+@Setter
+@ToString
```
andlar, line 26: I know there are thoughts about Lombok annotations, but `@Data` would eliminate these three, and add `@EqualsAndHashCode` which I think we end up needing sometimes. Is it worth talking about making a choice betwene either `@Data` and no specific ones, or eliminating the various "combination" annotations from Lombok, including `@Data`?

chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/manager/AttestationManager.java
```diff
@@ -57,4 +74,92 @@ public DeveloperAttestationSubmission saveDeveloperAttestation(DeveloperAttestat
...
+        } else {
+            AttestationPeriod mostRecentPastAttestationPeriod = attestationPeriodService.getMostRecentPastAttestationPeriod();
+            if (withinStandardSubmissionPeriod(mostRecentPastAttestationPeriod)) {
+                if (doesAttestationForDeveloperExist(developerId, mostRecentPastAttestationPeriod.getId())) {
+                    return !doesPendingAttestationChangeRequestForDeveloperExist(developerId);
+                } else {
+                    return false;
+                }
+            } else {
+                return !doesPendingAttestationChangeRequestForDeveloperExist(developerId);
```
andlar, line 106: I feel like the "withinStandardsSubmissionPeriod" part is unnecessary.

If we're in this block the Developer can't submit one, so the only reason an admin couldn't create an exception is if the Developer couldn't submit one due to a pending CR. I don't think there's anything in that logic that depends on the "standards submission period"

I feel like I need a flowchart on this, but would like to investigate the states that get us to each of these branches, as it'd be nice to prune out paths that we might not actually need

tmy1313: I think we do need the `withinStandardSubmissionPeriod` check...  

In the standard submission window, an ACB should only be able to create an exception when an attestation exists for the developer.  Otherwise, if the ACB were to create an exception, before the attestation for the developer has been accepted, it could make the submission window period shorter than the standard submission window when an attestation does not exist.

I know this is hard to explain and maybe we should discuss.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/DeveloperController.java
```diff
@@ -344,7 +344,7 @@ public PermissionDeletedResponse deleteUserFromDeveloper(
     }
 
     @Operation(summary = "Create a new attestation submission end date exception for a developer.",
-            description = "Security Restrictions: ROLE_ADMIN or ROLE_ONC",
+            description = "Security Restrictions: ROLE_ADMIN, ROLE_ONC, or ROLE_ACB_ADMIN",
```
andlar:
```suggestion
            description = "Security Restrictions: ROLE_ADMIN, ROLE_ONC, or ROLE_ONC_ACB",
```

chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/manager/AttestationManager.java
andlar: I agree with that logic, but on line 95 we check to see if a Developer can submit an attestation, and if so, return false from this method, never getting to this point. So if we're in the standard window, then a Developer can already create one, unless one is pending, or one already is published. In the first case the ACB can't create an attestation, which this line checks). In the second the ACB can, which this line also verifies. No logic depends on the standard window for either of those checks

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/dao/AttestationDAO.java
```diff
@@ -8,9 +8,11 @@
 
 import gov.healthit.chpl.attestation.domain.Attestation;
 import gov.healthit.chpl.attestation.domain.AttestationPeriod;
+import gov.healthit.chpl.attestation.domain.AttestationPeriodDeveloperException;
 import gov.healthit.chpl.attestation.domain.AttestationSubmittedResponse;
 import gov.healthit.chpl.attestation.domain.DeveloperAttestationSubmission;
 import gov.healthit.chpl.attestation.entity.AttestationEntity;
+import gov.healthit.chpl.attestation.entity.AttestationPeriodDeveloperExceptionEntity;
```
kekey1, line 15: When I saw this I was thinking it was a new type of Exception class and then realized it was not. I wish there was something else we could call this (at least in the code) besides an "exception" since that has a lot of meaning in Java. Anomaly, Variation, Deviation? Just wanted to mention it. At this point it would be a lot of written code to change if there even was another appropriate term.

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1277, opened 2022-03-21
OCD-3913: Do not include b6 in criteria that need to be updated to meet Cures certification

kekey1:
Do you guys mind if I leave the unit test in there? There is a lot mocked (and would be a lot to re-mock if we want to add any additional unit tests for the Cures Update Service in the future to test 24 m, 36 m functionality).


tmy1313 (approved):
I would have looked a long time before figuring that one out!  Maybe in the future we should return a copy of the criteria collections from this service to prevent something like this in the future?

andlar (approved):

PR #1301, opened 2022-05-11
feat: Remove one-time job to populate CHPL Product Number History

kekey1:
[#OCD-3902]


chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -858,14 +858,5 @@
                 </entry>
             </job-data-map>
         </job>
-
-        <job>
-            <name>populateChplProductNumberHistoryJob</name>
```
andlar, line 872: Do we need to add the "deleteJob" command to the "run at start up" steps?

kekey1: Yes!

andlar (dismissed):
I'm approving this as is, but it has merge conflicts, so I expect the approval to get erased at some point

andlar (approved):

tmy1313 (approved):

PR #1314, opened 2022-06-01
OCD-3960: Create a job to add missing measures to listings

kekey1:


andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/onetime/measures/AddMissingMeasuresToListingsJob.java
```diff
@@ -0,0 +1,199 @@
...
+                }
+            }
+        });
+        LOGGER.info("********* Completed the Add Missing Measures to Listings job. *********");
+    }
+
+    private void addMissingMeasures(CertifiedProductSearchDetails listing) {
+        LOGGER.info("Adding missing measures to listing " + listing.getId());
+        addMissingMeasure(listing, Criteria2015.B_1_OLD, "RT7 EH/CAH Medicare and Medicaid PI", "RT7 EH/CAH Medicare PI");
+        replaceListingMeasuresWithUpdatedMeasures(listing);
```
andlar, line 98: This is a very confusing method. Does this line need to be repeated? Why? Also, why isn't it repeated at the end of the method after doing the second g10 measure?

kekey1: Well each time I am adding replacement measures or possibly adding associated criteria to an existing replacement measure I wanted to set the new listing measures on the details object so that for the next criteria I'm working with the latest listing details. I don't know if any of them could overlap in this way but if I just add measure, add measure, add measure, add criteria to measure, add measure, etc and between all of those adding of things I don't go get the latest "current state" then it feels like I could/would be making changes to data that is stale. So I wanted the logic to go 1) add measure to listing, 2) update the details object i'm working from to have the latest measures (including the one i just added), 3) add next measure to listing 4) update the details object to have the latest measures 5) add next measure to listing, and so on and so forth. Open to method naming suggestions. We don't need this after the last g10 is done because there are no more updates to make.

andlar: Okay. I can understand that reasoning. I might want to think on this more if it weren't a one-time job that we're removing after it runs. That said, it isn't, so I'm good with it

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1326, opened 2022-06-29
[#OCD-3261] feat!: Update removeCriteriaJob for ERD Phase 2 requirements

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/onetime/criteriaremoval/RemoveCriteriaJob.java
```diff
@@ -30,18 +48,23 @@ public RemoveCriteriaJob() throws Exception {
...
+                    String[] criterionElements = criteria.split(":");
+                    CertificationCriterionDTO certDto = certCriteriaDao.getByNumberAndTitle(criterionElements[0], criterionElements[1]);
+                    certDto.setRemoved(true);
+                    certCriteriaDao.update(certDto);
+                    LOGGER.info("Updated criteria " + criteria);
+                } catch (final Exception ex) {
+                    LOGGER.error("Exception updating criteria " + criteria, ex);
+                }
             }
+            CacheManager.getInstance().clearAll();
```
kekey1, line 76: Can you also clear the shared store here? I suspect this job will be run after the shared store is deployed.

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/onetime/criteriaremoval/RemoveCriteriaJob.java
andlar: Can I? Or does that depend on OCD-3949 going to STG? Also not sure if that branch even _has_ a "clear all store" method

kekey1: Don't know! Question for @tmy1313 . I believe it needs it at some point anyway so I want to leave this comment here but approve the feature going forward if the clearing code is not available yet. If it depends on OCD-3949 then you'll need to make a commit with that clearing code when this goes to QA I guess.

tmy1313: My thought would be to add a checkbox to OCD-3949 that indicates that we need to clear the Shared Listing Store when this job is run and then implement the change when this ticket goes to STG.  I would leave the code for this ticket alone for now.  But this does leave implications for testing this change - how do we test this change in any environment where OCD-3949 has already been implemented?  We probably need to discuss...

kekey1 (approved):

tmy1313 (approved):

PR #1334, opened 2022-07-12
OCD-3951 - remove test which used removed response field

athomas-git:
OCD-3951 - remove test which use removed response field


chpl/chpl-api/e2e/collections/surveillance-controller.postman_collection.json
```diff
@@ -232,10 +232,6 @@
 									"pm.test(\"ROLE_ONC user should have access to create surveillance\", function () {\r",
 									"    pm.response.to.have.status(200);\r",
 									"});\r",
-									"pm.test(\"ROLE_ONC user initiated surveillance is showing authority as ROLE_ACB\", function () {\r",
```
kekey1, line 235: If I do a "find" on this file I still see the field "authority" in 5 or 6 places throughout. It will no longer be present in the API responses or requests so I think it should be totally removed from this file as well.

kekey1 (changes requested):

kekey1 (approved):

chpl/chpl-api/e2e/collections/surveillance-controller.postman_collection.json
athomas-git: Removed "authority" from all request bodies

ashwinimore (approved):

PR #1341, opened 2022-08-09
OCD-3999 - Replaced SVAP values should still be allowed for new listings with ICS

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/normalizer/SvapNormalizer.java
```diff
@@ -23,30 +22,31 @@
...
-            LOGGER.error("Could not initialize svap criteria map for flexible upload.", ex);
-        }
+        this.svapDao = svapDao;
     }
 
     public void normalize(CertifiedProductSearchDetails listing) {
-        if (!CollectionUtils.isEmpty(listing.getCertificationResults())) {
-            listing.getCertificationResults().stream()
-                .forEach(certResult -> fillInSvapData(certResult));
+        try {
```
kekey1, line 33: Instead of putting try/catch around this whole block of code, could you just try/catch the svapDao.getAllSvapCriteriaMap() method? And initialize the svapCriteriaMap object to an empty ArrayList so that if the DAO call fails it's just an empty list. I think that would match the behavior that would have occurred before. Unless you did it this way on purpose for a reason I didn't think of - then I'm happy to learn something!

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/validation/reviewer/SvapReviewer.java
```diff
@@ -100,9 +100,15 @@ private void reviewRegulatoryTextCitationRequired(CertifiedProductSearchDetails
 
     private void reviewSvapMarkedAsReplaced(CertifiedProductSearchDetails listing,
             CertificationResult certResult, CertificationResultSvap svap) {
-        if (svap.getSvapId() != null && BooleanUtils.isTrue(svap.getReplaced())) {
-            listing.getWarningMessages().add(msgUtil.getMessage("listing.criteria.svap.replaced",
+        if (svap.getSvapId() != null
```
kekey1, line 103: There is a suite of svap AQA tests for flexible upload. We hadn't covered this particular message, but I tested the file on DEV and noticed that this error now appears for one of the uploaded listings being tested - so this PR will need an AQA update. I am comfortable doing it since I wrote most of those if you'd like. Or happy to show you.

chpl/chpl-service/src/main/java/gov/healthit/chpl/sharedstore/SharedStoreDAO.java
```diff
@@ -56,6 +56,15 @@ public void remove(String type, List<String> keys) {
         entityManager.createQuery(delete).executeUpdate();
     }
 
+    @Transactional()
+    public void removeByDomain(String domain) {
```
kekey1, line 60: Is this called from anywhere? In the diff here, I see it as a new method but I cannot find it referenced again in any of the other diffs. (Certainly could have missed it)

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/upload/listing/normalizer/SvapNormalizer.java
tmy1313: image
`​Local variable svapCriteriaMap defined in an enclosing scope must be final or effectively final​`

I originally wrote the method as you suggested, but this compile error was shown.  I changed it to the current structure to remove the compile error.  I figured since the ​svapDao.getAllSvapCriteriaMap​ is the only method that can throw a ​EntityRetrievalException​ exception, that this structure is effectively the same.

chpl/chpl-service/src/main/java/gov/healthit/chpl/sharedstore/SharedStoreDAO.java
tmy1313: This is called in `SharedStoreProvider.removeAll()`
image

kekey1 (dismissed):

andlar (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #1351, opened 2022-08-25
OCD-4006: Send email to developer user(s) when attestation submission is

kekey1:
re-opened

[#OCD-4006]


chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/manager/AttestationExceptionEmail.java
```diff
@@ -0,0 +1,84 @@
...
+        private String getActingBody() {
+            if (resourcePermissions.isUserRoleAcbAdmin()) {
+                List<CertificationBodyDTO> acbsForUser = resourcePermissions.getAllAcbsForCurrentUser();
+                if (CollectionUtils.isEmpty(acbsForUser)) {
+                    LOGGER.warn("No ACBs were found for the current user " + AuthUtil.getCurrentUser().getSubjectName());
+                    return "an ONC-ACB";
+                }
+                return acbsForUser.get(0).getName();
+            } else if (resourcePermissions.isUserRoleAdmin()) {
+                return "CHPL Admin";
```
andlar: In the past, when we had the "Surveillance initiated by" field for Surveillance, if the ACB did it the value was "ONC-ACB", and if either ROLE_ADMIN or ROLE_ONC did it the value was just "ONC". I think we should keep that same thing here.

The basic idea is that if someone with ROLE_ADMIN does it, we're acting on behalf of ONC, and so the messaging should indicate that. If you disagree we can bring this up with the PO.

chpl/chpl-service/src/main/java/gov/healthit/chpl/attestation/manager/AttestationExceptionEmail.java
```diff
@@ -0,0 +1,84 @@
...
+            } else if (resourcePermissions.isUserRoleOnc()) {
+                return "ONC";
+            } else {
+                return "";
```
andlar: We probably need some other default here, or should throw an exception. There's a default for the ACB; can we use something similar? Maybe "an administrator" and LOG a warning?

andlar (approved):

tmy1313 (approved):

PR #1361, opened 2022-09-28
OCD-3263: Adjust listing validation for ERD-Phase-2 and Grace Period

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/validation/listing/Edition2015ListingValidator.java
```diff
@@ -224,64 +241,73 @@ public class Edition2015ListingValidator extends Validator {
...
+        comparisonReviewers.add(devBanComparisonReviewer);
+        comparisonReviewers.add(measureComparisonReviewer);
+        comparisonReviewers.add(criteriaComparisonReviewer);
+        comparisonReviewers.add(testTaskCriteriaComparisonReviewer);
+        comparisonReviewers.add(ucdCriteriaComparisonReviewer);
+        comparisonReviewers.add(testFunctionalityAllowedByRoleReviewer);
+        comparisonReviewers.add(listingStatusAndUserRoleReviewer);
+        if (!ff4j.check(FeatureList.ERD_PHASE_2)
+                || (ff4j.check(FeatureList.ERD_PHASE_2) && ff4j.check(FeatureList.ERD_PHASE_2_GRACE_PERIOD))) {
+            comparisonReviewers.add(privacyAndSecurityCriteriaReviewerPreErdPhase2);
```
andlar: This looks like we're adding the "p&sCriteria Pre-ERD-Phase-2" in the time before ERD-Phase-2 and after ERD-Phase-2-Grace-Period, but not during the grace period. Is that right? So it would add it today, and in June of 2023, but not in Feb of 2023?

kekey1: Hmmm. I mean for it to apply 1) now, before Dec 31 - so ERD-Phase-2 is off, and 2) During the grace period - so ERD-Phase-2 is ON and the grace-period flag is ON.  That is how I thought the flags would work. On Dec-31 we will turn on BOTH flags. I don't think it would be a valid state to have the grace period flag on with the ERD-Phase-2 flag off. So this validator should run now and feb 2023 but not june 2023. In June 2023, the other p&s reviewer should run but it should never run before that (higher up in the diff). Can you double-check my logic?

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1380, opened 2022-10-27
OCD-4056 - Remove Deprecated endpoints eligible for removal close to now

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/search/ListingSearchService.java
```diff
@@ -630,23 +629,6 @@ public int compare(ListingSearchResult listing1, ListingSearchResult listing2) {
...
-        searchResponse.setPageNumber(listingSearchResponse.getPageNumber());
-        searchResponse.setPageSize(listingSearchResponse.getPageSize());
-        searchResponse.setRecordCount(listingSearchResponse.getRecordCount());
-        searchResponse.setResults(listingSearchResponse.getResults().stream()
-                .map(listingSearchResult -> convertToBasicSearchResult(listingSearchResult))
-                .collect(Collectors.toList()));
-        return searchResponse;
-    }
-
     private CertifiedProductBasicSearchResult convertToBasicSearchResult(ListingSearchResult searchResult) {
```
kekey1: Can this method also be deleted?

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/ActivityController.java
```diff
@@ -136,27 +127,6 @@ public ActivityMetadataPage metadataForListings(@RequestParam(required = false)
                 ActivityConcept.CERTIFIED_PRODUCT, start, end, pageNum, pageSize);
     }
 
-    @Deprecated
```
kekey1: I can't leave a comment on the line that I want, but I think a number of fields in this class are no longer referenced (atlManager, userManager, resourcePermissions) and are candidates for deleting from this file.

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/SearchController.java
```diff
@@ -166,144 +164,6 @@ public SearchController(ListingSearchService searchService) {
         return searchService.findListings(searchRequest);
     }
 
-    @Deprecated
-    @DeprecatedApi(friendlyUrl = "/search/beta",
```
kekey1: This is annoying, but since we don't have /search/beta anymore, is the plan to rename /search/v2 to just be /search (meaning deprecate the existing endpoint and add a new one called as just "/search"? Or should it be /listings/search or  /certified-products/search? Maybe a bigger question than this PR, I just thought if we want to do that we could deprecate the "v2" one and add the new one at this time).

kekey1 (changes requested):
Two changes requested and one comment either requiring a change or just discussion.

kekey1 (dismissed):

andlar (dismissed):

andlar (approved):

PR #1384, opened 2022-11-03
OCD-4073: Dependabot commons-text

kekey1:
I'm actually not sure we are using this library at all... but it's hard to tell. I removed it and there were no compilation errors but sometimes there are run-time dependencies and I can't test everything.


andlar (dismissed):
If we're not sure we're using it, maybe this is an opportunity to just remove it and see how it goes on DEV/QA for a while? We could add burn-in to the ticket and just let it sit for a bit. Any dead code we can remove is a good thing, right?

andlar (approved):

tmy1313 (approved):

PR #1385, opened 2022-11-07
OCD-3793 - b7 and b7 Cures should have Test Data

tmy1313:


kekey1 (approved):

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/CertificationResultRules.java
```diff
@@ -165,6 +165,13 @@ private void setRulesUsingDatabase(CertificationCriterionAttributeDAO certificat
                         .optionName(TEST_TOOLS_USED)
                         .build());
             }
+            if (attribute.getTestData()) {
```
andlar, line 168: I know it's a tiny thing, but the other rules in this method are sorted alphabetically. Can we move this one two rules up?

chpl/chpl-service/src/main/java/gov/healthit/chpl/util/CertificationResultRules.java
```diff
@@ -165,6 +165,13 @@ private void setRulesUsingDatabase(CertificationCriterionAttributeDAO certificat
...
+                rules.get(attribute.getCriterion().getNumber()).add(
+                        CertificationResultOption.builder()
+                        .canHaveOption(true)
```
andlar, line 171: Do we ever call this builder with `.canHaveOption(false)`, or even `.canHaveOption(aVariable)`? If not, would it make sense / reduce code to just default this to `true`? Is that even possible? Or worth it?

andlar (approved):

PR #1410, opened 2022-12-07
OCD-4081: Updates to Questionable URL Report

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/urlStatus/QuestionableUrlReportGenerator.java
```diff
@@ -170,8 +170,11 @@ private List<FailedUrlResult> filterUrls(List<FailedUrlResult> badUrls, JobExecu
                 .filter(badUrl -> isNotListingUrl(badUrl) || (isUrlRelatedTo2015Edition(badUrl) && isUrlRelatedToActiveListing(badUrl)))
                 .filter(badUrl -> doesUrlResultMatchAllowedStatusCodes(badUrl, jobContext))
                 .collect(Collectors.toList());
+        } else {
+            return badUrls.stream()
+                .filter(badUrl -> isNotListingUrl(badUrl) || (isUrlRelatedTo2015Edition(badUrl) && isUrlRelatedToActiveListing(badUrl)))
```
andlar, line 175: Given the change in the DAO related to the query against Active Certificates, does the part after the `||` always evaluate to `true` now, for any URL that is a ListingUrl? So the `isNotListing` call will be true for non-Listing URLs, and the second clause will be true for all Listing URLs, so this whole filter will always be `true || false` or `false || true`, which means the filter never filters anything?

kekey1: I thought so, too, and only added this in after I found it was not the case. When we collect status data for URLs related to listings, we are using only URLs for listings that are Active/Suspended. Later when we get into creating this report we take each failed URL and figure out what listing(s) it relates to so we can include each one in the report. Sometimes the failed URL that was used for an Active/Suspended listing is ALSO used in an Inactive listing so this filter is meant to get rid of those.

andlar (dismissed):

tmy1313 (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1414, opened 2022-12-12
OCD- 4061 - AQA tests to verify response and status when passing invalid date parameters

athomas-git:
OCD- 4061 - AQA tests to verify response and status when passing invalid date parameters


kekey1 (approved):
Were there meaningful changes in the developer controller test file? I couldn't tell.

athomas-git:
The only new test added to the developer controller is that /developer/search end point is returning 200 status and no response when passing invalid date parameters

ashwinimore (approved):

PR #1420, opened 2022-12-19
OCD-4095 - Listing Details pages do not show Direct Review updates

tmy1313:


chpl/chpl-api/src/main/resources/log4j2-xinclude-loggers-local.xml
```diff
@@ -99,9 +99,6 @@
     <Logger name="developerAccessEmailJobLogger" level="INFO" additivity="false">
         <AppenderRef ref="developerAccessEmailJob" />
     </Logger>
-    <Logger name="directReviewCacheRefreshJobLogger" level="INFO" additivity="false">
```
kekey1: Why don't we use this logger locally any longer?

chpl/chpl-service/src/main/java/gov/healthit/chpl/directreview/DirectReviewUpdateEmailService.java
```diff
@@ -1,4 +1,4 @@
-package gov.healthit.chpl.service;
+package gov.healthit.chpl.directreview;
```
kekey1: I like moving these DR-specific services to their own package. I would like to settle on the package naming though. All of the other DR code is under domain/compliance. I don't care if we call the package "compliance" or "directreview" but just want it to be consistent. I think I had a notion of a compliance package with surveillance and directreview subpackages but never got there. What do you think? Should we also move the several files in domain/compliance (like the DR domain objects and deserializer) to be together with these services?

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/compliance/ListingDeserializer.java
```diff
@@ -17,23 +17,57 @@
 
 @Component
 public class ListingDeserializer extends JsonDeserializer<List<DeveloperAssociatedListing>> {
+    private static final String ID_FIELD = "id";
+    private static final String CHPL_PRODUCT_NBR_FIELD = "chplProductNumber";
```
kekey1: Can you just write out the word NUMBER in the variable name?

chpl/chpl-service/src/main/java/gov/healthit/chpl/directreview/DirectReviewListingSharedStoreHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private DirectReview findDirectReview(List<DirectReview> directReviews, String jiraKey) {
+        return directReviews.stream()
+                .filter(dr -> dr.getJiraKey().equals(jiraKey))
+                .findAny()
+                .orElse(null);
+    }
+
+    private List<CertifiedProductSearchDetails> getUniqueListings(List<DirectReview> allDirectReviews) {
```
kekey1: I think this method is trying to get all the listings related to all direct reviews to potentially remove those listings from the shared store. There is a little bit more complicated logic to say "is a direct review associated to this listing" than looking at the Developer-Associated Listings. The logic for that is encapsulated in the DirectReviewSearchService getDirectReviewsRelatedToListing method (I think the method's comment block is accurate).  I wonder if you could make use of that method in some way? If a Direct Review doesn't have any Developer-Associated Listings it can still be associated to listings related to the developer and I think we would want to clear those listings from the shared store.

kekey1 (changes requested):

chpl/chpl-api/src/main/resources/log4j2-xinclude-loggers-local.xml
tmy1313: I removed it locally just for testing purposes.  Forgot to revert the change.

chpl/chpl-service/src/main/java/gov/healthit/chpl/compliance/directreview/DirectReviewListingSharedStoreHandler.java
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            //TODO: this needs to switched to use developer id when OCD-4066 gets to STG!!!
+            SearchRequest searchRequest = SearchRequest.builder()
+                    .certificationEditions(Stream.of(CertificationEditionConcept.CERTIFICATION_EDITION_2015.getYear()).collect(Collectors.toSet()))
+                    .developer(developer.getName())
+                    .pageSize(SearchRequest.MAX_PAGE_SIZE)
+                    .build();
+
+            return listingSearchService.getAllPagesOfSearchResults(searchRequest).parallelStream()
+                    .map(searchResult ->  getListing(searchResult.getId()))
```
kekey1: I'm wondering why you need to get the listing details here. I could be missing it, but I suspect the only thing you really need is the ID which you'll use to remove the listing from the shared store. Just trying to avoid extra work if it's not needed. I see the next line filters if the listing is not null. Was there a case where it was null?

chpl/chpl-service/src/main/java/gov/healthit/chpl/compliance/directreview/DirectReviewListingSharedStoreHandler.java
```diff
@@ -0,0 +1,120 @@
...
+            @Lazy CertifiedProductDetailsManager certifiedProductDetailsManager, @Lazy ListingSearchService listingSearchService,
+            DeveloperDAO developerDAO) {
+        this.sharedListingStoreProvider = sharedListingStoreProvider;
+        this.certifiedProductDetailsManager = certifiedProductDetailsManager;
+        this.listingSearchService = listingSearchService;
+        this.developerDAO = developerDAO;
+    }
+
+    public void handler(List<DirectReview> allDirectReviews) {
+        getUniqueDevelopers(allDirectReviews).stream()
```
kekey1: This looks good - will do what we talked about. I was wondering if you changed the order of checks here if it would it be more efficient (enough to make it worthwhile)? Like, 99% of the time we expect the direct reviews won't have changed, so would it make sense to look for that first? Then the method becomes a no-op most of the time. And only if you find that any changed do you go through and get all the developers + all their listings. It's probably not that much extra work being done but since I have time I was thinking about that. I'm fine if you want to leave it how it is, too.

kekey1 (dismissed):
I was expecting a change in the DirectReviewCachingService as well to call DirectReviewListingSharedStoreHandler.handle in the "getDirectReviews" method(s) that are annotated with CachePut so that if a user gets updated direct reviews when navigating to the developer's page it will force the listing data to refresh as well. Sorry if you're just not finished yet!

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/compliance/directreview/DirectReviewListingSharedStoreHandler.java
```diff
@@ -0,0 +1,126 @@
...
+    private DirectReview findDirectReview(List<DirectReview> directReviews, String jiraKey) {
+        return directReviews.stream()
+                .filter(dr -> dr.getJiraKey().equals(jiraKey))
+                .findAny()
+                .orElse(null);
+    }
+
+    private List<CertifiedProductSearchDetails> getListingDataForDeveloper(Developer developer, Logger logger) {
+        try {
+            //TODO: this needs to switched to use developer id when OCD-4066 gets to STG!!!
```
andlar: Adding a comment here so we don't forget this `TODO`

andlar (approved):

andlar (changes requested):
Don't want to forget the TODO based on OCD-4066

kekey1 (dismissed):

andlar (dismissed):

andlar (approved):

kekey1 (approved):

PR #1424, opened 2022-12-30
OCD-4086: Make Cures Update true-to-false be questionable activity

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/questionableactivity/entity/package-info.java
```diff
@@ -2,4 +2,4 @@
  * Entity classes mapping java beans to hibernate-managed database connections
  * for questionable activities
  */
-package gov.healthit.chpl.entity.questionableActivity;
+package gov.healthit.chpl.questionableactivity.entity;
```
andlar: Don't we usually just delete the package-info files when we come across them?

chpl/chpl-service/src/main/java/gov/healthit/chpl/listener/QuestionableActivityListener.java
```diff
@@ -45,40 +45,39 @@ public void setEnvironment(final Environment e) {
...
 
     private void checkQuestionableActivityForListingEdit(ActivityConcept concept, Long objectId, String activityDescription,
-            CertifiedProductSearchDetails originalData, CertifiedProductSearchDetails newData, String reason) {
+            Date activityDate, CertifiedProductSearchDetails originalData, CertifiedProductSearchDetails newData, String reason) {
 
         if (originalData == null || newData == null || AuthUtil.getCurrentUser() == null) {
             return;
         }
 
         // all questionable activity from this action should have the exact same date and user id
```
andlar: There are a few comments in this file that don't quite make sense if the date is now a passed in value. I know comments aren't super important, but _wrong_ comments seem like something we shouldn't allow

kekey1: I feel like the comments are still true... all questionable activity from this action SHOULD have the exact same date. It's just being passed into the method. However, point take, and I updated the comments.

andlar (approved):

tmy1313 (approved):

PR #1432, opened 2023-01-14
OCD-3766: Test functionality -> functionality tested

kekey1:
- Deprecate GET /data/test_functionalities
- Add GET functionalities-tested
- Deprecate allowedTestFunctionalities field of listing details
- Deprecate testFunctionality field of listing details
- Add functionalitiesTested field to listing details, and update XML accordingly
- Remove the concept of a Functionality Tested being associated to a year/edition
- Change the criterion attribute lookup to be by criterion ID (although we had the capability in the data, we were still using the number in practice)


chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/CacheNames.java
```diff
@@ -25,7 +25,7 @@ private CacheNames() {}
     public static final String FIND_SURVEILLANCE_RESULT_TYPE = "findSurveillanceResultType";
     public static final String TEST_PROCEDURES = "testProcedures";
     public static final String TEST_DATA = "testData";
-    public static final String TEST_FUNCTIONALITY_MAPS = "testFunctionalityMaps";
+    public static final String FUNCTIONALITY_TESTED_MAPS = "testFunctionalityMaps";
```
andlar: Should the name be changed too?

chpl/chpl-service/pom.xml
```diff
@@ -411,7 +411,7 @@
                         <source>src/main/java/gov/healthit/chpl/domain/CertificationResult.java</source>
                         <source>src/main/java/gov/healthit/chpl/domain/CertificationResultAdditionalSoftware.java</source>
                         <source>src/main/java/gov/healthit/chpl/domain/CertificationResultTestData.java</source>
-                        <source>src/main/java/gov/healthit/chpl/domain/CertificationResultTestFunctionality.java</source>
+                        <source>src/main/java/gov/healthit/chpl/functionalityTested/CertificationResultFunctionalityTested.java</source>
```
andlar: Is there something about these sources that they're supposed to be alphabetized?

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/caching/CacheNames.java
kekey1: Yes but I didn't want to because I have to change it in every environment and it hardly seems worth it. Updated it though.

chpl/chpl-service/pom.xml
kekey1: I don't know, but moved it.

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1433, opened 2023-01-18
OCD-4094 - Standardize Surveillance Requirement and Nonconformity values

tmy1313:


kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/NonconformityType.java
```diff
@@ -87,11 +87,14 @@ public void setClassification(NonconformityClassification classification) {
...
 
+    public CertificationEdition getCertificationEdition() {
+        return certificationEdition;
+    }
+
     public void setEdition(String edition) {
         this.edition = edition;
     }
 
-    @JsonIgnore
```
andlar: Does removing this annotation put the `edition` field into the API's response object somewhere? If so, do we need to do that if we can just use the full `certificationEdition` that was added a few lines above?

Will this affect the XSD at all?

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/domain/NonconformityType.java
```diff
@@ -51,6 +51,10 @@ public void setId(Long id) {
         this.id = id;
     }
 
+    public CertificationEdition getCertificationEdition() {
```
andlar, line 54: Does adding this element change the XSD? If it does, we'll need the XSD-Impact label and a change to the XSD-MODIFICATION file

tmy1313: It does not look to be in the XSD.
image

kekey1 (approved):

andlar (approved):

PR #1437, opened 2023-01-24
[#OCD-4078] fix: update report to use most current ID for assurances data

andlar:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/developer/attestation/DeveloperAttestationCheckInReportDataCollector.java
```diff
@@ -28,7 +28,7 @@
 public class DeveloperAttestationCheckInReportDataCollector {
 
     private static final Long INFO_BLOCKING_CONDITION = 1L;
-    private static final Long ASSURANCES_CONDITION = 2L;
+    private static final Long ASSURANCES_CONDITION = 7L;
```
kekey1, line 31: Is this just a one-time change or is it every time we add a new form for a new period? One-time change seems okay but if it will frequently change I think we need a different strategy...

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/developer/attestation/DeveloperAttestationCheckInReportDataCollector.java
andlar: It'll need to change each time we change the form. So if the questions don't change we don't need to change this file. I don't like this as a good long-term solution, but for the purposes of scope creep and timing I am okay with this as a single change right now. Not against adding a ticket to make this better next time

kekey1: I agree with adding a ticket. This is the type of thing we will forget every time. And we should not have IDs that are hardcoded if they are going to change. What happens now when someone tries to edit/open an old version of the form for a previous attestation period?

andlar: The report that's generated only has data from the "current" period, so previous period data won't show in the report anyway. If someone _does_ open a previous period, that will all work correctly, as the rest of the system doesn't have this hard-coded stuff. The hard-coded thing is _only_ for the purposes of this specific report. I'll put in a ticket to make the report better

kekey1 (approved):

tmy1313 (approved):

PR #1442, opened 2023-02-03
OCD-4050: Don't allow all products to be transferred from developer

kekey1:


andlar (dismissed):

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

.gitignore
```diff
@@ -16,6 +16,8 @@ local.properties
 .settings/
 .loadpath
 target/
+chpl/chpl-api/src/main/webapp/META-INF/
+chpl/chpl-api/src/main/webapp/.gitignore
```
andlar: Usually the `.gitignore` file is saved in git, to make sure all the things git should ignore get ignored. Is there something different about this file?

kekey1: Ugh. My new version of Eclipse is auto-generating this META-INF directory AND auto-generating a .gitignore inside of it. I think it's a setting that I can't figure out... I have tried. Todd doesn't have this problem but all of our settings look the same. I can get Eclipse to STOP auto-generating the folder/file but then I'm not able to use the built-in Tomcat and deploy the project as a web application. I have been trying so hard not to commit these auto-generated files but I'm sure I'll mess up at some point... so I decided that it was relatively harmless to add the Eclipse auto-generated stuff to the main .gitignore and that's what I did.

andlar: Ignoring the autogenerated stuff is a great idea, but this change also ignores the file that determines what stuff is ignored, I think. I can't think of a reason to ignore the file that defines what should be ignored. I'm fine with line 19, but is there stuff in the file in line 20 that now won't be ignored because it we're ignoring the file that tells git to ignore it?

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1478, opened 2023-04-18
OCD-3896: Send user an email when their password is changed

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/AuthenticationController.java
```diff
@@ -86,22 +83,37 @@ public AuthenticationController(AuthenticationManager authenticationManager, BCr
...
+        String jwt = authenticationManager.refreshJWT();
+        return AuthenticationResponse.builder()
+                .token(jwt)
+                .build();
     }
 
     @Hidden
+    @Deprecated
+    @DeprecatedApi(friendlyUrl = "/auth/keep_alive",
+            message = "This endpoint is deprecated and will be removed. Please use /auth/keep_alive.",
```
andlar:
```suggestion
            message = "This endpoint is deprecated and will be removed. Please use /auth/keep-alive.",
```

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/AuthenticationController.java
```diff
@@ -153,12 +165,66 @@ public UpdatePasswordResponse changePassword(@RequestBody UpdatePasswordRequest
...
+            })
+    @Deprecated
+    @DeprecatedApi(friendlyUrl = "/auth/change_password",
+            message = "This endpoint is deprecated and will be removed. Please use /auth/change-password.",
+            removalDate = "2023-10-31")
+    @RequestMapping(value = "/change_password", method = RequestMethod.POST,
+            produces = "application/json; charset=utf-8")
+    public UpdatePasswordResponse changePasswordDeprecated(@RequestBody UpdatePasswordRequest request)
+            throws UserRetrievalException, MultipleUserAccountsException {
+        UpdatePasswordResponse response = new UpdatePasswordResponse();
```
andlar, line 183: If all this code is the same code as is in the new endpoint, is there a reason not to either extract it into a private method or just call the other public one so it's just in one place?

andlar (dismissed):

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/AuthenticationController.java
kekey1: I thought about doing that, but I felt that 1) it makes more work when you remove the deprecated endpoint because you no longer really need the private method so you would probably un-extract that code, and 2) if you want to make a change only to the non-deprecated version of the endpoint before the deprecated one is removed then you have to move things around again. That said, I certainly CAN extract it into a private method and proceed that way. What is your preference?

andlar: It is definitely a preference thing. It does seem unlikely we'd want to change the functionality behind the endpoint either to distinguish between the two, or to make a change that wouldn't cause a need to deprecate & create a new endpoint, so I see no need to make changes here

andlar (dismissed):

tmy1313 (approved):

andlar (approved):

PR #1487, opened 2023-05-03
OCD-4147: Show attestation warnings for ONC-ACB users

kekey1:


andlar (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/service/ListingApplicabilityService.java
tmy1313, line 1: I like this class, but think maybe it should be renamed or moved to a different package.  Since this is not where we globally determine if a listing is RWT applicable, someone might think this how to determine if a listing is eligible for RWT.  I would suggest adding the word "attestation" to the class name or moving it to an attestation specific package

kekey1: Moved! Thank you for the feedback and you are absolutely right.

andlar (dismissed):

tmy1313 (dismissed):

andlar (dismissed):

tmy1313 (approved):

andlar (approved):

PR #1535, opened 2023-08-28
OCD-4293: Use listing status in reports rather than edition

kekey1:


andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/BrokenSurveillanceRulesCreatorJob.java
```diff
@@ -117,14 +117,13 @@ public void execute(JobExecutionContext jobContext) throws JobExecutionException
 
     private List<ListingSearchResult> getListingsForReport() {
         return listingSearchDao.getListingSearchResults().stream()
-                .filter(listing -> isEdition2015(listing)
-                        && (isCertificationStatusSuspendedByAcb(listing)
-                                || hasSurveillances(listing)))
+                .filter(listing -> isCertificationStatusSuspendedByAcb(listing)
+                                || (isActive(listing) && hasSurveillances(listing)))
```
tmy1313, line 121: Is the `isActive(...)` necessary here?

tmy1313 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/BrokenSurveillanceRulesCreatorJob.java
kekey1: Well, I think I put that there because the old logic would have returned true if there was a "2015 edition listing and it had a surveillance" so I changed it to "is active and has surveillance". I don't think I want to include listings in this report if they have surveillance but are not active listings?

andlar (approved):

tmy1313 (approved):

PR #1544, opened 2023-09-19
OCD-4351 - Clear Shared Store when Criteria Attributes are modified

tmy1313:
OCD-4351


kekey1 (commented):
I agree with the changes made here, but I think there is at least one other needed. In the TestToolNormalizer, I see the testToolCriteriaMaps are loaded in it's constructor. This was fine when they couldn't be changed but now they can be and I think this might cause issues down the line if we leave the code like this. 


tt constructor

chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -44,7 +44,7 @@
             <name>downloadFileJob2015</name>
             <group>systemJobs</group>
             <description>Generate download files for 2015.</description>
-            <job-class>gov.healthit.chpl.scheduler.job.CertifiedProductDownloadableResourceCreatorJob</job-class>
+            <job-class>gov.healthit.chpl.scheduler.job.downloadfile.CertifiedProductDownloadableResourceCreatorJob</job-class>
```
kekey1, line 59: Does this work on DEV without clearing the existing Quartz jobs? I feel like if you change the package names here you have to have a pre-processing command to delete the currently existing one. I didn't think they would get updated.

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/downloadfile/GenerateListingDownloadFilesAspect.java
```diff
@@ -0,0 +1,155 @@
...
+            LOGGER.error("Could not retrieve list of Quartz jobs", e);
+            return List.of();
+        }
+    }
+
+    private boolean isJobAlreadyScheduled(DownloadJobRunInformation info) {
+        return getAllScheduledSystemJobs().stream()
+                .filter(systemJob -> systemJob.getName().equalsIgnoreCase(info.getJobName())
+                        && systemJob.getNextRunDate().getTime() == info.getRunDateTime().toInstant().toEpochMilli())
+                .findAny()
```
andlar, line 106: I think this will always return "true" to "isAlreadyScheduled", as we have the 2011 and 2014 jobs in a "once per quarter" recurring schedule. I know we talked about removing that recurring schedule, but I don't think we can do that until we annotate all the other places that update the data in such a way that we'd want to regen those files. Is there a way to look for "isAlreadyScheduled SOON"?

kekey1 (dismissed):

andlar (dismissed):

andlar (approved):

kekey1 (approved):

PR #1557, opened 2023-10-17
OCD-4275: Separate download files into 2011, 2014, Inactive, Active

kekey1:


chpl/chpl-resources/src/main/resources/jobs.xml
```diff
@@ -20,22 +20,6 @@
...
-        </delete-job>
-        <delete-job>
-            <name>directReviewDownloadFileGeneration</name>
-            <group>systemJobs</group>
-        </delete-job>
-        <delete-job>
-            <name>removeCriteriaJob</name>
-            <group>systemJobs</group>
-        </delete-job>
         <!-- END remove after PROD push -->
```
andlar, line 23: Do we need to add a "delete" command to remove the "downloadFileJob2015" old job?

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertifiedProductDAO.java
```diff
@@ -307,6 +308,31 @@ public List<CertifiedProductDetailsDTO> getListingsByStatus(Collection<Certifica
                 .collect(Collectors.toList());
     }
 
+    @Transactional(readOnly = true)
+    public List<CertifiedProductDetailsDTO> getListingsByStatusExcludingEdition(
```
andlar: It's weird for the method name to say "excluding edition", but the where clause to use "editions to exclude". Maybe the name could be `getListingsByStatusExcludingRetiredEditions`? I think we got confirmation that 2015 won't be Retired, so that would exclude 2011 and 2014 only

andlar (dismissed):

chpl/chpl-resources/src/main/resources/jobs.xml
kekey1: It's already there. It doesn't show as a diff because it was there from some previous ticket and never removed, so I left it.

chpl/chpl-service/src/main/java/gov/healthit/chpl/dao/CertifiedProductDAO.java
kekey1: Updated. Thank you for the suggestion.

tmy1313 (dismissed):

andlar (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1590, opened 2023-12-18
OCD-4317 - Update Charts to account for Editionless

tmy1313:


andlar (approved):

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/StatisticsController.java
```diff
@@ -51,6 +52,11 @@ public NonconformityTypeStatisticsResult getNonconformityCountByCriteria() {
         return response;
     }
 
+
+    @Deprecated
```
kekey1, line 56: I sort of thought we might stop calculating these statistics as well. Since the job wasn't changed when is the plan for stopping calculating and saving this data?

kekey1 (approved):

PR #1650, opened 2024-05-07
OCD-4563: Notify subscribers about new types of changes to listings

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/subscriptions/subjects/formatter/ObservationSubjectFormatterFactory.java
```diff
@@ -35,6 +44,12 @@ public ObservationSubjectFormatter getSubjectFormatter(SubscriptionObservation o
             return criteriaAddedFormatter;
         } else if (lookupUtil.getCertificationCriteriaRemovedSubjectId().equals(observationSubjectId)) {
             return criteriaRemovedFormatter;
+        } else if (lookupUtil.getRwtPlansUrlChangedSubjectId().equals(observationSubjectId)) {
```
andlar: Just from a code format point of view, should this be a switch / case thing? Or, could we just remove all the "else"s? If we return out of the function on any "true" if statement, then the next ifs won't even be reached

Certainly not a change that's necessary, but am curious if there's interest in some sort of small refactor. Especially as we hope to add more of these down the line, and I _think_ the "if else if else if else" pattern is usually frowned upon. Could be wrong

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/subscriptions/subjects/formatter/RwtPlansUrlChangedFormatter.java
```diff
@@ -0,0 +1,60 @@
...
+import gov.healthit.chpl.subscription.domain.SubscriptionObservation;
+import gov.healthit.chpl.util.DateUtil;
+import lombok.extern.log4j.Log4j2;
+
+@Component
+@Log4j2(topic = "subscriptionObservationsNotificationJobLogger")
+public class RwtPlansUrlChangedFormatter extends ObservationSubjectFormatter {
+    private static final String DESCRIPTION_REMOVED = "RWT Plans URL %s was removed.";
+    private static final String DESCRIPTION_ADDED_UNFORMATTED = "RWT Plans URL %s was added.";
+    private static final String DESCRIPTION_UPDATED_UNFORMATTED = "RWT Plans URL was changed from %s to %s";
```
andlar: The first two text lines end with a period; the last one doesn't. I don't know what the other observation strings look like, but we should be consistent across all of them, including the three pre-existing ones

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/subscriptions/subjects/formatter/ServiceBaseUrlListChangedFormatter.java
```diff
@@ -0,0 +1,83 @@
...
+import gov.healthit.chpl.util.DateUtil;
+import gov.healthit.chpl.util.Util;
+import lombok.extern.log4j.Log4j2;
+
+@Component
+@Log4j2(topic = "subscriptionObservationsNotificationJobLogger")
+public class ServiceBaseUrlListChangedFormatter extends ObservationSubjectFormatter {
+    private static final String DESCRIPTION_REMOVED_UNFORMATTED = "Service Base URL List %s was removed from %s.";
+    private static final String DESCRIPTION_ADDED_UNFORMATTED = "Service Base URL List %s was added to %s.";
+    private static final String DESCRIPTION_UPDATED_UNFORMATTED = "Service Base URL List was changed from %s to %s on %s";
```
andlar: What are the `_UNFORMATTED` tags for on these (and the other ones)? They seem redundant

kekey1: Just to indicate the message need to be properly formatted (with String.format, farther down in the class) to replace the %s values in the message with real text. I can change them

andlar: That makes sense, but the RWT Removed one also has a `%s` in it, but isn't marked as `UNFORMATTED`. I'm fine leaving it in if it means something, but it should be consistent across the strings

andlar (dismissed):

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/subscriptions/subjects/formatter/ObservationSubjectFormatterFactory.java
kekey1: I would be happy to do this, except the "id" that I want to switch on is a Long and Java only allows you to switch on String, enum, byte, short, char, and int. In my research this is because switch is expected to be used on a smallish list of things and if the list is big enough to be of type "long" then you probably wouldn't switch on it. We will never have that many subjects so it doesn't _have_ to be a "long" type, but all of our IDs are either Long or UUID so I don't know about making this one different. I could, I could also refactor and do something with an enum, or convert the Long to a String here and switch on it that way. I dunno. I agree with you that a switch would be better but I feel like it's not a simple change at this point.

andlar: Fair enough. What about the idea of just removing the `else` portions of all of the `else if` statements? If we return from inside the `if`, then we'll never get to the rest of the checks. I know I'm doing that a lot in the JS stuff where it's effectively short-circuiting the next checks, and I have heard people not like that kind of action, but it makes sense to me. More a passing thought than anything else

andlar (approved):

tmy1313 (approved):

PR #1663, opened 2024-06-05
OCD-4597 - Begin collecting data for (a)(9) to (b)(11) over time reporting

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriamigrationreport/CriteriaMigrationCount.java
```diff
@@ -0,0 +1,39 @@
...
+    @Schema(description = "Date which the counts were calculated")
+    private LocalDate reportDate;
+
+    @Schema(description = "Number of listings attesting the original criteria as of the reportDate")
+    private Integer originalCriterionCount;
+
+    @Schema(description = "Number of listings attesting the original criteria as of the reportDate")
+    private Integer updatedCriterionCount;
+
+    @Schema(description = "Number of listings where the original criteria was attested to, then unattested to, and trhe updated criteria has been attested to")
```
kekey1:
```suggestion
    @Schema(description = "Number of listings where the original criteria was attested to, then unattested to, and the updated criteria has been attested to")
```

chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriamigrationreport/CriteriaMigrationDefinitionEntity.java
```diff
@@ -0,0 +1,78 @@
...
+    @OneToMany(fetch = FetchType.LAZY, mappedBy = "id")
+    @Column(name = "criteria_migration_definition_id", nullable = false)
+    private List<CriteriaMigrationCountEntity> criteriaMigrationCounts;
+
+    public CriteriaMigrationDefinition toDomain() {
+        return CriteriaMigrationDefinition.builder()
+                .id(this.id)
+                .originalCriterion(this.originalCriterion.toDomain())
+                .updatedCriterion(this.updatedCriterion.toDomain())
+                .criteriaMigrationCounts(!Hibernate.isInitialized(this.criteriaMigrationCounts) ? null : this.criteriaMigrationCounts.stream()
```
kekey1, line 73: I've never seen this Hibernate.isInitialized method used before. Is there something unique about this entity or the HQL that gets it that makes this necessary? Is it like checking if there's a session before it would lazily try to fetch data? I think we should write our queries/code so that never happens right? Just wanted to understand this.

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriamigrationreport/CriteriaMigrationDefinitionEntity.java
```diff
@@ -0,0 +1,78 @@
...
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter
+@ToString
+@SuperBuilder
+@AllArgsConstructor
+@NoArgsConstructor
+@Entity
+@Table(name = "criteria_migration_definition")
```
andlar, line 37: If there are additional entities, don't we need a data-model PR as well?

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriamigrationreport/CriteriaMigrationDefinitionEntity.java
tmy1313: Yes - there are some cases where I just get the structure of the report: CriteriaMigrationReport -> CriteriaMigrationDefinitions.

Then, sometimes I need to the structure of the report and the associated data we have collected:  CriteriaMigrationReport -> CriteriaMigrationDefinitions -> CriteriaMigrationCounts 

So, there are instances where the "counts" were not fetched intentionally.  When I intentionally did not fetch the counts, I would get transaction errors since Hibernate would try to retrieve the counts on this particular line.  The `Hibernate.isInitalized` seemed to handle the problem.  

Might have gotten the answer from here:  https://stackoverflow.com/questions/1589144/is-there-a-helper-to-know-whether-a-property-has-been-loaded-by-hibernate

chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriamigrationreport/CriteriaMigrationDefinitionEntity.java
tmy1313: Added missing PR

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

kekey1 (approved):

andlar (approved):

PR #1682, opened 2024-07-15
OCD-4545 - Edit Cognito user information

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/user/cognito/CognitoUpdateUserValidator.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private Set<String> validateUser(User user) {
+        Set<String> validationErrors = new HashSet<String>();
+
+        if (!doesUserExistInCognito(user.getCognitoId())) {
+            validationErrors.add(msgUtil.getMessage("user.notFound"));
+        }
+
+        if (StringUtils.isEmpty(user.getFullName())) {
+            validationErrors.add(msgUtil.getMessage("cognito.user.create.fullName.empty"));
```
kekey1, line 50: I expected to see a change in errors.properties for this, but it's not showing that file changed

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CognitoUserController.java
```diff
@@ -182,4 +182,13 @@ public void createUser(@RequestBody CreateUserFromInvitationRequest userInfo) th
         }
     }
 
+    @Operation(summary = "Modify user information.", description = "",
+            security = {
+                    @SecurityRequirement(name = SwaggerSecurityRequirement.API_KEY),
+                    @SecurityRequirement(name = SwaggerSecurityRequirement.BEARER)
+            })
+    @RequestMapping(value = "/{ssoUserId}", method = RequestMethod.PUT, consumes = MediaType.APPLICATION_JSON_VALUE, produces = "application/json; charset=utf-8")
```
kekey1: I notice in some other code the variable is called cognitoUserId instead of ssoUserId. I was just wondering which we are going with. And also maybe we don't need this in the request mapping at all? It doesn't look like it's used.

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/user/cognito/CognitoUpdateUserValidator.java
tmy1313: This is an existing message and did not need to be added.

kekey1 (approved):

andlar (approved):

PR #1687, opened 2024-07-25
OCD-4646: set up CODEOWNERS

andlar:


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @kekey1
+* @tmy1313
```
kekey1, line 2: do you want to be in here?

kekey1 (approved):

.github/CODEOWNERS
andlar: I'm iffy on this one. I think I can sometimes provide value, but you and @tmy1313 do the bulk of this work. I feel the same way on the UI stuff, where I'd be happy to have either of you in there as well, but don't want to oversubscribe anyone

tmy1313 (approved):

PR #1728, opened 2024-10-23
OCD-4660 - Update functionality for SSO flag

tmy1313:
OCD-4660


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/AuthenticationController.java
```diff
@@ -59,21 +76,142 @@ public class AuthenticationController {
...
+    @Operation(summary = "Log user out.",
+            description = "Invalidates all of the tokens associated with the user.",
+            security = {
+                    @SecurityRequirement(name = SwaggerSecurityRequirement.API_KEY)
+            }
+        )
+    @RequestMapping(value = "/logout", method = RequestMethod.POST,
+            consumes = MediaType.APPLICATION_JSON_VALUE, produces = "application/json; charset=utf-8")
+    public void logout(@RequestBody CognitoLogoutRequest request) {
+        cognitoAuthenticationManager.invalidateTokensForUser(request.getEmail());
```
andlar, line 139: Does this need a FF4J check? It looks like most of the other endpoints have one. For the ones that don't, should they as well?

chpl/chpl-service/src/main/java/gov/healthit/chpl/auth/user/JWTAuthenticatedUser.java
```diff
@@ -22,6 +23,8 @@
 public class JWTAuthenticatedUser implements UserDetails, Authentication {
     private static final long serialVersionUID = -7558546038256722930L;
 
+    @DeprecatedResponseField(message = "This field is deprecated and will be removed.", removalDate = "2024-11-01")
```
andlar: I believe we usually give six months warning on deprecation

tmy1313: This actually doesn't need `@DeprecatedResponseField`.  It is not an object that is publicly available and is only used internally within the API.

andlar (dismissed):

kekey1 (approved):

chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -362,9 +485,9 @@ public DeletedUser deleteUser(@PathVariable("userId") Long userId) throws UserRe
     @PreAuthorize("isAuthenticated()")
     public @ResponseBody UsersResponse getUsers() {
         List<User> users = null;
-        if (AuthUtil.getCurrentUser().getAuthenticationSystem().equals(AuthenticationSystem.COGNITO)) {
+        if (ff4j.check(FeatureList.SSO)) {
```
kekey1, line 483: This endpoint is marked as Deprecated/DeprecatedApi but I am not sure that it is. It is being used from the Users -> User Management page in both flag states. If it is truly deprecated then I would expect a different endpoint to be called when the SSO flag is ON.

kekey1 (changes requested):

kekey1 (approved):

andlar (approved):

PR #1732, opened 2024-10-29
feat: add new endpoints to add organizations to an existing user

tmy1313:
OCD-4721


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/UserManagementController.java
```diff
@@ -92,6 +96,24 @@ public UserManagementController(UserManager userManager, InvitationManager invit
         this.authorizationLengthInDays = authorizationLengthInDays;
     }
 
+    @Operation(summary = "Update the currently logged in user with an additional organization.",
+            description = "Gives the user permission on the object in the invitation (usually an additional ACB or Developer)."
```
kekey1: I think this documentation is copied from the CHPL-login workflow methods. The line about which endpoints to call in which order is not right for the Cognito methods. Can you check the other Cognito auth methods as well? They are not currently in the diff but it looks like they might have the same documentation.

chpl/chpl-service/src/main/java/gov/healthit/chpl/user/cognito/CognitoApiWrapper.java
```diff
@@ -181,9 +186,28 @@ public User getUserInfo(UUID cognitoId) throws UserRetrievalException {
             }
         }
         return null;
+    }
+
+    public User getUserByPassingCognitoUserAttributeCache(String accessToken) throws UserRetrievalException {
```
kekey1: When I first read this method name I thought it was two words "by passing" but I think it's meant as one word "bypassing" - can you change the camelcase to make this clear? getUserBypassingCognitoUserAttributeCache, or maybe even simpler getUserNoCache?

kekey1 (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/user/cognito/CognitoApiWrapper.java
```diff
@@ -171,39 +171,37 @@ public AuthenticationResultType respondToNewPasswordRequiredChallenge(CognitoNew
...
+        AdminGetUserResponse response = cognitoClient.adminGetUser(request);
+        if (response == null || response.sdkHttpResponse() == null || !response.sdkHttpResponse().isSuccessful()) {
+            return null;
         }
-        return null;
+        return createUserFromGetUserResponse(response);
     }
 
+
+<<<<<<< Updated upstream
```
kekey1: Merge issue

kekey1 (changes requested):

chpl/chpl-service/src/main/java/gov/healthit/chpl/user/cognito/CognitoApiWrapper.java
```diff
@@ -371,6 +367,31 @@ public void updateUser(User user) throws UserRetrievalException {
         cognitoClient.adminUpdateUserAttributes(request);
     }
 
+    @CacheEvict(value = CacheNames.COGNITO_USERS, key = "#user.cognitoId")
```
kekey1: It looks like we cache users both with UUID as a key and email as a key (see above method `getUserInfo(String email)`). But here we are evicting only the UUID key. I'm not 100% sure I'm correct but I think so... can you check on this?

kekey1 (changes requested):

kekey1 (dismissed):

andlar (dismissed):

kekey1 (approved):

andlar (approved):

PR #1738, opened 2024-11-13
OCD-4725 - As an end-user, I want to see a breakdown of criteria attributes - Standards

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriaattribute/StandardReportDao.java
```diff
@@ -0,0 +1,108 @@
...
+                        .standard(getStandard((Long) result[1]))
+                        .count((Long) result[2])
+                        .build())
+                .toList();
+    }
+
+    public List<StandardListingReport> getStandardListingReports() {
+        String hql = "SELECT cc, s, cpd.chplProductNumber "
+                + "FROM CertificationCriterionEntity cc, "
+                + "CertificationEditionEntity ce, "
```
kekey1: I was wondering why you need to join CertificationEditionEntity here but not in the query above. Could this possibly exclude newer criteria that have no edition? Maybe it's not needed?

chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriaattribute/StandardReportDao.java
```diff
@@ -0,0 +1,108 @@
...
+                        .criterion(((CertificationCriterionEntity) result[0]).toDomain())
+                        .standard(((StandardEntity) result[1]).toDomain())
+                        .chplProductNumber((String) result[2])
+                        .build())
+                .toList();
+    }
+
+    private CertificationCriterion getCertificationCriterion(Long id) {
+        try {
+            return certificationCriterionDao.getById(id);
```
kekey1: I was wondering if you would want to use CertificationCriterionService here which has a method to get by ID. I was just thinking it would be less queries that have to run since we have all the criteria cached in that class. Hibernate has it's own caching thing as well, so I dunno. But I'll throw it out as a suggestion to use that service instead.

chpl/chpl-service/src/main/java/gov/healthit/chpl/report/criteriaattribute/TestToolReportDao.java
```diff
@@ -73,13 +75,15 @@ public List<TestToolReport> getTestToolReports() {
     }
 
     public List<TestToolListingReport> getTestToolListingReports() {
-        String hql = "SELECT cc.id as certificationCriterionId, tt.id as testToolId, cpd.chplProductNumber "
+        String hql = "SELECT cc, tt, cpd.chplProductNumber "
                 + "FROM CertificationCriterionEntity cc, "
+                + "CertificationEditionEntity ce, "
```
kekey1: Same question about excluding criteria that have no edition. I double checked and there is only 1 now but I imagine there would be more in the future.

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1772, opened 2025-01-08
OCD-4785 - Handle Service Base URL Datadog "Body not empty" check failure

tmy1313:


chpl/chpl-service/src/main/java/gov/healthit/chpl/scheduler/job/urluptime/DatadogSyntheticsTestService.java
```diff
@@ -110,7 +110,7 @@ public SyntheticsAPITest createSyntheticsTest(String url, List<Long> developerId
                                         .type(SyntheticsAssertionType.STATUS_CODE)),
                                 new SyntheticsAssertion(new SyntheticsAssertionTarget()
                                         .operator(SyntheticsAssertionOperator.MATCHES)
-                                        .target("/[\\S\s]+[\\S]+/")
+                                        .target("/[\\S]/")
```
andlar: I'm thrown by the regex not having the `+` after the `[\S]`, though I don't think I'm against it. Just want to make sure it's what we want

kekey1: I approve the PR, but was wondering if we could make the regex a constant and use it in the job just to keep us honest about using the same values in both places.

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #1780, opened 2025-02-05
OCD-4517: AQA tests for /real-world-testing/plans and /results endpoints

athomas-git:
OCD-4517: AQA tests for /real-world-testing/plans and /results endpoints

### DIFF
--- a/chpl/chpl-api/e2e/collections/report-data.postman_collection.json
+++ b/chpl/chpl-api/e2e/collections/report-data.postman_collection.json
@@ -618,23 +618,72 @@
 			"response": []
 		},
 		{
-			"name": "GET /report-data/non-conformities/types - 200 status and valid response",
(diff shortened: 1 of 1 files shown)


ashwinimore (approved):

chpl/chpl-api/e2e/collections/report-data.postman_collection.json
```diff
@@ -618,23 +618,72 @@
 			"response": []
 		},
 		{
-			"name": "GET /report-data/non-conformities/types - 200 status and valid response",
+			"name": "GET /report-data/real-world-testing/plans - 200 status and valid response",
```
andlar, line 621: Are we not testing the non-conformities endpoints any more? It looks like these changes remove that test. If the non-conformities endpoint is deprecated, or no longer used, or something, then I could see why that would make sense, but I don't see any indication of that change in the API PR, or on the API Information page on QA

PR #1792, opened 2025-02-24
OCD-4783: Endpoints for new system admin pages

kekey1:


chpl/chpl-api/src/main/java/gov/healthit/chpl/web/controller/CqmCriteriaController.java
```diff
@@ -0,0 +1,38 @@
...
+import gov.healthit.chpl.cqm.CqmCriteriaManager;
+import gov.healthit.chpl.util.SwaggerSecurityRequirement;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.security.SecurityRequirement;
+import io.swagger.v3.oas.annotations.tags.Tag;
+
+@Tag(name = "cqms", description = "Endpoints related to Clinical Quality Measures.")
+@RestController
+@RequestMapping("/cqms")
+public class CqmCriteriaController {
```
andlar: Should `Criteria` be in this controller name / filename / etc.?

kekey1: I don't know. For that matter, is it really "eCqms"? We have really inconsistent naming of this throughout. The database table is called cqm_criterion. In many places in the code we call one of these a CqmCriterion. In many other places we do not. There are plenty of variable and method names where it is simply cqm/cqms. When I made the class I was thinking of it as something analogous to the CertificationCriteriaController, so I went with CqmCriteriaController. I also sometimes get the term "CQM Criterion" confused with the 4 c-criteria that can be associated with any attested CQM.

tmy1313 (dismissed):

andlar (dismissed):

tmy1313 (dismissed):

andlar (approved):

tmy1313 (approved):

PR #1849, opened 2025-06-26
OCD-4909: Update CMS ID logic to support annual changes to format

kekey1:


chpl/chpl-service/src/main/java/gov/healthit/chpl/certificationId/CertificationIdDAO.java
```diff
@@ -308,15 +312,15 @@ private List<CertificationIdAndCertifiedProductEntity> getAllCertificationIdsWit
                 CertificationIdAndCertifiedProductEntity.class).getResultList();
     }
 
-    private String generateCertificationIdString(List<CertifiedProductDetailsDTO> listings, String year) throws EntityCreationException {
+    private String generateCertificationIdString(String year) throws EntityCreationException {
         // Form the EHR Certification ID prefix and edition year identifier.
         // The identifier begins with the two-digit year followed by a "C" to indicate
```
andlar: I know comments are dumb, but this stuff is even more obsolete than it used to be. Should we remove it, if only to avoid having to read it again?

andlar (dismissed):

chpl/chpl-service/src/main/java/gov/healthit/chpl/certificationId/CertificationIdDAO.java
kekey1: I cleaned up some of the comments. I actually think some of them are useful in this file, because they explain Noby's intent to me which I think is important if we ever found out that in some scenario the code didn't run as intended or needed to explain some historical decision to ASTP. But I agree on removing this section since we aren't going to be generating IDs like that anymore.

andlar (approved):

PR #1865, opened 2025-08-08
OCD-4897: updated tests to remove any gap value

athomas-git:
OCD-4897: updated tests to remove any gap value


kekey1 (approved):
I can't view the changes at all in Github. The file is too large. I approve as long as the tests are passing and you removed everywhere there was a "gap" field

PR #1879, opened 2025-09-30
OCD-5054: Make test data read-only in the API

kekey1:


kekey1:
Summary of the PR:

The main objective is to make "test data" read-only within the API.
A significant amount of code has been removed, indicating the removal of logic that previously allowed editing or mutating test data resources.
The PR impacts a wide surface area across the repository, suggesting changes to various endpoints or service classes related to test data.

andlar (approved):

